PR #26, opened 2018-12-08
add eta to log

thangvubk:
Add eta to text log. Here is output from mmdet:
``2018-12-08 07:17:36,804 - INFO - Epoch [3][100/7330]    lr: 0.02000, eta: 9:45:12, time: 0.473, data_time: 0.007, loss_rpn_cls: 0.0338, loss_rpn_reg: 0.0327, loss_cls: 0.2254, acc: 92.6582, loss_reg: 0.1204, loss: 0.4123``
p/s: Not sure ``time`` is already included ``data_time`` or not.

### DIFF
--- a/mmcv/runner/hooks/logger/text.py
+++ b/mmcv/runner/hooks/logger/text.py
@@ -1,8 +1,13 @@
-from .base import LoggerHook
 import datetime
+
+from .base import LoggerHook
 
 
 class TextLoggerHook(LoggerHook):
+
+    def __init__(self, interval=10, ignore_last=True, reset_flag=False):
+        super(TextLoggerHook, self).__init__(interval, ignore_last, reset_flag)
+        self.time_sec_tot = 0
 
     def log(self, runner):
         if runner.mode == 'train':
@@ -15,10 +20,10 @@ class TextLoggerHook(LoggerHook):
             log_str = 'Epoch({}) [{}][{}]\t'.format(runner.mode, runner.epoch,
                                                     runner.inner_iter + 1)
         if 'time' in runner.log_buffer.output:
-            tot_time_sec = runner.log_buffer.output['time'] + \
-                runner.log_buffer.output['data_time']
-            eta_sec = tot_time_sec * \
-                (len(runner.data_loader) * runner.max_epochs - runner._iter)
+            self.time_sec_tot += (runner.log_buffer.output['time'] * 
+                                  self.interval)
+            time_sec_avg = self.time_sec_tot / (runner.iter + 1)
+            eta_sec = time_sec_avg * (runner.max_iters - runner.iter - 1)
             eta_str = str(datetime.timedelta(seconds=int(eta_sec)))
             log_str += ('eta: {}, '.format(eta_str))
             log_str += (
--- a/mmcv/runner/hooks/logger/text.py
+++ b/mmcv/runner/hooks/logger/text.py
@@ -25,7 +25,7 @@ class TextLoggerHook(LoggerHook):
             time_sec_avg = self.time_sec_tot / (runner.iter + 1)
             eta_sec = time_sec_avg * (runner.max_iters - runner.iter - 1)
             eta_str = str(datetime.timedelta(seconds=int(eta_sec)))
-            log_str += ('eta: {}, '.format(eta_str))
+            log_str += 'eta: {}, '.format(eta_str)
             log_str += (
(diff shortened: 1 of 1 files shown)


mmcv/runner/hooks/logger/text.py
```diff
@@ -1,4 +1,5 @@
 from .base import LoggerHook
+import datetime
```
hellock: We adopt the following import orders in the project.
```
system imports
[blank line]
third party imports
[blank line]
relative imports
```

mmcv/runner/hooks/logger/text.py
```diff
@@ -14,6 +15,12 @@ def log(self, runner):
             log_str = 'Epoch({}) [{}][{}]\t'.format(runner.mode, runner.epoch,
                                                     runner.inner_iter + 1)
         if 'time' in runner.log_buffer.output:
+            tot_time_sec = runner.log_buffer.output['time'] + \
+                runner.log_buffer.output['data_time']
+            eta_sec = tot_time_sec * \
+                (len(runner.data_loader) * runner.max_epochs - runner._iter)
```
hellock: There are some indicators that can be used to compute the remaining iters.
- runner.max_iters
- runner.iter

mmcv/runner/hooks/logger/text.py
```diff
@@ -14,6 +15,12 @@ def log(self, runner):
...
+            eta_str = str(datetime.timedelta(seconds=int(eta_sec)))
+            log_str += ('eta: {}, '.format(eta_str))
```
hellock: The time per iteration is not very stable so I think it might be better to just show a rough eta, e.g., 16.1 h

mmcv/runner/hooks/logger/text.py
```diff
@@ -14,6 +15,12 @@ def log(self, runner):
...
             log_str += (
                 'time: {log[time]:.3f}, data_time: {log[data_time]:.3f}, '.
                 format(log=runner.log_buffer.output))
```
hellock: Would it be better to show time and data_time first and then eta?

hellock:
Thanks for the PR, I think it is useful. `time` has already included `data_time`.

mmcv/runner/hooks/logger/text.py
thangvubk: In my opinion (as a user), I prefer to see eta first. Time and data_time is important for performance comparison but they are nearly unchanged during training so less informative from a user point of view.

mmcv/runner/hooks/logger/text.py
thangvubk: I think the best solution is using moving average eta since runner is generalized for both long and short training.

mmcv/runner/hooks/logger/text.py
hellock: This also makes sense, the order may be left unchanged.

mmcv/runner/hooks/logger/text.py
hellock: Yes, it would be better to use moving average or all the past time for time estimation.
Different metrics needs different processing and currently all metrics are computed as the average of past 50 iterations (time, loss, etc.) or no average (mAP, lr). An abstract of metric is required for controlling the behavior more flexibly.

thangvubk: I totally agree with your opinion.

hellock:
I think it is ok to left the refactoring for further work and this PR is acceptable.

mmcv/runner/hooks/logger/text.py
```diff
@@ -14,6 +20,12 @@ def log(self, runner):
             log_str = 'Epoch({}) [{}][{}]\t'.format(runner.mode, runner.epoch,
                                                     runner.inner_iter + 1)
         if 'time' in runner.log_buffer.output:
+            self.time_sec_tot += (runner.log_buffer.output['time'] *
+                                  self.interval)
+            time_sec_avg = self.time_sec_tot / (runner.iter + 1)
+            eta_sec = time_sec_avg * (runner.max_iters - runner.iter - 1)
+            eta_str = str(datetime.timedelta(seconds=int(eta_sec)))
+            log_str += ('eta: {}, '.format(eta_str))
```
hellock: A minor comment: the brackets here are unnecessary since this line is short. I will merge this PR after the fix.

thangvubk: thanks. I didnt notice that.

PR #27, opened 2018-12-19
add open-mmlab urls

OceanPang:


mmcv/runner/checkpoint.py
```diff
@@ -80,9 +93,21 @@ def load_checkpoint(model,
...
     if filename.startswith('modelzoo://'):
-        from torchvision.models.resnet import model_urls
+        import torchvision
+        import pkgutil
+        import inspect
+        model_urls = dict()
+        for _, name, ispkg in pkgutil.walk_packages(
+                torchvision.models.__path__):
+            _zoo = getattr(torchvision.models, name)
+            if inspect.ismodule(_zoo):
```
hellock: We can use `ispkg` to check if is a module.

hellock:
```python
import_module('torchvision.models.' + name)
```

Issue #24, opened 2018-11-27
Set logger format and filename outside runner?


wangg12:
I think it would be more convenient to set a custom logger outside runner because I want to use a custom logger without changing the code of mmcv.

PR #29, opened 2019-01-06
allow passing custom logger to runner

wangg12:
#24

### DIFF
--- a/mmcv/runner/runner.py
+++ b/mmcv/runner/runner.py
@@ -27,6 +27,7 @@ class Runner(object):
         work_dir (str, optional): The working directory to save checkpoints
             and logs.
         log_level (int): Logging level.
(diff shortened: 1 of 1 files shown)


mmcv/runner/runner.py
```diff
@@ -27,14 +27,16 @@ class Runner(object):
         work_dir (str, optional): The working directory to save checkpoints
             and logs.
         log_level (int): Logging level.
+        logger: Custom logger. If `None`, use default logger.
```
hellock: We need to make the docstring consistent with the following format.

argument name (argument type): some text

wangg12: Is it `logging.Logger`?

hellock: Yep.

PR #35, opened 2019-02-16
add cosine lr schedule

wangg12:


wangg12:
[https://github.com/open-mmlab/mmdetection/issues/320](https://github.com/open-mmlab/mmdetection/issues/320)

hellock:
Thanks a lot! It will be merged after validation. You can also provide some results if you have tried the cosine schedule.

wangg12:
In my experiment, cosine lr schdule is comparable to step schedule.

PR #60, opened 2019-04-21
fix possible encoding error when install

wangg12:
When install on windows, there is a potential bug:
```
UnicodeDecodeError: 'gbk' codec can't decode byte 0x99 in position 501: illegal multibyte sequence
```

Fix according to [https://github.com/JuliaPy/pyjulia/pull/219/files](https://github.com/JuliaPy/pyjulia/pull/219/files)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,6 @@
 import sys
+from io import open  # for Python 2 (identical to builtin in Python 3)
+
 from setuptools import find_packages, setup
 
 install_requires = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,4 +1,6 @@
 import sys
+from io import open
```
hellock: It may be better to add a comment here to address the python2 compatibility

wangg12: OK. I've added.

hellock:
Thanks!

PR #61, opened 2019-04-22
Add caffe2_xavier_init()

hellock:
The implementations of caffe2 and pytorch are different. `XavierFill` in Caffe2 corresponds to `kaiming_uniform_` in PyTorch.


innerlee:
What's the purpose of bias? It is not used.

PR #66, opened 2019-04-30
add bgr2hls and hls2bgr

wangg12:


hellock:
Thanks for the new features! You need to write some unit tests in [test_image.py](https://github.com/open-mmlab/mmcv/blob/master/tests/test_image.py) to fully cover your newly added coded.

PR #72, opened 2019-05-30
Tentative Pythran support for flow_warp

serge-sans-paille:
This should have performance close to C++ without the hassle of writing actual C++ code.
The goal of this PR is to start a discussion and gather feedback, not necessarily to have the code in.


serge-sans-paille:
Patch ready for review! Any thoughts?

zhouzaida:
hi @serge-sans-paille , thanks for your contribution. Could you provide the performance comparison

serge-sans-paille:
@zhouzaida With a few changes in pythran (actually https://github.com/serge-sans-paille/pythran/pull/1812 ) and in the .py file (see below), I get the following timing with cython:

```
$ python -m timeit -s 'from fwm_cython import flow_warp_c as f; import numpy as np; x = np.ones((128, 128, 3)); y = np.ones((128, 128, 2))' 'f(x, y)'
2000 loops, best of 5: 161 usec per loop
```

and with pythran

```
$ python -m timeit -s 'from fwm_pythran import flow_warp_c as f; import numpy as np; x = np.ones((128, 128, 3)); y = np.ones((128, 128, 2))' 'f(x, y)'
https://github.com/serge-sans-paille/pythran/pull/1812
```

Updated pythran code:

```
import numpy as np


def EnforceRange(x, MaxValue):
    return min(max(x, 0), MaxValue)

def flow_warp_core(img, flow, filling_value, interpolator):
    out = np.zeros_like(img)
    height, width, channels = out.shape
    for h in range(height):
        for w in range(width):
            y = w + flow[h, w, 0]
            x = h + flow[h, w, 1]

            if x < 0 or x >= height - 1 or y < 0 or y >= width - 1:
                for i in range(channels):
                    out[h, w, i] = filling_value
                continue

            interpolator(img, width, height, x, y, out[h, w])
    return out


# pythran export flow_warp_c(float64[:,:,:], float64[:,:,2], int?, int?)
def flow_warp_c(img, flow, filling_value=0, interpolate_mode=1):
    if interpolate_mode == 0:
        return flow_warp_core(img, flow, filling_value, BilinearInterpolate)
    elif interpolate_mode == 1:
        return flow_warp_core(img, flow, filling_value, NNInterpolate)
    else:
        raise NotImplementedError("Interpolation Method")


def BilinearInterpolate(img, width, height, x, y, out):
    xx = int(x)
    yy = int(y)

    dx = max(min(x - xx, 1.), 0.)
    dy = max(min(y - yy, 1.), 0.)

    for m in range(2):
        for n in range(2):
            u = EnforceRange(yy + n, width)
            v = EnforceRange(xx + m, height)
            s = abs(1 - m - dx) * abs(1 - n - dy)
            for i in range(len(out)):
                out[i] += img[v, u, i] * s


def NNInterpolate(img, width, height, x, y, out):
    xx = int(x)
    yy = int(y)

    dx = max(min(x - xx, 1.), 0.)
    dy = max(min(y - yy, 1.), 0.)

    m = 0 if dx < 0.5 else 1
    n = 0 if dy < 0.5 else 1

    u = EnforceRange(yy + n, width)
    v = EnforceRange(xx + m, height)
    for i in range(len(out)):
        out[i] += img[v, u, i]
```

zhouzaida:
got it

innerlee:
I'm not familiar with flow warp code. As long as it passes tests (or add tests for verifying accuracy if not exist), and does not causes slow down (importing mmcv or running the function), I'm okay with it.

One thing to notice is that this introduces a dependency pythran. Is it easy to install in windows/mac/etc.? Will the compilation process as smooth (or unsmooth as always) as before? Historically, things like numba will cause compatibility issues with numpy or things like that.

serge-sans-paille:
@innerlee pythran is now a build-time dependency for scipy, it builds fine on Linux, OSX and Windows, but requires the clang compiler to be installed on windows. That's not a runtime dependency, but that indeed make the build slightly more difficult on that platform.

ZwwWayne:
involve @nbei  to have a look.

innerlee:
And maybe @MeowZheng

ZwwWayne:
Since flow operators has been removed from mmcv and supported in [mmflow](https://github.com/open-mmlab/mmflow).

OpenMMLab-Assistant-004:
Hi @serge-sans-paille ！We are grateful for your efforts in helping improve mmcv open-source project during your personal time. 

PR #92, opened 2019-07-28
add tensor shape check when loading checkpoint

CoinCheung:
In many occasions, we need to initialize the model from the coco pretrain models in the model zoo. It is common that the users' datasets have different number of categories, which makes it not convenient to load from the 81 categories coco model directly. Thus I feel maybe it is meaningful to add a few lines to check the shape of the model paremeters and skip those with different shapes.

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -2,11 +2,13 @@ import os
 import os.path as osp
 import pkgutil
 import time
+import warnings
 from collections import OrderedDict
 from importlib import import_module
 
 import mmcv
 import torch
+import torchvision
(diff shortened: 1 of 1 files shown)


CoinCheung:
I do not understand the error message. Is it my fault ?

hellock:
Never mind. The error is caused by travis itself.

Some comments here:
- This PR seems not synchronized with the master branch due to the force push, you may want to fix and reopen it.
- Skipping the parameters of different shape may not be the default behavior, you can add an argument.
- The code can be added to `load_state_dict()` instead of `load_checkpoint()`.

CoinCheung:
OK, I am closing this and resubmit a new pr.

PR #95, opened 2019-07-30
Update lr_updater.py

harryhan618:
"Add min_lr to PolyLrUpdaterHook, so that it can anneal to min_lr. When min_lr=0, it behaves as usual. "

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -139,8 +139,9 @@ class ExpLrUpdaterHook(LrUpdaterHook):
 
 class PolyLrUpdaterHook(LrUpdaterHook):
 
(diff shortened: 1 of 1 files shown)


mmcv/runner/hooks/lr_updater.py
```diff
@@ -139,8 +139,9 @@ def get_lr(self, runner, base_lr):
 
 class PolyLrUpdaterHook(LrUpdaterHook):
 
-    def __init__(self, power=1., **kwargs):
+    def __init__(self, power=1., min_lr=0.,  **kwargs):
```
hellock: There are two whitespaces, while only one is expected.

harryhan618: It is revised.

PR #101, opened 2019-07-31
install cython automatically as a dependency

karlind:


setup.py
```diff
@@ -1,7 +1,9 @@
 import platform
 import sys
 from io import open  # for Python 2 (identical to builtin in Python 3)
-from setuptools import Extension, find_packages, setup
+from setuptools import Extension, find_packages, setup, dist
+
+dist.Distribution().fetch_build_eggs(['Cython'])
```
hellock: `numpy` may also be included..

karlind: `numpy` has been included

hellock:
Thanks for the fix! Could you check the linting and formatting errors?

karlind:
Actually I have already checked the errors.
The errors say,
```
$ flake8
./setup.py:8:1: E402 module level import not at top of file
./setup.py:9:1: E402 module level import not at top of file
The command "flake8" failed and exited with 1 during .
```
Because `dist.Distribution().fetch_build_eggs(['Cython', 'numpy>=1.11.1'])` must be exectued before `import`, so it seems not easy to fix these linting errors.

karlind:
May be we can ignore such errors by using
```
flake8 --ignore=E402 ./setup.py
```

hellock:
> May be we can ignore such errors by using
> 
> ```
> flake8 --ignore=E402 ./setup.py
> ```

Use inline comments instead.

```python
import numpy as np  # noqa: E402
```

PR #103, opened 2019-08-06
Fix multiple err_msg logging in distribuited mode

wxzs5:
Only need log error info 1 time when load state_dict in distributed mode.


hellock:
Thanks for the fix!
In mmdetection, we handle the logging with https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/env.py#L67-L68, but this PR is okay to me.

PR #110, opened 2019-08-20
Track iter progress

ZwwWayne:
Currently, the progress bar only supports functions & tasks, but not support iterators, which is also a common use of the progress bar, so I add two functions, track_iter_progress & track_enum_progress that could track the progress of iterations in for loop.


ZwwWayne:
Currently, the progress bar only supports functions & tasks, but not support iterators, which is also a common use of the progress bar, so I add two functions, track_iter_progress & track_enum_progress that could track the progress of iterations in for loop.

This branch has been reformated with yapf & pep8, sorry for missing reformate before.

PR #112, opened 2019-08-20
Track iter progress

ZwwWayne:
Currently, the progress bar only supports functions & tasks, but not support iterators, which is also a common use of the progress bar, so I add two functions, track_iter_progress & track_enum_progress that could track the progress of iterations in for loop.

This branch has been reformated with yapf & pep8, after adding unitest, it passed the local code coverage test.


docs/utils.md
```diff
@@ -54,6 +54,25 @@ mmcv.track_parallel_progress(func, tasks, 8)  # 8 workers
 
 ![progress](_static/parallel_progress.gif)
 
+What is more, if you want to iterate or enumerate a list of items and track the progress, `track_iter_progress`
```
hellock: `What is more` may be unnecessary.

mmcv/utils/progressbar.py
```diff
@@ -172,3 +172,35 @@ def track_parallel_progress(func,
...
+
+    Tasks are done with a simple for-loop.
+
+    Args:
+        func (callable): The function to be applied to each task.
+        tasks (list or tuple[Iterable, int]): A list of tasks or
+            (tasks, total num).
+        bar_width (int): Width of progress bar.
+
+    Returns:
```
hellock: `Yields` instead of `Returns`.

mmcv/utils/progressbar.py
```diff
@@ -172,3 +172,35 @@ def track_parallel_progress(func,
...
     return results
+
+
+def track_iter_progress(tasks, bar_width=50, **kwargs):
+    """Track the progress of tasks execution with a progress bar.
+
+    Tasks are done with a simple for-loop.
+
+    Args:
+        func (callable): The function to be applied to each task.
```
hellock: Incorrect argument.

ZwwWayne: Sorry for missing these, I have updated a new version of the docs and code comments.

Issue #114, opened 2019-08-21
Should progressbar goes to stderr?


innerlee:
Sometimes I want to capture the output of some of my downstream scripts. The progress is meant for indication and does not really belong to the things I want to capture in stdout.

PR #115, opened 2019-08-21
Let progress goes to stderr

innerlee:
Fix https://github.com/open-mmlab/mmcv/issues/114


hellock:
It is better to make it an argument of progress bar, like `file=sys.stdout`.

innerlee:
Not sure why tests fail

hellock:
> Not sure why tests fail

me too...

innerlee:
Let's blame pytest https://github.com/pytest-dev/pytest/issues/1132

PR #121, opened 2019-08-29
Add cyclic policy

zzzfinal0:
Add cyclic policy, of which parameters and behavior are almost the same as [torch.optim.lr_scheduler.CyclicLR](https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.CyclicLR).
Here is an example of mmdetection config:
```
        # optimizer
        optimizer = dict(type='SGD', lr=0.002, momentum=0.8, weight_decay=0.0001)
        # learning policy
        lr_config = dict(
            warmup='linear',
            warmup_iters=300,
            warmup_ratio=1.0 / 3,
            policy='cyclic',
            by_epoch = False,
            max_lr = 0.02,
            step_size_up=1000,
            max_momentum=0.9
)
```
Paper:[Cyclical Learning Rates for Training Neural Networks](https://arxiv.org/abs/1506.01186)
Performance comparison: [adaptive-and-cyclical-learning-rates-using-pytorch](https://towardsdatascience.com/adaptive-and-cyclical-learning-rates-using-pytorch-2bf904d18dee) 
More examples: [https://github.com/bckenstler/CLR](https://github.com/bckenstler/CLR)


zzzfinal0:
"The command "yapf -r -d mmcv/ tests/ examples/" failed and exited with 1 during ."
I have no idea how to fix that. Yapf finished with no error on my machine and no error.

PR #122, opened 2019-08-29
[feat]: find root directory of Version Control Systems

OneDirection9:
Usually, we put data into a `data` folder and put code into a `lib` folder. With `find_vcs_root` function can get the data folder easily. Example:

```python
# The file in the `lib` directory
root_dir = osp.join(find_vcs_root(__file__), 'data')
```

PS: The urls of `.isort.cfg` and `.style.yapf` in `CONTRIBUTING.md` are invalid.

Signed-off-by: Zhipeng Han <hanzhipeng9@gmail.com>


hellock:
Thanks for the contribution! Could you fix the minor linting error according to the CI results?

PR #132, opened 2019-09-11
Added solarize and posterize image transformations

dhananjaisharma10:
Hi!

I have added the solarize and posterize image transformations in the following module: `mmcv/image/transforms/colorspace.py` and have updated the necessary `__init__.py` modules. Please let me know if anything needs to be changed.

Thanks!


hellock:
Sorry for the late review. It would be great to add new features. Could you also add the necessary unit tests?

PR #135, opened 2019-09-24
is_filepath may be not right?

dongfangduoshou123:
"""
I see the is_filepath now is implemented as follows:
`def is_filepath(x):
    if is_str(x) or isinstance(x, Path):
        return True
    else:
        return False`

I think if is_str(x) return True, the x also may be not a valid filepath(example x = 'slgljglsjgj'
it is a string, but not a valid filepath.
"""

### DIFF
--- a/mmcv/utils/path.py
+++ b/mmcv/utils/path.py
@@ -14,7 +14,7 @@ else:
 
 
 def is_filepath(x):
-    if is_str(x) or isinstance(x, Path):
+    if os.path.isfile(x) or isinstance(x, Path):
         return True
     else:
         return False


hellock:
The method `is_filepath` acts as an wrapper for string paths and Path objects. It only performs the type check to see if it is a valid type for a file path instead of checking if it is an existing path.

dongfangduoshou123:
ok, thank you for your explain.

OpenMMLab-Assistant-004:
Hi @dongfangduoshou123 ！We are grateful for your efforts in helping improve mmcv open-source project during your personal time. 

PR #138, opened 2019-10-02
Report number of matching state_dict keys

Erotemic:
When trying to load pretrained weights, it would be useful to know how many weights successfully loaded in addition to knowing which weights were missing / unexpected. This PR adds a few lines to report those numbers when the weights don't match exactly.

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -84,12 +84,10 @@ def load_state_dict(module, state_dict, strict=False, logger=None):
     err_msg = []
     if unexpected_keys:
         err_msg.append('{} unexpected keys in source state_dict: {}\n'.format(
-            len(unexpected_keys),
-            ', '.join(unexpected_keys)))
+            len(unexpected_keys), ', '.join(unexpected_keys)))
     if missing_keys:
         err_msg.append('{} missing keys in source state_dict: {}\n'.format(
-            len(missing_keys),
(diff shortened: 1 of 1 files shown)


zhouzaida:
hi @Erotemic , thanks for your feedback. PyTorch will raise exceptions when keys are not matched so maybe it does not need.

Erotemic:
That is true, torch will error.

However, there is value in reporting the number of mismatched keys. It gives the user information that indicates if they just have a small number of extra keys (the number will be small) that they need to remove or if they inputted a completely incorrect set of weights (the number will be large). So, it does help with debugging.

I do want to give a shoutout to `torch_liberator.load_partial_state`: https://pypi.org/project/torch-liberator/ which uses a graph isomorphism algorithm to find a matching between source and target state dictionaries, and allows for loading only some of the weights.

zhouzaida:
got it. So do we need to invoke `torch_liberator.load_partial_state` to support loading partial weight

Erotemic:
You could use torch-liberator if you wanted to do partial-state-loading. I wasn't suggesting doing that here, it should be a different PR if mmcv wants to go down that road. I was just giving it a shout-out because its a cool piece of related work I did.

zhouzaida:
got it

ZwwWayne:
Suggest resolving conflict first. Sorry that it takes so long for this PR.

PR #141, opened 2019-10-11
use native module.load_state_dict

DXist:
nn.Module.load_state_dict supports `load_state_dict_pre_hooks` that help to maintain backward compatibility when parameter/buffer/module names are changed.

This PR changes `load_state_dict` implementation to use `nn.Module.load_state_dict` method.
`strict` mode is used by default sinces it quickly allows to find compatibility issues.

Previous load_state_dict implementation can be activated through `use_module_load_state_dict` flag.

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -136,7 +136,8 @@ def load_checkpoint(model,
                     filename,
                     map_location=None,
                     strict=False,
-                    logger=None):
+                    logger=None,
+                    use_module_load_state_dict=True):
     """Load checkpoint from a file or URI.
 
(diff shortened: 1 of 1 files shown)


hellock:
Thanks a lot! Switching to the official implementation is necessary for pre_hooks, but a more prefered solution may be to follow the official implementation and integrate it with the current `load_state_dict`. Self-implemented `load_state_dict` supports some additional features to check and show unmatched keys and shapes, which is also helpful.

ZwwWayne:
https://github.com/open-mmlab/mmcv/pull/164 already change to use `module._load_state_dict`, so this PR is closed.

PR #143, opened 2019-10-23
Added Weight and Biases logger.

ternaus:


ternaus:
It is a bit unclear to me how to write tests for W&B logger.

@vanpelt could you please give a hint?

hellock:
Awesome! Does it work with the W&B service now?

vanpelt:
@hellock yep, this should be good to go!  Only issue is a flake8 error in the test.  @ternaus can you fix that?

hellock:
requirements.txt has been removed and `mock` may be added [here](https://github.com/open-mmlab/mmcv/blob/master/.travis.yml#L11) since it is only required for testing.

PR #164, opened 2019-12-30
Load state dict

ZwwWayne:
The original implementation of `load_state_dict ` function directly copy the weights by their name, now it adopts the newest implementation of  `load_state_dict ` from PyTorch 1.3.1. Thus, it could use the `_load_from_state_dict` function implemented for each nn.module, which enables the version control during weight loading and provides flexibility.

This PR is related to the PR https://github.com/open-mmlab/mmdetection/pull/1894 for [mmdetection](https://github.com/open-mmlab/mmdetection) to enable the backward compatibility of `DCNModule`.

### DIFF
--- a/tests/test_progressbar.py
+++ b/tests/test_progressbar.py
@@ -1,13 +1,14 @@
 import sys
 import time
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
 
 import pytest
 
 import mmcv
+
+try:
+    from StringIO import StringIO
+except ImportError:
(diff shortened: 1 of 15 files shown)


tests/test_progressbar.py
```diff
@@ -1,14 +1,15 @@
...
-    from io import StringIO
 
 import pytest
 
 import mmcv
 
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
hellock: These are system imports.

PR #174, opened 2020-01-31
Adaptive progress bar length

innerlee:
not tested yet. wait for a minute before merge

### DIFF
--- a/tests/test_progressbar.py
+++ b/tests/test_progressbar.py
@@ -1,6 +1,6 @@
 # Copyright (c) Open-MMLab. All rights reserved.
-import sys
 import os
+import sys
 import time
 
 try:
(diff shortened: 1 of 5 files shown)


innerlee:
verified locally.

hellock:
Better to add some unittests for the adaptive behavior.

Another thing to confirm: will the checking in each update affects the fps, especially for a large fps?

innerlee:
image

tests/test_progressbar.py
```diff
@@ -1,7 +1,13 @@
 # Copyright (c) Open-MMLab. All rights reserved.
 import sys
+import os
```
hellock: `import os` should be the first statement according to alphabet.

innerlee: why does pre-commit not doing this automatically?

PR #175, opened 2020-01-31
Soon better than past

innerlee:
Sometimes weird things happen. This pr protects us from -1s.

image


hellock:
Should have been fixed in #37.

innerlee:
The screenshot is using master.

The issue is caused by max iter < iter, because sometimes the batch size is changed after resume

hellock:
Neither displaying "soon" nor a negative number is a nice solution. A negative ETA makes users soon realize there is an error, while "soon" may make them confusing. I tend to either fix it or leave it along.

As for this case, the negative ETA is caused by changing batch size, while `resume_from` is designed for continuing experiments without any changes. If architectures or parameters are changed, `load_from` is more suitable.

PR #176, opened 2020-02-02
Use cv2 to normalize img

innerlee:
About 3x speed up.

```python
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import cv2                                                  

In [2]: import numpy as np                                          

In [3]: img = np.random.randn(100,200,3).astype(np.float32)         

In [4]: mean = np.array([1,2,3])                                    

In [5]: std = mean / 10                                             

In [6]: %timeit (img - mean) / std                                  
291 µs ± 1.53 µs per loop (mean ± std. dev. of 7 runs, 1000 loops ea
ch)

In [7]: %timeit cv2.divide(cv2.subtract(img, mean.reshape(1,3)
   ...: .astype(np.float64)), std.reshape(1,3).astype(np.float64))  
87.9 µs ± 201 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Difference is small
```python
In [8]: ans_cv2 = cv2.divide(cv2.subtract(img, mean.reshape(1,3).astype(np
    ...: .float64)), std.reshape(1,3).astype(np.float64))                  

In [9]: ans_np = (img - mean) / std                                       

In [10]: np.abs(ans_np - ans_cv2).max()                                    
Out[10]: 4.291534423828125e-06
```

### DIFF
--- a/mmcv/image/transforms/normalize.py
+++ b/mmcv/image/transforms/normalize.py
@@ -15,7 +15,9 @@ def imnormalize(img, mean, std, to_rgb=True):
 
 
 def imdenormalize(img, mean, std, to_bgr=True):
-    img = (img * std) + mean
+    mean = mean.reshape(1, 3).astype(np.float64)
+    std = std.reshape(1, 3).astype(np.float64)
+    img = cv2.add(cv2.multiply(img, std), mean)
     if to_bgr:
         img = rgb2bgr(img)
     return img
--- a/mmcv/image/transforms/normalize.py
+++ b/mmcv/image/transforms/normalize.py
@@ -18,12 +18,8 @@ def imdenormalize(img, mean, std, to_bgr=True):
     assert img.dtype != np.uint8
     mean = mean.reshape(1, -1).astype(np.float64)
     std = std.reshape(1, -1).astype(np.float64)
-    print(img[0, 0])
     img = cv2.multiply(img, std)  # make a copy
-    print(img[0, 0])
-
     cv2.add(img, mean, img)  # inplace
-    print(img[0, 0])
     if to_bgr:
         cv2.cvtColor(img, cv2.COLOR_RGB2BGR, img)  # inplace
     print(img[0, 0])
--- a/mmcv/image/transforms/normalize.py
+++ b/mmcv/image/transforms/normalize.py
@@ -11,8 +11,8 @@ def imnormalize(img, mean, std, to_rgb=True):
     stdinv = 1 / np.float64(std.reshape(1, -1))
     if to_rgb:
         img = bgr2rgb(img)
-    img = cv2.subtract(img, mean)  # make a copy
-    cv2.multiply(img, stdinv, img)  # inplace
+    cv2.subtract(img, mean, img)
+    cv2.multiply(img, stdinv, img)
     return img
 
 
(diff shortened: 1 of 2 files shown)


innerlee:
multiplication is 1.3x faster than division.

mmcv/image/transforms/normalize.py
```diff
@@ -1,18 +1,23 @@
...
 
 
 def imnormalize(img, mean, std, to_rgb=True):
     img = img.astype(np.float32)
+    mean = mean.reshape(1, -1).astype(np.float64)
+    stdinv = 1 / std.reshape(1, -1).astype(np.float64)
     if to_rgb:
         img = bgr2rgb(img)
-    return (img - mean) / std
+    return cv2.multiply(cv2.subtract(img, mean), stdinv)
```
hellock: Need to convert it back to float32.

innerlee: maybe

innerlee: Actually no. The result keeps `float32` invariant.

innerlee:
Do we need to unit test the output of these functions? It will be an assurance when refactoring.

mmcv/image/transforms/normalize.py
innerlee: ref: https://github.com/innerlee/cvbenchmark/blob/master/cvbenchmark/img_array_normalize.py and https://github.com/innerlee/cvbenchmark#results

hellock: Nice! You may create a top-level directory `benchmarks` and add the benchmarks there.

innerlee: will do in a separate pr

hellock:
Unit tests may be added to increase the coverage rate.

innerlee:
Okay now it is way more faster. Tests added.

```bash

**Comment by innerlee** [comment]
Now even faster

```bash
# img_array_normalize cv2_div    200 loops, best of 5: 1.81 msec per loop
# img_array_normalize cv2_mult   200 loops, best of 5: 1.68 msec per loop
# img_array_normalize cv2_mult2  200 loops, best of 5: 1.33 msec per loop
# img_array_normalize cv2_mult3  200 loops, best of 5: 1.25 msec per loop
# img_array_normalize cv2_mult4  500 loops, best of 5: 594 usec per loop  <-- CURRENT
# img_array_normalize np         100 loops, best of 5: 2.43 msec per loop  <-- ORIGINAL
```

tests/test_image.py
```diff
@@ -56,6 +58,18 @@ def test_imwrite(self):
         os.remove(out_file)
         self.assert_img_equal(img, rewrite_img)
 
+    def test_imnormalize(self):
+        img = self.img
+        baseline = (img[:, :, ::-1] - self.mean) / self.std
+        img = mmcv.imnormalize(img, self.mean, self.std)
```
hellock: `to_rgb=False` is not covered.

innerlee:
editing on webpage using phone is silly

tests/test_image.py
```diff
@@ -56,6 +58,20 @@ def test_imwrite(self):
...
+        baseline = (rgbimg - self.mean) / self.std
+        img = mmcv.imnormalize(self.img, self.mean, self.std)
+        assert np.allclose(img, baseline)
+        img = mmcv.imnormalize(rgbimg, self.mean, self.std, to_rgb=False)
+        assert np.allclose(img, baseline)
+
+    def test_imdenormalize(self):
+        img = (self.img[:, :, ::-1] - self.mean) / self.std
+        baseline = (img * self.std + self.mean)[:, :, ::-1]
+        img = mmcv.imdenormalize(img, self.mean, self.std)
```
hellock: Also `to_rgb=False`.

innerlee:
ping @hellock

PR #177, opened 2020-02-02
Speed up bbox_clip

innerlee:
Believe it or not, this is 1.5x faster


innerlee:
hold this for now. it may contain a bug

innerlee:
Okay bug fixed. Benchmarks see https://github.com/innerlee/cvbenchmark#results and https://github.com/innerlee/cvbenchmark/blob/master/benchmarks/run_bbox_clip.py . BTW the cvbenchmark repo has refactored so that it can moved to mmcv repo. What do you think about its current structure?

hellock:
The benchmark structure looks good to me.

PR #180, opened 2020-02-08
Use official ddp to implement MMDDP

yhcao6:


mmcv/parallel/distributed.py
```diff
@@ -1,54 +1,10 @@
...
-import torch.distributed as dist
-import torch.nn as nn
-from torch._utils import (_flatten_dense_tensors, _take_tensors,
-                          _unflatten_dense_tensors)
+from torch.nn.parallel import DistributedDataParallel
 
 from .scatter_gather import scatter_kwargs
 
 
-class MMDistributedDataParallel(nn.Module):
```
hellock: Move the old implementation to `mmcv.parallel.distributed_deprecated`

PR #183, opened 2020-02-12
log meta

yhcao6:


mmcv/runner/runner.py
```diff
@@ -70,6 +73,7 @@ def __init__(self,
         else:
             self.logger = logger
         self.log_buffer = LogBuffer()
+        self.meta = meta
```
hellock: Check the data type of `meta`.

mmcv/runner/runner.py
```diff
@@ -31,6 +31,8 @@ class Runner(object):
         log_level (int): Logging level.
         logger (:obj:`logging.Logger`): Custom logger. If `None`, use the
             default logger.
+        meta (dict): A dict records some import information such as environment
```
hellock: The type annotation is not correct. It should be `(dict | None)` or `(dict, optional)`

PR #188, opened 2020-02-15
Update version to 0.3.0

hellock:
Breaking changes:
- MMDistributedDataParallel now is a child class of DistributedDataParallel, and this leads to an api change.

Users need to make the following modifications.

1. Change from
    ```python
    model = MMDistributedDataParallel(model.cuda())
    ```
    to
    ```python
    model = MMDistributedDataParallel(
        model.cuda(),
        device_ids=[torch.cuda.current_device()],
        broadcast_buffers=False)
    ```

2. For users who also use DistributedOptimizerHook, change from
    ```python
    def after_train_iter(self, runner):
        runner.optimizer.zero_grad()
        runner.outputs['loss'].backward()
        allreduce_grads(runner.model.parameters(), self.coalesce,
                        self.bucket_size_mb)
        if self.grad_clip is not None:
            self.clip_grads(runner.model.parameters())
        runner.optimizer.step()
    ```
    to
    ```python
    def after_train_iter(self, runner):
        runner.optimizer.zero_grad()
        runner.outputs['loss'].backward()
        if self.grad_clip is not None:
            self.clip_grads(runner.model.parameters())
        runner.optimizer.step()
    ```

### DIFF
--- a/mmcv/version.py
+++ b/mmcv/version.py
@@ -1,2 +1,2 @@
 # Copyright (c) Open-MMLab. All rights reserved.
-__version__ = '0.2.16'
+__version__ = '0.3.0'


twmht:
@hellock 

why you set `broadcast_buffers` to false here?

PR #191, opened 2020-02-18
add inplace imflip

JoannaLXY:
add inplace imflip_()

### DIFF
--- a/mmcv/image/transforms/geometry.py
+++ b/mmcv/image/transforms/geometry.py
@@ -23,7 +23,7 @@ def imflip(img, direction='horizontal'):
 
 
 def imflip_(img, direction='horizontal'):
-    """Inplace flip an image horizontally or vertically.(overwrite img!)
+    """Inplace flip an image horizontally or vertically.
     Args:
         img (ndarray): Image to be flipped.
         direction (str): The flip direction, either "horizontal" or "vertical".
--- a/mmcv/image/transforms/geometry.py
+++ b/mmcv/image/transforms/geometry.py
@@ -22,6 +22,22 @@ def imflip(img, direction='horizontal'):
         return np.flip(img, axis=0)
 
 
+def imflip_(img, direction='horizontal'):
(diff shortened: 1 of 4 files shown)


tests/test_image.py
```diff
@@ -313,6 +313,38 @@ def test_imflip(self):
             for j in range(w):
                 assert flipped_img[i, j] == img[h - 1 - i, j]
 
+    def test_imflip_(self):
+        # test horizontal flip (color image)
+        img = np.random.rand(80, 60, 3)
+        h, w, c = img.shape
+        flipped_img = mmcv.imflip_(img.copy())
+        assert flipped_img.shape == img.shape
```
hellock: You may also need to check if it is truely an inplace operation.

mmcv/image/transforms/geometry.py
```diff
@@ -22,6 +22,22 @@ def imflip(img, direction='horizontal'):
...
+        direction (str): The flip direction, either "horizontal" or "vertical".
+
+    Returns:
+        ndarray: The flipped image(inplace).
+    """
+    assert direction in ['horizontal', 'vertical']
+    if direction == 'horizontal':
+        return cv2.flip(img, 1, img)
+    else:
+        return cv2.flip(img, 0, img)
```
hellock: How about the speed of `cv2.flip(img, 1)` and `np.flip(img, axis=1)`?

innerlee: np.flip is a lazy operation. But it will pay eventually because of it makes array non-contiguous.

cv2 inplace flip is fast, and do not have contiguous issue.

Benchmark (compare with np.flip().copy()) see https://github.com/innerlee/cvbenchmark/blob/master/benchmarks/run_flip.py
and the results:

```bash
./mmbench benchmarks/run_flip.py
# flip cv2  500 loops, best of 5: 915 usec per loop
# flip np   50 loops, best of 5: 7.66 msec per loop
```

PR #198, opened 2020-02-29
Add PaviLoggerHook

OceanPang:
This PR modifies some Logger hooks.
1. Key `init_kwargs` is added to `WandbLoggerHook`.
2. Add `PaviLoggerHook` that based on Pavi. 
    Support `add_scalars`, `add_graph`, `add_snapshot_file`.

### DIFF
--- a/mmcv/runner/hooks/logger/wandb.py
+++ b/mmcv/runner/hooks/logger/wandb.py
@@ -1,10 +1,12 @@
 # Copyright (c) Open-MMLab. All rights reserved.
 import numbers
 
-from ...dist_utils import master_only
+from mmcv.runner import master_only
+from ..hook import HOOKS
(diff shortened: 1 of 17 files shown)


mmcv/runner/hooks/logger/wandb.py
```diff
@@ -8,13 +8,14 @@
 class WandbLoggerHook(LoggerHook):
 
     def __init__(self,
-                 log_dir=None,
+                 init_kwargs=dict(),
```
hellock: Use `None` as the default value to avoid potential issues.

OceanPang: The default key of this key should be an empty dict(), which can init the Writer if there are no init_kwargs given.

PR #202, opened 2020-03-01
Update PaviLoggerHook

OceanPang:
This PR adds a def `is_scalar` to PaviLoggerHook.
Unittest is also added.

### DIFF
--- a/mmcv/runner/hooks/logger/pavi.py
+++ b/mmcv/runner/hooks/logger/pavi.py
@@ -10,12 +10,22 @@ from ..hook import HOOKS
 from .base import LoggerHook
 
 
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/logger/pavi.py
```diff
@@ -1,11 +1,26 @@
...
 
+import numpy as np
+import torch
+
 from mmcv.runner import master_only
 from ..hook import HOOKS
 from .base import LoggerHook
 
 
+def is_scalar(val, np_valid=True, torch_valid=True):
```
hellock: - Add docstring here.
- `np_valid` and `torch_valid` may be renamed to `include_np`, `include_torch`.

PR #208, opened 2020-03-12
support pathlib in imageio

nbei:
RT.

### DIFF
--- a/mmcv/image/io.py
+++ b/mmcv/image/io.py
@@ -1,5 +1,6 @@
 # Copyright (c) Open-MMLab. All rights reserved.
 import os.path as osp
+from pathlib import Path
 
 import cv2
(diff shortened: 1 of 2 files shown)


mmcv/image/io.py
```diff
@@ -61,9 +62,9 @@ def imread(img_or_path, flag='color', channel_order='bgr'):
     """Read an image.
 
     Args:
-        img_or_path (ndarray or str): Either a numpy array or image path.
-            If it is a numpy array (loaded image), then it will be returned
-            as is.
+        img_or_path (ndarray or str): Either a numpy array or str or
```
hellock: The type annotation is not updated.

nbei: Updated.

PR #209, opened 2020-03-14
Warmup byepoch

kennymckormick:
Feature: warmup byepoch
Summary: Sometimes warmup should last n epochs. Since we may use different batchsize for training, the exact number of iterations may change. To use warmup byepoch, when building lr_updater, you need to set `warmup_byepoch=True`.


hellock:
There are some unexpected files included.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -12,6 +12,7 @@ def __init__(self,
                  warmup=None,
                  warmup_iters=0,
                  warmup_ratio=0.1,
+                 warmup_byepoch=False,
```
hellock: `warmup_by_epoch`

mmcv/runner/runner.py
```diff
@@ -336,6 +337,15 @@ def run(self, data_loaders, workflow, max_epochs, **kwargs):
...
                          get_host_info(), work_dir)
         self.logger.info('workflow: %s, max: %d epochs', workflow, max_epochs)
+
+        # set epoch length, in case we use warmup_byepoch
+        # only one tuple in can be with mode 'train'
+        for i, flow in enumerate(workflow):
+            mode, epochs = flow
+            if mode == 'train':
+                self.epoch_len = len(data_loaders[i])
+                break
```
hellock: No need to set `epoch_len` here, we can set it in the hook to reduce the influence to the runner.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -62,6 +64,9 @@ def before_run(self, runner):
         self.base_lr = [
             group['initial_lr'] for group in runner.optimizer.param_groups
         ]
+        if self.warmup_byepoch:
+            assert runner.epoch_len
+            self.warmup_iters *= runner.epoch_len
```
hellock: This can be implemented in `before_train_epoch`, where we can obtain the epoch_len by `len(runner.data_loader)`.

mmcv/runner/runner.py
```diff
@@ -254,6 +255,7 @@ def train(self, data_loader, **kwargs):
         self.model.train()
         self.mode = 'train'
         self.data_loader = data_loader
+        self.active_loader = data_loader
```
hellock: What is the difference from `self.data_loader` and `self.active_loader`? It may not be necessary to assign a new attribute here.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -29,6 +30,12 @@ def __init__(self,
         self.warmup = warmup
         self.warmup_iters = warmup_iters
         self.warmup_ratio = warmup_ratio
+        self.warmup_by_epoch = warmup_by_epoch
+
+        if self.warmup_by_epoch:
+            self.warmup_epochs = self.warmup_iters
+            self.warmup_iters = None
```
hellock: Is it necessary?

mmcv/runner/runner.py
kennymckormick: Sorry, didn't see that. My bad.

PR #213, opened 2020-03-17
add bias_init_with_prob

xvjiarui:
This PR added `bias_init_with_prob` function to init conv/fc layer in cnn.


hellock:
Unit tests are missing.

PR #214, opened 2020-03-18
Resize on loading

innerlee:
Useful for loading large jpegs.

Use case:
If your training images are large (say, 1080p or 720p), and if you only need no more than 256px of height, this can speed up things more than 2x.

*PREVIEW, NOT READY FOR MERGE*


ZwwWayne:
@innerlee , I still think this is beneficial feature, do we want to merge it or just close it?

innerlee:
@ZwwWayne If we can find a real use case, then I think it is worth merging. Do you know anyone that might be interested in using this?

ZwwWayne:
Are the supported scaling factor limited to the current choices?  Can they be more flexible?

innerlee:
Yes options are limited. You can query them here https://github.com/lilohuang/PyTurboJPEG/blob/master/turbojpeg.py#L226. Last time I checked, all multiples of 1/8 are supported, but only part of them are faster (1/2, 1/4 and 1/8 are, at least).

I have another pull request at our internal mmaction2 repo, that can crop image before decoding. That one might be more useful for most tasks, if anyone have interest :)

PR #221, opened 2020-04-03
Add MlflowLoggerHook

daavoo:
This P.R. adds a new class to `mmcv.runner.hooks.logger` for logging metrics and (optionally) a trained model to [MLflow](https://www.mlflow.org/docs/latest/index.html); similar to existing options for pavi, wandb or tensorboard.

An unit test following the existing ones is included.

The `wandb` unit test has been moved from `test_runner `to `test_hooks`.

### DIFF
--- a/mmcv/runner/hooks/logger/mlflow.py
+++ b/mmcv/runner/hooks/logger/mlflow.py
@@ -7,7 +7,7 @@ from .base import LoggerHook
 
 
 @HOOKS.register_module
-class MLflowLoggerHook(LoggerHook):
+class MlflowLoggerHook(LoggerHook):
 
     def __init__(self,
                  experiment_name,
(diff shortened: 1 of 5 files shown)


tests/test_hooks.py
```diff
@@ -3,17 +3,24 @@
...
+try:
+    import torch
+    import torch.nn as nn
+    from torch.utils.data import DataLoader
+except ImportError:
+    warnings.warn('torch is not available')
+    torch = None
+
+only_if_torch_available = pytest.mark.skipif(
+    torch is None, reason='torch is not available')
```
hellock, line 19: PyTorch will be made a compulsive requirement.

mmcv/runner/hooks/logger/mlflow.py
```diff
@@ -0,0 +1,78 @@
...
+from mmcv.runner import master_only
+from ..hook import HOOKS
+from .base import LoggerHook
+
+
+@HOOKS.register_module
+class MlflowLoggerHook(LoggerHook):
+
+    def __init__(self,
+                 experiment_name=None,
```
hellock, line 13: Maybe simply `exp_name`?

daavoo: Done!

tests/test_hooks.py
daavoo: Decorators have been removed

PR #223, opened 2020-04-06
add contains to registry

JoannaLXY:
Signed-off-by: lixuanyi <lixuanyi@sensetime.com>


mmcv/utils/registry.py
```diff
@@ -42,6 +42,9 @@ def get(self, key):
         """
         return self._module_dict.get(key, None)
 
+    def __contains__(self, key):
+        return self.get(key) is not None
```
hellock, line 46: Move the magic method to the top (after __len__).

PR #240, opened 2020-04-23
Support optimizers with no momentum

Mxbonn:
Similar to https://github.com/open-mmlab/mmdetection/pull/2509, while most optimizers will have momentum or betas, not all have them.

### DIFF
--- a/mmcv/runner/runner.py
+++ b/mmcv/runner/runner.py
@@ -206,10 +206,10 @@ class Runner(object):
         """
         if self.optimizer is None:
             raise RuntimeError(
-                'lr is not applicable because optimizer does not exist.')
+                'momentum is not applicable because optimizer does not exist.')
         return [
-            group['momentum']
-            if 'momentum' in group.keys() else group['betas'][0]
+            group['momentum'] if 'momentum' in group.keys() else
+            group['betas'][0] if 'betas' in group.keys() else 0
             for group in self.optimizer.param_groups
         ]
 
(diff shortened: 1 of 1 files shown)


mmcv/runner/runner.py
```diff
@@ -206,10 +206,10 @@ def current_momentum(self):
         """
         if self.optimizer is None:
             raise RuntimeError(
-                'lr is not applicable because optimizer does not exist.')
+                'momentum is not applicable because optimizer does not exist.')
         return [
-            group['momentum']
-            if 'momentum' in group.keys() else group['betas'][0]
+            group['momentum'] if 'momentum' in group.keys() else
```
ZwwWayne, line 211: It might be better to move the logic before `return`. The current version is a little bit hard to read.

Mxbonn: Not sure if moving the logic before return will make it easier to read as you will need extra logic to get the `momentum` and `betas` keys as param_groups might be an empty array?

The current implementation has the same logic as before, while complex else if structures can make things hard to read, this additional if-else is formatted in exactly the same way and about the same kind of logic.

ZwwWayne: But adding a new if-else increases the complexity. So the following code should be much easier to read.
```python
momentums = []
for group in self.optimizer.param_groups:
    if 'momentum' in group.keys():
        momentums.append(group['momentum'])
    elif 'betas' in group.keys():
        momentums.append(group['betas'][0])
    else:
        momentums.append(0)
return momentums
```

PR #251, opened 2020-04-27
Support specifying names for registry.register_module()

hellock:
Old API:
```python
CATS = mmcv.Registry('cat')

@CATS.register_module
class BritishShorthair:
    pass
```

New API:
```python
CATS = mmcv.Registry('cat')

@CATS.register_module()
class BritishShorthair:
    pass

@CATS.register_module(name='CuteCat')
class Munchkin:
    pass
```

### DIFF
--- a/mmcv/utils/registry.py
+++ b/mmcv/utils/registry.py
@@ -1,4 +1,5 @@
 import inspect
+import warnings
 from functools import partial
 
 from .misc import is_str
@@ -24,7 +25,7 @@ class Registry(object):
     def __repr__(self):
         format_str = self.__class__.__name__ + \
                      f'(name={self._name}, ' \
-                     f'items={list(self._module_dict.keys())})'
+                     f'items={self._module_dict})'
         return format_str
 
     @property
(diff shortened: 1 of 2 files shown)


mmcv/utils/registry.py
```diff
@@ -46,44 +47,82 @@ def get(self, key):
...
+            >>>     pass
+
+            >>> backbones = Registry('backbone')
+            >>> class ResNet:
             >>>     pass
             >>> backbones.register_module(ResNet)
 
         Args:
-            module (:obj:`nn.Module`): Module to be registered.
+            cls (type): Module class to be registered.
```
xvjiarui, line 99: It does not match the arguments.

mmcv/utils/registry.py
```diff
@@ -46,44 +47,82 @@ def get(self, key):
         """
         return self._module_dict.get(key, None)
 
-    def _register_module(self, module_class, force=False):
+    def _register_module(self, module_class, module_name=None, force=False):
         if not inspect.isclass(module_class):
```
xvjiarui, line 51: Not tested.

xvjiarui:
We may also add the PR comments into docs.

PR #256, opened 2020-04-29
Fix Deadlock fix

### LINKED ISSUE #255: Bug: Deadlock 'fix' introduced deadlock.
Since https://github.com/open-mmlab/mmcv/pull/252 is merged my mmdetection code hangs after evaluation. After reverting the specific commit `git revert c203419f57c2e25ab4307420b9a3688f99e01dea`, my code runs again as expected..

Mxbonn:
Resolves #255

### DIFF
--- a/mmcv/runner/hooks/logger/text.py
+++ b/mmcv/runner/hooks/logger/text.py
@@ -98,14 +98,15 @@ class TextLoggerHook(LoggerHook):
         log_dict['iter'] = runner.inner_iter + 1
         # only record lr of the first param group
         log_dict['lr'] = runner.current_lr()[0]
-        memory = None
-        if torch.cuda.is_available():
-            memory = self._get_max_memory(runner)
         if mode == 'train':
             log_dict['time'] = runner.log_buffer.output['time']
             log_dict['data_time'] = runner.log_buffer.output['data_time']
 
             # statistic memory
+            if torch.cuda.is_available():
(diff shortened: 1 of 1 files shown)


Mxbonn:
@hejm37

hejm37:
I am sorry to introduce the bug in #252. The reason I am not aware of the bug is that I am not using the default runner in [runner.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/runner.py). In my own use case, all processes call [`log()`](https://github.com/open-mmlab/mmcv/blob/d27f1f05a842335b4c3e8de07b452c79c862206c/mmcv/runner/hooks/logger/text.py#L118) while only process 0 will call it in the default setting of mmdetection.

Revert PR #252 shall fix issue #255, @hellock. I will fix the bug in #252 in out own runner.

mmcv/runner/hooks/logger/text.py
```diff
@@ -98,14 +98,15 @@ def log(self, runner):
...
-            memory = self._get_max_memory(runner)
         if mode == 'train':
             log_dict['time'] = runner.log_buffer.output['time']
             log_dict['data_time'] = runner.log_buffer.output['data_time']
 
             # statistic memory
+            if torch.cuda.is_available():
+                memory = self._get_max_memory(runner)
+            else:
+                memory = None
```
hellock, line 109: You may just revert to version before #252.

PR #261, opened 2020-05-01
Fixed config import in config file

### LINKED ISSUE #248: call Config.fromfile in a config file
If we want to define a config based on another config, the current version seems to raise AttributeError. Here is an example to reproduce the error.
`./config_a.py`
```
a = 1
b = 1
```
`./config_b.py`
```
from mmcv import Config
cfg = Config.fromfile('./config_a.py')
a = cfg.a + cfg.b
```
`./test.py`
```
from mmcv import Config
cfg = Config.fromfile('./config_b.py')
# do something according to cfg.a
```
When run with `python test.py`, it throws the following message.
```
  File "test.py", line 3, in <module>
    cfg = Config.fromfile('./config_b.py')
  File "/home/yanglei/anaconda3/lib/python3.6/site-packages/mmcv/utils/config.py", line 152, in fromfile
    cfg_dict, cfg_text = Config._file2dict(filename)
  File "/home/yanglei/anaconda3/lib/python3.6/site-packages/mmcv/utils/config.py", line 91, in _file2dict
    mod = import_module('_tempconfig')
  File "/home/yanglei/anaconda3/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/tmp/tmpr6yo3isa/_tempconfig.py", line 3, in <module>
  File "/home/yanglei/anaconda3/lib/python3.6/site-packages/mmcv/utils/config.py", line 264, in __getattr__
    return getattr(self._cfg_dict, name)
  File "/home/yanglei/anaconda3/lib/python3.6/site-packages/mmcv/utils/config.py", line 34, in __getattr__
    raise ex
AttributeError: 'ConfigDict' object has no attribute 'a'
```
The issue may lie in https://github.com/open-mmlab/mmcv/blob/master/mmcv/utils/config.py#L89,
where a fixed temp name is used. Although `tempfile.TemporaryDirectory` creates different directories, the `sys.path` works as a global variable.

When we call `Config.fromfile` in `test.py`, `/tmp/tmpxxx1` is inserted to `sys.path`. When it comes to `Config.fromfile` in `config_b.py`, `/tmp/tmpxxx2` is inserted to `sys.path`. It may confuse `import_module('_tempconfig')` as `'_tempconfig.py'` exists under both directories.

A possible solution may use `osp.basename(filename)` to replace `'_tempconfig'`. But it is still likely that users call `Config.fromfile('./xx/config_b.py')` in `./config_b.py`, leading to a name conflict in `import_module`.

Is there any other better ways to solve it?

xvjiarui:
Fixed #248

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -85,10 +85,14 @@ class Config(object):
         check_file_exist(filename)
         if filename.endswith('.py'):
             with tempfile.TemporaryDirectory() as temp_config_dir:
+                temp_config_file = tempfile.NamedTemporaryFile(
+                    dir=temp_config_dir, suffix='.py')
+                temp_config_name = osp.basename(temp_config_file.name)
                 shutil.copyfile(filename,
-                                osp.join(temp_config_dir, '_tempconfig.py'))
+                                osp.join(temp_config_dir, temp_config_name))
+                temp_module_name = osp.splitext(temp_config_name)[0]
                 sys.path.insert(0, temp_config_dir)
-                mod = import_module('_tempconfig')
+                mod = import_module(temp_module_name)
                 sys.path.pop(0)
                 cfg_dict = {
                     name: value
@@ -96,7 +100,9 @@ class Config(object):
                     if not name.startswith('__')
                 }
                 # delete imported module
-                del sys.modules['_tempconfig']
+                del sys.modules[temp_module_name]
+                # close temp file
+                temp_config_file.close()
         elif filename.endswith(('.yml', '.yaml', '.json')):
             import mmcv
             cfg_dict = mmcv.load(filename)
--- a/tests/data/config/code.py
+++ b/tests/data/config/code.py
@@ -1,3 +1,3 @@
-from mmcv import Config
+from mmcv import Config  # isort:skip
 cfg = Config.fromfile('./tests/data/config/a.py')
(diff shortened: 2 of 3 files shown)


mmcv/utils/config.py
```diff
@@ -85,18 +85,24 @@ def _file2dict(filename):
         check_file_exist(filename)
         if filename.endswith('.py'):
             with tempfile.TemporaryDirectory() as temp_config_dir:
+                temp_config_file = tempfile.NamedTemporaryFile(
+                    dir=temp_config_dir, suffix='.py')
```
hellock, line 89: If we use random filenames, the temp dir may not be necessary. We can just use the default temp dir.

xvjiarui: We had better avoid to insert the default `tmp` folder into `sys.path.`.

Issue #264, opened 2020-05-02
CosineAnealingLrUpdaterHook can never be used due to title()


c1aris:
Calling `hook_type = lr_config.pop('policy').title() + 'LrUpdaterHook'` makes it impossible to construct a `CosineAnealingLrUpdaterHook` because `lr_config.pop('policy').title()` will always be `Cosineanealing`.

Maybe falling back to `CosineLrUpdaterHook` is a good idea.

PR #265, opened 2020-05-03
Fix CosineAnealingLr register bug

ZwwWayne:
Fix https://github.com/open-mmlab/mmcv/issues/264.

### DIFF
--- a/mmcv/runner/runner.py
+++ b/mmcv/runner/runner.py
@@ -388,7 +388,10 @@ class Runner(object):
     def register_lr_hook(self, lr_config):
         if isinstance(lr_config, dict):
             assert 'policy' in lr_config
-            hook_type = lr_config.pop('policy').title() + 'LrUpdaterHook'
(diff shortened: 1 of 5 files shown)


mmcv/runner/runner.py
```diff
@@ -388,7 +388,10 @@ def run(self, data_loaders, workflow, max_epochs, **kwargs):
     def register_lr_hook(self, lr_config):
         if isinstance(lr_config, dict):
             assert 'policy' in lr_config
-            hook_type = lr_config.pop('policy').title() + 'LrUpdaterHook'
+            policy_type = lr_config.pop('policy')
+            if policy_type == policy_type.lower():
+                policy_type = policy_type.title()
+            hook_type = policy_type + 'LrUpdaterHook'
```
hellock, line 394: Add some comments for the above logic, otherwise users may get confused.

ZwwWayne: Sure

PR #271, opened 2020-05-05
Save/load hooks meta

cybergrind:
This PR is not final and just illustrates one possible way to resolve the issue and probably isn't generic enough.

Right now we can save meta information to the checkpoint, but we cannot extract it because of `Runner` discards it immediately after loading, this requires overriding the default `Runner` with a custom class to handle this situation.
Saving meta information is crucial in some situations (eg. when you want to save the best checkpoint and you want to resume training without saving hook information outside the checkpoint)

Here is example from current code. We have no way to get `checkoint['meta']` without overriding methods in `Runner` or we have to do double loading of checkpoint:

```python
    def resume(self,
               checkpoint,
               resume_optimizer=True,
               map_location='default'):
        if map_location == 'default':
            device_id = torch.cuda.current_device()
            checkpoint = self.load_checkpoint(
                checkpoint,
                map_location=lambda storage, loc: storage.cuda(device_id))
        else:
            checkpoint = self.load_checkpoint(
                checkpoint, map_location=map_location)

        if 'optimizer' in checkpoint and resume_optimizer:
            self.optimizer.load_state_dict(checkpoint['optimizer'])

        self.logger.info('resumed epoch %d, iter %d', self.epoch, self.iter)
```

### DIFF
--- a/mmcv/runner/hooks/checkpoint.py
+++ b/mmcv/runner/hooks/checkpoint.py
@@ -21,7 +21,7 @@ class CheckpointHook(Hook):
         self.args = kwargs
 
     @master_only
-    def after_train_epoch(self, runner):
+    def after_epoch(self, runner):
         if not self.every_n_epochs(runner, self.interval):
             return
 
--- a/mmcv/runner/runner.py
+++ b/mmcv/runner/runner.py
@@ -260,6 +260,11 @@ class Runner(object):
         else:
             meta.update(epoch=self.epoch + 1, iter=self.iter)
 
+        for hook in self.hooks:
+            hook_meta = getattr(hook, 'meta', None)
+            if hook_meta:
+                meta['hooks'][hook.__class__.__name__] = hook_meta
+
         filename = filename_tmpl.format(self.epoch + 1)
         filepath = osp.join(out_dir, filename)
         optimizer = self.optimizer if save_optimizer else None
@@ -329,6 +334,13 @@ class Runner(object):
 
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
+
+        hooks_meta = checkpoint['meta'].get('hooks', {})
+        for hook in self.hooks:
+            name = hook.__class__.__name__
+            if name in hooks_meta:
+                hook.meta = hooks_meta[name]
+
         if 'optimizer' in checkpoint and resume_optimizer:
             self.optimizer.load_state_dict(checkpoint['optimizer'])
 


hellock:
Thanks for your suggestions! Dealing with meta info properly is important. Could you also provide some examples to show the usage cases?

cybergrind:
Yeah, sure. Here is how we're storing best loss across training in checkpoint:

```python
class BestCheckpoint(Hook):
    """Store and update best checkpoint after each validation
    """

    def __init__(self, key='loss', minimize=True,
                 checkpoint_extension=CHECKPOINT_EXTENSION,
                 latest_fname=LATEST_CHECKPOINT_NAME,
                 best_fname=BEST_CHECKPOINT_NAME):
        self.key = key  # output dict key to watch
        self.minimize = minimize
        self.latest_fname = latest_fname + checkpoint_extension
        self.best_fname = best_fname + checkpoint_extension
        self.meta = {self.key: None}

    def after_val_epoch(self, runner):
        runner.log_buffer.average()
        target = runner.log_buffer.output[self.key]
        best = self.meta[self.key]

        if best:
            if self.minimize and target >= best:
                return
            if not self.minimize and target <= best:
                return

        self._store(runner, target)

    def _store(self, runner, target):
        latest_path = pathlib.Path(runner.work_dir) / self.latest_fname
        best_path = pathlib.Path(runner.work_dir) / self.best_fname

        if not latest_path.exists():
            runner.logger.warning(f'missing "latest" checkpoint at {latest_path} ')
            return

        if best_path.exists():
            best_path.unlink()
        best_path.symlink_to(latest_path.resolve().relative_to(runner.work_dir))

        if self.meta[self.key]:
            runner.logger.info(f'validation "{self.key}" updated from {self.meta[self.key]}'
                               f' to {target}, stored best checkpoint at {best_path}')
        else:
            runner.logger.info(f'stored best checkpoint at {best_path}')

        self.meta[self.key] = target
```

mmcv/runner/hooks/checkpoint.py
```diff
@@ -21,7 +21,7 @@ def __init__(self,
         self.args = kwargs
 
     @master_only
-    def after_train_epoch(self, runner):
+    def after_epoch(self, runner):
```
hellock, line 24: This may not be expected since the model weights before and after the val epoch are the same. Saving the checkpoint in `after_epoch()` results in duplicated checkpoints.

hellock:
Thanks for the example of saving the best checkpoint. It is a nice idea. This PR looks good to me except that it changes an unrelated behavior.

mmcv/runner/hooks/checkpoint.py
cybergrind: @hellock yeah, I dug deeper and found that whole hooks system isn't working as expected because there is no global `after_epoch` callback that will run after all other handlers were called. We need to decide what to do with #276 and then I will change this PR.

Thank you for your time!

cybergrind:
@hellock it isn't really unrelated because we need some robust mechanism to gather/define meta parameters strictly before saving. It can be implemented as a part of the `Runner`, for example, some kind of `Runner.save_meta/get_meta` But we should have a mechanism that will ensure that:

1. On 1 stage we can define meta parameters for checkpoint
2. Checkpointer hook saves a checkpoint
3. Prohibit changing any meta parameters until next epoch has started

Because we might want to save meta from non-hooks (eg. we might want to save dataset random.seed when doing really long iterations thus when we have checkpoints not only after an epoch is done) probably more generic way will be better.

zhouzaida:
Thanks for your nice suggestions! We have supported saving or loading the meta of hooks. More details at #962

OpenMMLab-Assistant003:
Hi @cybergrind！We are grateful for your efforts in helping improve this open-source project during your personal time. 
Thank you again for your contribution❤ @cybergrind

PR #276, opened 2020-05-11
Runner: add global before and after epoch hook callbacks

cybergrind:
Right now there are no **global** hook callbacks that run after and before an epoch.

`after_epoch` just callback that called after `after_train_epoch` and `after_val_epoch`.
Working on #271 I've realized that I want to implement two hooks: Checkpointer > BestCheckpointer

And get such order of hooks:
1. `BestCheckpointer.after_val_epoch` - compares current loss vs best loss across training and save flag for future + add meta information
2. `Checkpointer.after_epoch` - saves checkpoint with new meta information
3. `BestCheckpointer.after_epoch` - creates a symlink to file from step 2

I didn't find any ways to implement this behavior without introducing new hook's callbacks. 

Any thoughts on this?

PS. Probably we need to use a different name than `true_after_epoch` so I'm open for suggestions

### DIFF
--- a/mmcv/runner/hooks/hook.py
+++ b/mmcv/runner/hooks/hook.py
@@ -24,6 +24,12 @@ class Hook(object):
     def after_iter(self, runner):
         pass
 
+    def true_before_epoch(self, runner):
+        pass
+
+    def true_after_epoch(self, runner):
+        pass
+
     def before_train_epoch(self, runner):
         self.before_epoch(runner)
 
--- a/mmcv/runner/runner.py
+++ b/mmcv/runner/runner.py
@@ -364,6 +364,7 @@ class Runner(object):
         self.call_hook('before_run')
 
         while self.epoch < max_epochs:
+            self.call_hook('true_before_epoch')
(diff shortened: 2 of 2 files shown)


cybergrind:
hey @hellock. Is there anything that I can do to help expedite this PR?

zhouzaida:
hi @cybergrind , sorry for the late reply. MMCV has supported saving best checkpoint via [EvalHook](https://github.com/open-mmlab/mmcv/blob/d3b057240829a0662125d5e83e002b33cd3fa874/mmcv/runner/hooks/evaluation.py#L33).
Do you have any suggestions to improve the Hook?

cybergrind:
@zhouzaida yeah. it is fine to close this PR.

EvalHook still isn't what do we want, because it basically forces us to rerun validation (cause we're using validation results as the measure for the best checkpoint). We just need to be able to save checkpoints after validation (+ store meta information about validation to the checkpoint). It is solved with custom runner and custom hooks, so we're fine

PR #278, opened 2020-05-13
Support conv layers' own `init_weights` method

dreamerlin:


hellock:
Unit tests are missing.

mmcv/cnn/bricks/conv_module.py
```diff
@@ -150,14 +150,16 @@ def norm(self):
...
-            nonlinearity = 'leaky_relu'
-            a = self.act_cfg.get('negative_slope', 0.01)
-        else:
-            nonlinearity = 'relu'
-            a = 0
-
-        kaiming_init(self.conv, a=a, nonlinearity=nonlinearity)
+        # If convolution layer has its `init_weights` method,
+        # it should have initialized weights when building it.
+        if not hasattr(self.conv, 'init_weights'):
```
ZwwWayne, line 155: It seems that in PyTorch, conv layers use `reset_parameters` to initialize weights rather than `init_weights`
https://pytorch.org/docs/stable/_modules/torch/nn/modules/conv.html#Conv1d

dreamerlin: > It seems that in PyTorch, conv layers use `reset_parameters` to initialize weights rather than `init_weights`
> https://pytorch.org/docs/stable/_modules/torch/nn/modules/conv.html#Conv1d

This change is for some custom conv modules like [conv2plus1d](http://gitlab.sz.sensetime.com/open-mmlab/mmaction-lite/blob/master/mmaction/models/common/conv2plus1d.py) to use it own `init_weights`.

dreamerlin: This changed is for some custom conv modules which use their own `init_weights`.

ZwwWayne: Yeah, I get it. And it will initialize pytorch's conv layers anyway. Could you make the comments more specific and point out 
1. It will initialize pytorch's conv layers anyway
2. It is mainly for customized conv modules with its own initialization manner. And we do not want ConvModule to overrides the initialization.

PR #290, opened 2020-05-20
Add a BaseRunner and rename Runner to EpochBasedRunner

hellock:


mmcv/runner/epoch_based_runner.py
```diff
@@ -0,0 +1,168 @@
...
+        self.call_hook('before_train_epoch')
+        for i, data_batch in enumerate(data_loader):
+            self._inner_iter = i
+            self.call_hook('before_train_iter')
+            if self.batch_processor is None:
+                outputs = self.model.train_step(data_batch, self.optimizer,
+                                                **kwargs)
+            else:
+                outputs = self.batch_processor(
+                    self.model, data_batch, train_mode=True, **kwargs)
```
xvjiarui, line 34: When both `batch_processor` and `train_step` is implemented, `train_step` will be overwritten by `batch_processor`, which is not ideal.

PR #291, opened 2020-05-21
add PetrelBackend

xvjiarui:


mmcv/fileio/file_client.py
```diff
@@ -20,18 +21,64 @@ def get_text(self, filepath):
...
+        path_maps (dict|None): path mapping dict from local path to Petrel
+            path. When `path_maps={'src': 'dst'}`, `src` in `filepath` will be
+            replaced by `dst`. Default: None.
+    """
+
+    def __init__(self, path_maps=None):
         try:
             import ceph
+            warnings.warn('Ceph is deprecate in favor of Petrel.',
+                          DeprecationWarning)
```
hellock, line 36: DeprecationWarning seems to be ignored by default. You may just use the default warning type.

mmcv/fileio/file_client.py
```diff
@@ -20,18 +21,64 @@ def get_text(self, filepath):
...
+    """Ceph storage backend.
 
-    def __init__(self):
+    Args:
+        path_maps (dict|None): path mapping dict from local path to Petrel
+            path. When `path_maps={'src': 'dst'}`, `src` in `filepath` will be
+            replaced by `dst`. Default: None.
+    """
+
+    def __init__(self, path_maps=None):
```
hellock, line 32: `path_mapping` looks better.

PR #296, opened 2020-05-23
Fix type in imnormalize

mmeendez8:
Fix  bug described in https://github.com/open-mmlab/mmdetection/issues/2789


mmcv/image/photometric.py
```diff
@@ -14,7 +14,7 @@ def imnormalize(img, mean, std, to_rgb=True):
     Returns:
         ndarray: The normalized image.
     """
-    img = np.float32(img) if img.dtype != np.float32 else img.copy()
+    img = np.float32(img.copy())
```
hellock, line 17: `img = img.copy().astype(np.float32)`

hellock:
It seems the commit history is a little messy. You may checkout a branch from the latest master branch.

PR #230, opened 2020-04-14
Support default/external json for open-mmlab models

xvjiarui:
This PR supported to specify open-mmlab models by json file. 
To override the default json file, first export `$MMCV_HOME=/your/path/`, then place a new json file under `$MMCV_HOME/urls`.

PR #301, opened 2020-05-27
BC of model zoo: add deprecate urls

xvjiarui:
This PR added backward compatibility of #230 .


mmcv/model_zoo/deprecate.json
```diff
@@ -0,0 +1,6 @@
+{
```
hellock, line 1: `deprecated.json` should be better.

PR #304, opened 2020-05-29
Use mapping rather than dict for special keys

ZwwWayne:
This PR support to use mapping rather than dict when the key name is not a valid identifier.

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -242,7 +242,8 @@ class Config(object):
                     f'dict({_indent(_format_dict(v_), indent)}),'
                     for v_ in v).rstrip(',')
                 if use_mapping:
-                    attr_str = f'{str(k)}: {v_str}'
+                    k_str = f"'{k}'" if isinstance(k, str) else str(k)
+                    attr_str = f'{k_str}: {v_str}'
                 else:
                     attr_str = f'{str(k)}={v_str}'
                 attr_str = _indent(attr_str, indent) + ']'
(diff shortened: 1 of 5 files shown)


mmcv/utils/config.py
```diff
@@ -219,46 +219,70 @@ def _indent(s_, num_spaces):
...
+            use_mapping = _contain_invalid_identifier(input_dict)
+            if use_mapping:
+                r += '{'
+            for idx, (k, v) in enumerate(input_dict.items()):
+                is_last = idx >= len(input_dict) - 1
                 end = '' if outest_level or is_last else ','
                 if isinstance(v, dict):
                     v_str = '\n' + _format_dict(v)
-                    attr_str = f'{str(k)}=dict({v_str}'
+                    if use_mapping:
```
xvjiarui, line 272: Add `a={'x/x': {'a.0': 233}}` to test this.

mmcv/utils/config.py
```diff
@@ -219,46 +219,70 @@ def _indent(s_, num_spaces):
...
+        def _format_list(k, v, use_mapping=False):
             # check if all items in the list are dict
             if all(isinstance(_, dict) for _ in v):
                 v_str = '[\n'
                 v_str += '\n'.join(
                     f'dict({_indent(_format_dict(v_), indent)}),'
                     for v_ in v).rstrip(',')
-                attr_str = f'{str(k)}={v_str}'
+                if use_mapping:
+                    attr_str = f'{str(k)}: {v_str}'
```
xvjiarui, line 245: Add `a={'x/x': [{'a.0': 1., 'b.0': 2.}, {'c.1': 3.}]}` to test this.

PR #305, opened 2020-05-29
Support scale_rate in VideoReader

loveunk:


hellock:
Thanks for your contribution.
I do not suggest adding a scale rate in `VideoReader`. The reader should be decoupled with any 
post processing like resizing, flipping, etc, and it will not be faster if we adding the resizing into the reader.

You may implement such a reader in your own code.
```python
class ScaledVideoReader(VideoReader):

    def __init__(self, *args, scale_rate=1.0, **kwargs):
        super().__init__(*args, **kwargs)
        self._scale_rate = scale_rate
        if (scale_rate != 1.0):
            self._width = int(round(self._width * scale_rate))
            self._height = int(round(self._height * scale_rate))

    def read(self):
        img = super().read()
        if self._scale_rate != 1.0:
            img = imresize(img, (self._width, self._height))
        return img
```

loveunk:
@hellock Thank you for your comments and suggestion. I will close this PR.

PR #306, opened 2020-05-29
Possible deadlock prevention during epoch transition

Johnson-Wang:
Add sleep(2) during epoch transition


gabmis:
Hello, I'm looking into a simillar behaviour in my own code, did you ever find out what caused this deadlock? Do you know if this sleep mechanism is still necessary today? Any information on this topic would be greatly appreciated :)
Best,
Gabriel

Johnson-Wang:
Sorry, I just came up with this workaround and found this method magically worked. The deadlock could possibly be caused by the delayed release of several workers during epoch transition.

PR #312, opened 2020-06-02
Fix deprecation warning collections.abc

mmeendez8:
## Description
You are probably aware, but when training I am getting a few deprecation warnings:
`mmcv/mmcv/parallel/collate.py:23: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working`

These imports should be updated:
https://docs.python.org/3.8/library/collections.abc.html#collections.abc.Sequence

The note in 3.7 docs:

> Changed in version 3.3: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

### DIFF
--- a/mmcv/parallel/collate.py
+++ b/mmcv/parallel/collate.py
@@ -1,5 +1,6 @@
 # Copyright (c) Open-MMLab. All rights reserved.
-from collections.abc import Sequence, Mapping
+from collections.abc import Mapping, Sequence
+
 import torch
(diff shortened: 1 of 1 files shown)


hellock:
Thanks for the fix!

PR #319, opened 2020-06-05
Add CosineRestartLrUpdaterHook

hejm37:
This merge request adds `CosineRestartLrUpdaterHook` which is commonly used in low level tasks like restoration and generation.

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -292,6 +292,8 @@ def get_position_from_periods(iteration, cumulative_periods):
     for i, period in enumerate(cumulative_periods):
         if iteration <= period:
             return i
+    raise ValueError(f'Current iteration {iteration} exceeds '
+                     f'cumulative_periods {cumulative_periods}')
 
 
 @HOOKS.register_module()
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -259,15 +259,15 @@ class CosineRestartLrUpdaterHook(LrUpdaterHook):
         else:
             progress = runner.iter
 
-        idx = get_position_from_periods(progress, self.cumulative_periods)
-        current_weight = self.restart_weights[idx]
-        nearest_restart = 0 if idx == 0 else self.cumulative_periods[idx - 1]
-        current_periods = self.periods[idx]
-
         if self.min_lr_ratio is not None:
             target_lr = base_lr * self.min_lr_ratio
         else:
             target_lr = self.min_lr
+
+        idx = get_position_from_periods(progress, self.cumulative_periods)
+        current_weight = self.restart_weights[idx]
+        nearest_restart = 0 if idx == 0 else self.cumulative_periods[idx - 1]
+        current_periods = self.periods[idx]
 
         alpha = min((progress - nearest_restart) / current_periods, 1)
         return annealing_cos(base_lr, target_lr, alpha, current_weight)
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -221,6 +221,66 @@ class CosineAnealingLrUpdaterHook(LrUpdaterHook):
 
 
 @HOOKS.register_module()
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/lr_updater.py
```diff
@@ -220,6 +220,66 @@ def get_lr(self, runner, base_lr):
...
 
+@HOOKS.register_module()
+class CosineRestartLrUpdaterHook(LrUpdaterHook):
+    """Cosine annealing with restarts learning rate scheme.
+
+    Args:
+        period (list[int]): Period for each cosine anneling cycle.
+        restart_weights (list[float]): Restart weights at each restart
+            iteration. Default: [1].
+        eta_min (float): The mimimum lr. Default: 0.
```
ZwwWayne, line 231: 1. May use min_lr and min_lr_ratio
2. May support `min_lr_ratio` to support relative LR schedule rather than absolute LR.

hejm37: Updated according to the comment.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -220,6 +220,78 @@ def get_lr(self, runner, base_lr):
...
 
+@HOOKS.register_module()
+class CosineRestartLrUpdaterHook(LrUpdaterHook):
+    """Cosine annealing with restarts learning rate scheme.
+
+    Args:
+        periods (list[int]): Periods for each cosine anneling cycle.
+        restart_weights (list[float]): Restart weights at each restart
+            iteration. Default: [1].
+        min_lr (float): The mimimum lr. Default: 0.
```
ZwwWayne, line 231: Change docstrings according to the changed arguments.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -220,6 +220,78 @@ def get_lr(self, runner, base_lr):
...
+        nearest_restart = 0 if idx == 0 else self.cumulative_periods[idx - 1]
+        current_periods = self.periods[idx]
+
+        if self.min_lr_ratio is not None:
+            target_lr = base_lr * self.min_lr_ratio
+        else:
+            target_lr = self.min_lr
+
+        alpha = min((progress - nearest_restart) / current_periods, 1)
+        return target_lr + current_weight * 0.5 * (base_lr - target_lr) * (
```
ZwwWayne, line 270: Already supported a function `annealing_cos` in line 304.

hejm37: I've noticed the `annealing_cos`. However, if I modify the function to 

```python
def annealing_cos(start, end, factor, weight=1):
    """Cosine anneal from `weight * (start - end) + end` to `end` as pct goes from 0.0 to 1.0."""
    cos_out = cos(pi * factor) + 1
    return end + 0.5 * weight * (start - end) * cos_out
```

The `weight * (start - end) + end` to `end` does not make much sense to me.

hejm37: All right. It in fact makes sense.

hejm37: I've updated the code~

mmcv/runner/hooks/lr_updater.py
```diff
@@ -295,7 +369,9 @@ def get_lr(self, runner, base_lr):
                                      progress / (end_iter - start_iter))
 
 
-def annealing_cos(start, end, factor):
-    """Cosine anneal from `start` to `end` as pct goes from 0.0 to 1.0."""
+def annealing_cos(start, end, factor, weight=1):
+    """Cosine anneal from `weight * start + (1 - weight) * end` to `end` as
```
ZwwWayne, line 373: We can use a standard docstring for this function now.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -220,6 +220,80 @@ def get_lr(self, runner, base_lr):
         return annealing_cos(base_lr, target_lr, progress / max_progress)
 
 
+@HOOKS.register_module()
+class CosineRestartLrUpdaterHook(LrUpdaterHook):
```
ZwwWayne, line 224: Can we add a unit test to test this hook?

PR #326, opened 2020-06-08
Add grad norm to the logger

Johnson-Wang:
If the users specify grad_clip in the config, it's better to monitor the grad_norm so that one can judge if the training is stable or not from the grad_norm.

### DIFF
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -14,11 +14,15 @@ class OptimizerHook(Hook):
         params = list(
             filter(lambda p: p.requires_grad and p.grad is not None, params))
         if len(params) > 0:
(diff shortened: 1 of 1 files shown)


hellock:
Is there a screenshot to show what the log looks like after applying it?

Johnson-Wang:
If the grad_clip is not specified, the output is the same with before. 
If grad_clip is specified, the output is like
`2020-06-08 16:36:52,839 - mmdet - INFO - Epoch [1][25/612]	lr: 0.00000, eta: 0:20:40, time: 1.515, data_time: 0.671, memory: 8436, loss_cls: 0.3105, loss_bbox: 0.6498, num_pos: 4.8125, accuracy: 1.0000, loss: 0.9603, grad_norm: 4.8798`

PR #331, opened 2020-06-11
Add `_NonLocalNd` module

dreamerlin:
This pull request adds "NonLocal" modules, which contains `NonLocal1d`, `NonLocal2d`, `NonLocal3d`, `_NonLocalNd`.


mmcv/cnn/__init__.py
```diff
@@ -1,7 +1,8 @@
 # Copyright (c) Open-MMLab. All rights reserved.
 from .alexnet import AlexNet
 from .bricks import (ACTIVATION_LAYERS, CONV_LAYERS, NORM_LAYERS,
-                     PADDING_LAYERS, UPSAMPLE_LAYERS, ConvModule, Scale,
+                     PADDING_LAYERS, UPSAMPLE_LAYERS, ConvModule, NonLocal1d,
+                     NonLocal2d, NonLocal3d, Scale, _NonLocalNd,
```
OceanPang, line 5: We can remove `_NonLocalNd` from the init file.

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self,
+                 in_channels,
+                 sub_sample=False,
+                 conv_cfg=dict(type='Conv1d'),
+                 **kwargs):
+        super(NonLocal1d, self).__init__(
+            in_channels, conv_cfg=conv_cfg, **kwargs)
+
+        self.sub_sample = sub_sample
+        max_pool_layer = nn.MaxPool1d(kernel_size=2)
```
OceanPang, line 177: We can define the `max_pool_layer` only if we use sub_sample

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+        self.use_scale = use_scale
+        self.inter_channels = in_channels // reduction
+        self.mode = mode
+
+        if mode not in ['embedded_gaussian', 'dot_product']:
+            raise ValueError(
+                "Mode should be in 'embedded_gaussian' or 'dot_product', "
+                f'but got {mode} instead.')
+
+        # g, theta, phi are defaulted as `nn.Conv2d`.
```
xvjiarui, line 51: `ConvNd`

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+    """Basic Non-local module.
+
+    This module is proposed in
+    "Non-local Neural Networks"
+    Paper reference: https://arxiv.org/abs/1711.07971
+
+    Args:
+        in_channels (int): Channels of the input feature map.
+        reduction (int): Channel reduction ratio. Default: 2.
+        use_scale (bool): Whether to scale pairwise_weight by 1/inter_channels.
```
xvjiarui, line 20: `1/sqrt(inter_channels)`

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.init_weights(**kwargs)
+
+    def init_weights(self, std=0.01, zeros_init=True):
+        for m in [self.g, self.theta, self.phi]:
+            normal_init(m.conv, std=std)
+        if zeros_init:
+            constant_init(self.conv_out.conv, 0)
+            if self.conv_out.norm_cfg is not None:
+                constant_init(self.conv_out.norm, 0)
```
xvjiarui, line 87: This is overriding the default init of `ConvModule`. 
So when `norm_cfg is not None`, both of `self.conv_out.conv` and `self.conv_out.norm` will be zero limited. 
Maybe we should zero init `self.conv_out.conv` only when `self.conv_out.norm is None`.

mmcv/cnn/bricks/non_local.py
xvjiarui, line 20: when the mode is `embedded_gaussian`.

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+        pairwise_weight = pairwise_func(theta_x, phi_x)
+
+        # NonLocal1d y: [N, H, C]
+        # NonLocal2d y: [N, HxW, C]
+        # NonLocal3d y: [N, TxHxW, C]
+        y = torch.matmul(pairwise_weight, g_x)
+        # NonLocal1d y: [N, C, H]
+        # NonLocal2d y: [N, C, H, W]
+        # NonLocal3d y: [N, C, T, H, W]
+        y = y.permute(0, 2, 1).reshape(n, self.inter_channels, *x.size()[2:])
```
xvjiarui, line 150: `y = y.permute(0, 2, 1).contiguous().reshape(n, self.inter_channels, *x.size()[2:])` will avoid bug in PyTorch in 1.4.

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,234 @@
...
+            self.g = nn.Sequential(self.g, max_pool_layer)
+            self.phi = nn.Sequential(self.phi, max_pool_layer)
+
+
+class NonLocal3d(_NonLocalNd):
+    """3D Non-local module.
+
+    Args:
+        in_channels (int): Same as `NonLocalND`.
+        sub_sample (bool): Whether to apply max pooling after pairwise
```
xvjiarui, line 216: We may add `Note that the sub_sample is applied on spatial only. `

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def init_weights(self, std=0.01, zeros_init=True):
+        for m in [self.g, self.theta, self.phi]:
+            normal_init(m.conv, std=std)
+        if zeros_init and self.conv_out.norm_cfg is None:
+            constant_init(self.conv_out.conv, 0)
+        else:
+            normal_init(self.conv_out.conv, std=std)
+            if self.conv_out.norm_cfg is not None:
+                normal_init(self.conv_out.conv, std=std)
```
OceanPang, line 90: This init function still has problems. 
The `if` under `else` seems unnecessary. We also need to zero init when there is a `norm` in the module.

mmcv/cnn/bricks/non_local.py
```diff
@@ -0,0 +1,235 @@
...
+            conv_cfg=conv_cfg,
+            norm_cfg=norm_cfg,
+            act_cfg=None)
+
+        self.init_weights(**kwargs)
+
+    def init_weights(self, std=0.01, zeros_init=True):
+        for m in [self.g, self.theta, self.phi]:
+            normal_init(m.conv, std=std)
+        if zeros_init and self.conv_out.norm_cfg is None:
```
xvjiarui, line 85: https://github.com/open-mmlab/mmcv/pull/331#discussion_r441386988 not fully resolved.
```python
if zeros_init:
    if self.conv_out.norm_cfg is None: 
         constant_init(self.conv_out.conv, 0)
    else:
        constant_init(self.conv_out.norm, 0)
``` should be good.

File: mmcv/cnn/bricks/non_local.py [hunk 10]
```diff
@@ -0,0 +1,241 @@
...
+        # NonLocal3d pairwise_weight: [N, TxHxW, TxHxW]
+        pairwise_weight = torch.matmul(theta_x, phi_x)
+        if self.use_scale:
+            # theta_x.shape[-1] is `self.inter_channels`
+            pairwise_weight /= theta_x.shape[-1]**0.5
+        pairwise_weight = pairwise_weight.softmax(dim=-1)
+        return pairwise_weight
+
+    @staticmethod
+    def dot_product(theta_x, phi_x):
```
hellock, line 108: `dot_product` and `embedded_gaussian` have equal status. Using staticmethod for one and normal method for another is not good.

PR #333, opened 2020-06-12
added pillow backend in loading

Johnson-Wang:


mmcv/image/io.py
```diff
@@ -60,6 +69,52 @@ def _jpegflag(flag='color', channel_order='bgr'):
...
+            array[:, :, :3] = array[:, :, (2, 1, 0)]  # RGB to BGR
+    else:
+        if img.mode == 'LA':
+            # When the mode is 'LA', the default conversion will fill in the
+            #  canvas with black, which sometimes shadows black objects in
+            #  the foreground.
+            #
+            # Therefore, a random color (124, 117, 104) is used for canvas
+            img = img.convert('RGBA')
+            canvas = Image.new('RGB', img.size, (124, 117, 104))
```
ZwwWayne, line 100: (124, 117, 104) seems to be mean of images in RGB order. Is it OK if the image if in BGR order?

tests/test_image/test_io.py
```diff
@@ -65,6 +65,51 @@ def test_imread(self):
...
+        mean = img_pil_gray_alpha[300:, 400:].mean(axis=(0, 1))
+        assert_allclose(img_pil_gray_alpha[300:, 400:] - mean, 0)
+        assert img_pil_gray_alpha.shape == (400, 500, 3)
+        img_pil_gray_alpha = mmcv.imread(self.gray_alpha_img_path, 'unchanged')
+        assert img_pil_gray_alpha.shape == (400, 500, 2)
+        img_pil_palette = mmcv.imread(self.palette_img_path, 'grayscale')
+        assert img_pil_palette.shape == (300, 400)
+        img_pil_palette = mmcv.imread(self.palette_img_path)
+        assert img_pil_palette.shape == (300, 400, 3)
+        img_pil_palette = mmcv.imread(self.palette_img_path, 'unchanged')
```
xvjiarui, line 88: Is it equivalent to testing a static image with the palette, like that annotation for segmentation?

Johnson-Wang: Yes. The purpose here is to test image with palette format. But I did not find test data in this format, so I created one.

mmcv/image/io.py
```diff
@@ -60,6 +69,52 @@ def _jpegflag(flag='color', channel_order='bgr'):
...
+            canvas.paste(img, mask=img.split()[3])  # 3 is alpha channel
+            img = canvas
+        elif img.mode != 'RGB':
```
hellock, line 103: Put the most used case at the first in if-conditions.

mmcv/image/io.py
```diff
@@ -60,6 +69,52 @@ def _jpegflag(flag='color', channel_order='bgr'):
...
 
 
+def _pillow2array(img, flag='color', channel_order='bgr'):
+    """Convert a pillow image to numpy array
+
+    Args:
+        img (:obj:`PIL.Image.Image`): The image loaded using PIL
+        flag (str): Flags specifying the color type of a loaded image,
+            candidates are `color`, `grayscale` and `unchanged`.
+        channel_order (str): Order of channel, candidates are `bgr` and `rgb`.
```
hellock, line 79: The channel order of the output image array.

mmcv/image/io.py
```diff
@@ -60,6 +69,52 @@ def _jpegflag(flag='color', channel_order='bgr'):
...
+        if flag == 'color':
+            array = np.array(img)
+            if channel_order != 'rgb':
+                array = array[:, :, ::-1]  # RGB to BGR
+        elif flag == 'grayscale':
+            img = img.convert('L')
+            array = np.array(img)
+        else:
+            raise ValueError(
+                'flag must be "color", "grayscale" or "unchanged"')
```
hellock, line 114: f'flag must be "color", "grayscale" or "unchanged", but got {flag}'

PR #339, opened 2020-06-12
Add `train_step` and `val_step` in MMDP.

dreamerlin:
Add `train_step` and `val_step` in MMDP for non-distributed training cases.

### DIFF
--- a/mmcv/parallel/data_parallel.py
+++ b/mmcv/parallel/data_parallel.py
@@ -1,4 +1,6 @@
 # Copyright (c) Open-MMLab. All rights reserved.
+from itertools import chain
+
 from torch.nn.parallel import DataParallel
 
 from .scatter_gather import scatter_kwargs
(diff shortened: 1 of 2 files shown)


mmcv/runner/base_runner.py
```diff
@@ -69,7 +69,11 @@ def __init__(self,
                     'batch_processor and model.train_step()/model.val_step() '
                     'cannot be both available.')
         else:
-            assert hasattr(model, 'train_step')
+            if is_parallel_module(model):
+                _model = model.module
+            else:
+                _model = model
+            assert hasattr(_model, 'train_step')
```
hellock, line 76: - MMDDP has already implemented `train_step()`.
- In the runner, `model.train_step()` is called instead of `model.module.train_step()`.

dreamerlin: MMDP doesn't have `train_step()` and `val_step()` method, which will raise error in some non-distributed training. So I add these two methods in MMDP like those in MMDDP.

mmcv/parallel/data_parallel.py
```diff
@@ -8,3 +9,15 @@ class MMDataParallel(DataParallel):
 
     def scatter(self, inputs, kwargs, device_ids):
         return scatter_kwargs(inputs, kwargs, device_ids, dim=self.dim)
+
+    def train_step(self, *inputs, **kwargs):
+        inputs, kwargs = self.scatter(inputs, kwargs,
+                                      [torch.cuda.current_device()])
+        output = self.module.train_step(*inputs[0], **kwargs[0])
+        return output
```
hellock, line 17: Is the implementation correct?

dreamerlin: Ops, it is wrong. The latest codes have rewritten it.

hellock: It seems still incorrect.

PR #340, opened 2020-06-12
Harmless changes

innerlee:
to please pylint


mmcv/fileio/file_client.py
```diff
@@ -192,7 +192,7 @@ def get_text(self, filepath):
         return value_buf
 
 
-class FileClient(object):
+class FileClient():
```
hellock, line 195: `class FileClient` should be enough.

innerlee: ha

PR #346, opened 2020-06-15
fix bug in using inner iter when by_epcoh==True

nbei:
The new iter-based runner may have some potential bugs in using `inner_iter` and we just make `inner_iter` have correct manner and other related hooks(logger hooks) called with conditions on `iter`.


xvjiarui:
Hi @nbei 
Thanks for the fix. 
Since the `BaseLogger` has `by_epoch` arguments now, we may also add following code [here](https://github.com/open-mmlab/mmcv/blob/9f04477f933db682bfa82176d0d8aeebeeb87425/mmcv/runner/iter_based_runner.py#L223). 
```
if log_config is not None:
    log_config.setdefault('by_epoch', False)
```
Btw, could you please sign up CLA?

PR #347, opened 2020-06-16
Add custom_group to DefaultOptimizerConstrutor

hejm37:


mmcv/runner/optimizer/default_constructor.py
```diff
@@ -115,23 +143,35 @@ def add_params(self, params, module, prefix=''):
...
-                        'weight_decay'] = self.base_wd * norm_decay_mult
-                # depth-wise conv
-                elif is_dwconv:
-                    param_group[
-                        'weight_decay'] = self.base_wd * dwconv_decay_mult
-                # bias lr and decay
-                elif name == 'bias':
+            # if the parameter match one of the custom keys, ignore other rules
+            is_custom = False
+            for key in custom_keys:
```
xvjiarui, line 148: We may sort the `custome_keys` according to its length. 
In this way, the most specific key will be matched first. In other words, `priority = len(key)`. 
In this way, when we have some like 
```python
custom_keys={
       'backbone.conv1': {'lr_mlt': 1.},
       'backbone': {'lt_mult':10}
```
we could set `backbone` `lr_mult` w.o. changing `backbone.conv1` `lr_mult`.

hejm37: You can take a look at the docstring~

tests/test_optimizer.py
```diff
@@ -355,6 +355,85 @@ def test_default_optimizer_constructor():
...
+    groups.append(['sub.conv1.weight', 'sub.conv1.bias'])
+    group_settings.append({
+        'lr': 0.001,
+        'momentum': momentum,
+        'weight_decay': 0,
+    })
+    # group 3
+    groups.append(['sub.gn.weight', 'sub.gn.bias'])
+    group_settings.append({
+        'lr': 0.0001,
```
xvjiarui, line 409: `base_lr * 0.01` should be better.

tests/test_optimizer.py
```diff
@@ -355,6 +355,85 @@ def test_default_optimizer_constructor():
...
+    groups.append(['param1', 'sub.param1'])
+    group_settings.append({
+        'lr': 0.1,
+        'momentum': momentum,
+        'weight_decay': base_wd,
+    })
+    # group 2
+    groups.append(['sub.conv1.weight', 'sub.conv1.bias'])
+    group_settings.append({
+        'lr': 0.001,
```
xvjiarui, line 402: `base_lr * 0.1` should be better.

tests/test_optimizer.py
```diff
@@ -355,6 +355,85 @@ def test_default_optimizer_constructor():
...
+
+    # check params groups
+    param_groups = optimizer.param_groups
+
+    groups = []
+    group_settings = []
+    # group 1
+    groups.append(['param1', 'sub.param1'])
+    group_settings.append({
+        'lr': 0.1,
```
xvjiarui: `base_lr * 10` should better. We may also comment that `param1` is matched based on the length.

PR #348, opened 2020-06-16
Update config to support predefined variables

JoannaLXY:
This PR adds to config to support predefined variables. It will convert the variable in `{{var}}` with its real value.  This can be especially useful when defining `work_dir` that is similar to the filename of the config file.
Currently, it supports four predefined variables: 
**{{fileDirname}}** -  the current opened file's dirname, e.g. `/home/your-username/your-project/folder`
**{{fileBasename}}** - the current opened file's basename, e.g. `file.ext`
**{{fileBasenameNoExtension}}** - the current opened file's basename with no file extension, e.g. `file`
**{{fileExtname}}** -  the current opened file's extension, e.g. `.ext`
These variable names are referred from https://code.visualstudio.com/docs/editor/variables-reference.

An example to illustrate the use case:  
Config file `abc.py`:
``` Python
item1 = 1
item2 = {'a': 0}
item3 = '{{fileBasenameNoExtension}}'
```
The config dict will be converted as `{'item1': 1, 'item2': {'a': 0}, 'item3': 'abc'}`.


mmcv/utils/config.py
```diff
@@ -92,16 +93,38 @@ def _validate_py_syntax(filename):
...
+                    fileBasename=fileBasename,
+                    fileBasenameNoExtension=fileBasenameNoExtension,
+                    fileExtname=fileExtname)
+                config_file = open(filename).read()
+                for key, value in support_templates.items():
+                    regexp = r'\{\{\s*' + str(key) + r'\s*\}\}'
+                    config_file = re.sub(regexp, value, config_file)
+                tmp_config_file = open(temp_config_file.name, 'w')
+                tmp_config_file.write(config_file)
+                tmp_config_file.close()
```
xvjiarui, line 123: We may extract these codes into a function to support future extension.

xvjiarui:
We may add docs about it at [here[(docs/utils.md).

mmcv/utils/config.py
hellock, line 123:
```python
with open(temp_config_file.name, 'w') as tmp_config_file:
    tmp_config_file.write(config_file)
```

mmcv/utils/config.py
```diff
@@ -92,16 +93,43 @@ def _validate_py_syntax(filename):
                               f'file {filename}')
 
     @staticmethod
-    def _file2dict(filename):
+    def _substitute_predefined_vars(filename, temp_config_name):
+        fileDirname = osp.dirname(filename)
+        fileBasename = osp.basename(filename)
+        fileBasenameNoExtension = osp.splitext(fileBasename)[0]
+        fileExtname = osp.splitext(filename)[1]
```
hellock, line 100: Although the predefined variables can be camel case, use snake case for python variables.

mmcv/utils/config.py
```diff
@@ -92,16 +93,43 @@ def _validate_py_syntax(filename):
...
+            raise IOError('Only py/yml/yaml/json type are supported now!')
+
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            temp_config_file = tempfile.NamedTemporaryFile(
+                dir=temp_config_dir, suffix=fileExtname)
+            temp_config_name = osp.basename(temp_config_file.name)
+            # Substitute predefined variables
+            if use_predefined_variables:
+                Config._substitute_predefined_vars(filename,
+                                                   temp_config_file.name)
```
xvjiarui, line 128: `.json` and `.yaml` is not tested for this.

tests/data/config/h.py
```diff
@@ -0,0 +1,3 @@
+item1 = '{{fileBasename}}'
+item2 = '{{fileDirname}}'
```
hellock, line 2: May add a case with whitespaces: `{{ fileDirname }}`

docs/utils.md
```diff
@@ -26,6 +26,39 @@ To load and use configs
...
 
+For all format configs, some predefined variables are supported. It will convert the variable in `{{var}}` with its real value.
+
+Currently, it supports four predefined variables:
+
+`{{fileDirname}}` - the current opened file's dirname, e.g. /home/your-username/your-project/folder
+
+`{{fileBasename}}` - the current opened file's basename, e.g. file.ext
+
+`{{fileBasenameNoExtension}}` - the current opened file's basename with no file extension, e.g. file
```
hellock, line 37: `fileBasenameWithoutExt`

PR #350, opened 2020-06-17
add backend arg

xvjiarui:


mmcv/image/io.py
```diff
@@ -131,10 +131,17 @@ def imread(img_or_path, flag='color', channel_order='bgr'):
             candidates are `color`, `grayscale` and `unchanged`.
             Note that the `turbojpeg` backened does not support `unchanged`.
         channel_order (str): Order of channel, candidates are `bgr` and `rgb`.
+        backend (str|None): The image decoding backend type. Options are `cv2`,
+            `pillow`, `turbojpeg`, `None`. If backend is None, the global
+            imread_backend will be used.
```
Johnson-Wang, line 136: Default value.

mmcv/image/io.py
```diff
@@ -165,22 +172,29 @@ def imread(img_or_path, flag='color', channel_order='bgr'):
...
 
     Returns:
         ndarray: Loaded image array.
     """
-    if imread_backend == 'turbojpeg':
+
+    if backend is None:
+        backend = imread_backend
+    assert backend in supported_backends
```
Johnson-Wang, line 191: raise ValueError instead of using assert.

mmcv/image/io.py
Johnson-Wang, line 136: Change the last line to:
If backend is None, the global imread_backend specified by \`mmcv.use_backend\` will be used.

mmcv/image/io.py
```diff
@@ -131,10 +131,20 @@ def imread(img_or_path, flag='color', channel_order='bgr'):
             candidates are `color`, `grayscale` and `unchanged`.
             Note that the `turbojpeg` backened does not support `unchanged`.
         channel_order (str): Order of channel, candidates are `bgr` and `rgb`.
+        backend (str|None): The image decoding backend type. Options are `cv2`,
+            `pillow`, `turbojpeg`, `None`. If backend is None, the global
+            imread_backend specified by `mmcv.use_backend` will be used.
```
hellock, line 136: \`\`mmcv.use_backend()\`\`

PR #365, opened 2020-06-23
fix bug for building hsigmoid with inplace=True in ConvModule

louzana:


hellock:
You may rebase the branch.

PR #374, opened 2020-06-30
Add tensor2imgs in misc

xvjiarui:


mmcv/image/misc.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import torch
```
hellock, line 2: Add `try except` to make the image module available without torch.

mmcv/image/misc.py
```diff
@@ -0,0 +1,41 @@
...
+        std (tuple[float], optional): Standard deviation of images.
+            Defaults to (1, 1, 1).
+        to_rgb (bool, optional): Whether the tensor was converted to RGB
+            format in the first place. If so, convert it back to BGR.
+            Defaults to True.
+
+    Returns:
+        list[np.ndarray]: A list that contains multiple images.
+    """
+
```
hellock, line 27: Raise an error if torch is None.

PR #375, opened 2020-07-01
Fix multiline string in config

### LINKED ISSUE #324: [BUG] Can not dump config with a multi-line string value
Assume a config "a.py" with content
```
A="""this is
a multi-line
string"""
```
When using cfg.dump(), the following error occurs:
```
File "<unknown>", line 1
    CODE='this is
                ^
SyntaxError: EOL while scanning string literal
```

This bug is introduced by `FormatCode`.

xvjiarui:
Fixed #324


mmcv/utils/config.py
```diff
@@ -227,17 +228,34 @@ def pretty_text(self):
 
         def _indent(s_, num_spaces):
             s = s_.split('\n')
+            # find multiline string interval
+            mls_intervals = []
+            for idx, line in enumerate(s):
+                if '\"\"\"' in line:
```
hellock, line 234: `\` is unnecessary

hellock: Consider `'''text'''`

hiyyg:
Any update on this? I saw the issue was closed, however the bug still exists.

PR #390, opened 2020-07-07
use torch.cummax in corner_pool for torch 1.5+

tianyuandu:
Synchronize from mmdetection.

https://github.com/open-mmlab/mmdetection/pull/3158

### DIFF
--- a/mmcv/ops/corner_pool.py
+++ b/mmcv/ops/corner_pool.py
@@ -113,7 +113,7 @@ class CornerPool(nn.Module):
         self.corner_pool = self.pool_functions[mode]
 
     def forward(self, x):
-        if torch.__version__ >= '1.5.0':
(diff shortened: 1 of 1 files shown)


mmcv/ops/corner_pool.py
```diff
@@ -98,10 +99,27 @@ class CornerPool(nn.Module):
...
+
     def __init__(self, mode):
         super(CornerPool, self).__init__()
         assert mode in self.pool_functions
+        self.mode = mode
         self.corner_pool = self.pool_functions[mode]
 
     def forward(self, x):
-        return self.corner_pool.apply(x)
+        if torch.__version__ >= '1.5.0':
```
hellock, line 116: `torch.__version__ != 'parrots' and torch.__version__ >= '1.5.0'`

PR #391, opened 2020-07-07
Rewrite readme.rst by markdown

ZwwWayne:
This PR fix the rst issue by using `M2R` and support to use README.md.


README.md
```diff
@@ -0,0 +1,73 @@
...
+pip install -e .
+```
+
+If you are on macOS, replace the last command with
+
+```bash
+CC=lang CXX=clang++ CFLAGS='-stdlib=libc++' pip install -e .
+```
+
+We also provide pre-build mmcv with corresponding Pytorch and CUDA versions
```
hellock, line 51: Note that the building process may take up to 10 mins. We also provide pre-built mmcv with corresponding Pytorch and CUDA versions to simplify the building.

PR #396, opened 2020-07-08
Add linear decay learning rate scheduler

xiaojieli0903:
Add linear decay learning rate scheduler, which is used when training shufflenet models.


mmcv/runner/hooks/lr_updater.py
```diff
@@ -413,3 +413,35 @@ def annealing_cos(start, end, factor, weight=1):
     """
     cos_out = cos(pi * factor) + 1
     return end + 0.5 * weight * (start - end) * cos_out
+
+
+@HOOKS.register_module()
+class LinearLrUpdaterHook(LrUpdaterHook):
```
hellock, line 419: It is a special case of `PolyLrUpdaterHook`. We may modify on that.

xiaojieli0903: Thanks!

PR #400, opened 2020-07-08
use env to control compile op or not

xvjiarui:


README.md
```diff
@@ -48,23 +49,37 @@ If you are on macOS, replace the last command with
 CC=lang CXX=clang++ CFLAGS='-stdlib=libc++' pip install -e .
 ```
 
+b. Install full version with CUDA ops.
+
+```bash
+MMCV_WITH_OPS=1 pip install mmcv
```
hellock, line 55: `pip install mmcv-full`

.github/workflows/build.yml
```diff
@@ -28,6 +31,38 @@ jobs:
...
+        if: ${{matrix.torchvision == '0.4.2'}}
+      - name: Install PyTorch
+        run: pip install torch==${{matrix.torch}}+cpu torchvision==${{matrix.torchvision}}+cpu -f https://download.pytorch.org/whl/torch_stable.html
+      - name: Install unittest dependencies
+        run: pip install pytest coverage lmdb PyTurboJPEG
+      - name: Build and install
+        run: rm -rf .eggs && pip install -e .
+      - name: Run unittests
+        run: |
+          pytest tests/ --ignore=tests/test_ops
```
hellock, line 64: No need to use `|` since there is only a single line.

PR #404, opened 2020-07-09
Compitable old interface

yhcao6:


mmcv/utils/misc.py
```diff
@@ -213,3 +214,52 @@ def requires_executable(prerequisites):
...
+            args_info = getfullargspec(old_func)
+            # get name of the function
+            func_name = old_func.__name__
+            if cls_name is not None:
+                func_name = f'{cls_name}.{func_name}'
+            if args:
+                arg_names = args_info.args[:len(args)]
+                for i in range(len(src_arg_name)):
+                    if src_arg_name[i] in arg_names:
+                        print(f'"{src_arg_name[i]}" is deprecate in '
```
hellock, line 246: Use warning instead of print.

mmcv/utils/misc.py
```diff
@@ -213,3 +214,52 @@ def requires_executable(prerequisites):
         1
     """
     return check_prerequisites(prerequisites, checker=_check_executable)
+
+
+def api_warning(src_arg_name, dst_arg_name, cls_name=None):
```
hellock, line 219: `deprecate_api`

hellock: We may use a dict instead of two tuples.

mmcv/utils/misc.py
```diff
@@ -213,3 +214,51 @@ def requires_executable(prerequisites):
         1
     """
     return check_prerequisites(prerequisites, checker=_check_executable)
+
+
+def deprecate_api_warning(name_dict, cls_name=None):
```
hellock, line 219: Either `deprecate_api` or `deprecated_api_warning`

Issue #423, opened 2020-07-15
Typo in lr update hook


sandylaker:
Is the class name [CosineAnealingLrUpdaterHook](https://github.com/open-mmlab/mmcv/blob/7cfc839ea5e6f395682f0af6c63f24500437cef5/mmcv/runner/hooks/lr_updater.py#L227) in `lr_updater`  misspelled?
It should be "Annealing", right?

hellock:
Yes! It would be appreciated if you would like to open a PR to fix it.

PR #431, opened 2020-07-17
Fix issue #423

sandylaker:
Fix typo of cosine annealing


hellock:
Thanks for the fix! You also need to update the unit tests and sign the CLA.

PR #437, opened 2020-07-19
Support ProgressBar update specific steps

wxzs5:
This PR could bring a flexable way to update specific steps for progressbar, lt's useful in some cases, for example:
```python
prog_bar.update(batch_size)
```
instead of:
```python
for _ in range(batch_size):
    prog_bar.update()
```


hellock:
Thanks for the new feature! Could you sign the CLA?

wxzs5:
OK, I signed.

PR #443, opened 2020-07-24
Add syncbuffer hook

Johnson-Wang:
Add sync buffer hook at the end of each epoch


mmcv/runner/base_runner.py
```diff
@@ -271,6 +271,22 @@ def register_hook(self, hook, priority='NORMAL'):
         if not inserted:
             self._hooks.insert(0, hook)
 
+    def register_hook_from_cfg(self, hook_cfg):
+        """Register a hook from its cfg.
+
+        Args:
+            hook_cfg (:obj:`mmcv.Config`): Hook config. It has keys 'type'
```
hellock, line 278: It should be a dict instead of :obj:\`mmcv.Config\`.

PR #457, opened 2020-08-02
Replace `re` with `regex`

dreamerlin:
This PR takes `regex` to replace `re` to solve the [issue](https://github.com/open-mmlab/mmaction2/issues/49): "re.error: bad escape \U at position 2", which usually happens in Windows platform.

### DIFF
--- a/mmcv/cnn/bricks/plugin.py
+++ b/mmcv/cnn/bricks/plugin.py
@@ -1,5 +1,6 @@
 import inspect
-import re
+
+import regex as re
 
 from .registry import PLUGIN_LAYERS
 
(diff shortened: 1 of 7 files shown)


mmcv/cnn/bricks/plugin.py
```diff
@@ -1,5 +1,6 @@
 import inspect
-import re
+
+import regex as re
```
hellock, line 3: Use regex only on windows.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 addict
 numpy
 pyyaml
+regex
```
hellock, line 4: `regex` is not compulsory.

hellock: `regex;sys_platform=='win32'`

PR #461, opened 2020-08-03
Add git hash

ZwwWayne:


mmcv/utils/version_utils.py
```diff
@@ -0,0 +1,29 @@
+import os
+import subprocess
+
+
+def get_git_hash():
```
hellock, line 5: `def get_git_hash(fallback='unknown'):`

mmcv/utils/version_utils.py
```diff
@@ -0,0 +1,29 @@
+import os
+import subprocess
+
+
+def get_git_hash(fallback='unknown'):
```
hellock, line 5: This method is not imported to `mmcv.utils`.

Issue #472, opened 2020-08-07
【BUG】DistSamplerSeedHook() doesn't support the customized batch_sampler


scnuhealthy:
I customize a BatchSampler which based on a DistributedSampler and put it into the dataloader. But when I run my program, an error occurs: AttributeError: 'SequentialSampler' object has no attribute 'set_epoch'. 
After reading the source code of pytorch and mmcv, I find the bug results from the code in DistSamplerSeedHook(). This hook is for shuffle the dataset at every epoch. 
The core code should be _runner.data_loader.batch_sampler.sampler.set_epoch(runner.epoch)_ rather than _runner.data_loader.sampler.set_epoch(runner.epoch)_ (if you customize the batchsampler, batch_sampler.sampler is not euqal to sampler according to the source code of pytorch).

hellock:
Thanks for the report! Would you mind create a PR to fix it?

scnuhealthy:
Sure, I will test and then pull a request.

PR #478, opened 2020-08-10
Fix the bug (#472)

scnuhealthy:


hellock:
Could you sign the CLA?

mmcv/runner/hooks/sampler_seed.py
```diff
@@ -6,4 +6,4 @@
 class DistSamplerSeedHook(Hook):
 
     def before_epoch(self, runner):
-        runner.data_loader.sampler.set_epoch(runner.epoch)
+        runner.data_loader.batch_sampler.sampler.set_epoch(runner.epoch)
```
ZwwWayne, line 9: This line of code will change the original behavior, i.e., the sampler will not be set for each epoch as expected.
The reason for bug #472 is that when the samplers implemented by Pytorch (e.g., `SequentialSampler`) does not have the function `set_epoch`.

scnuhealthy: I got it. Thanks.

PR #481, opened 2020-08-11
fix some pavi logger hooks

Johnson-Wang:
- The actual checkpoint is uploaded instead of the soft link.
- Log the evaluation results based on epoch instead of iter.


mmcv/runner/hooks/logger/pavi.py
```diff
@@ -90,13 +91,19 @@ def log(self, runner):
             tags['momentum'] = momentums[0]
 
         if tags:
-            self.writer.add_scalars(runner.mode, tags, runner.iter)
+            # TODO: Design the logger for evaluation hook as well.
+            if runner.mode == 'val':
+                self.writer.add_scalars(runner.mode, tags, runner.epoch)
```
hellock, line 96: Some repos adopt `IterBasedRunner`.

mmcv/runner/hooks/logger/pavi.py
```diff
@@ -61,7 +65,13 @@ def before_run(self, runner):
             self.init_kwargs = dict()
         self.init_kwargs['task'] = self.run_name
         self.init_kwargs['model'] = runner._model_name
-
+        if runner.meta is not None and 'config_dict' in runner.meta:
+            meta = runner.meta['config_dict'].copy()
+            meta.setdefault('max_iter', runner.max_iters)
```
hellock, line 70: What's the purpose of this line?

mmcv/runner/hooks/logger/pavi.py
```diff
@@ -61,7 +65,13 @@ def before_run(self, runner):
...
+            # non-serializable values are first converted in mmcv.dump to json
+            meta_json = json.loads(mmcv.dump(meta, file_format='json'))
```
hellock, line 72: The variable name `meta_json` is improper. It is still a dict instead of json string.

mmcv/runner/hooks/logger/pavi.py
Johnson-Wang: `max_.*iter` will be parsed as the maximum iteration to show the training progress bar in pavi.

hellock: We can leave a comment here.

PR #492, opened 2020-08-13
Add `tin_shift` function

dreamerlin:
This PR adds `tin_shift` function which is used in [TIN](http://arxiv.org/abs/2001.06499) model. The codes are referred from https://github.com/deepcs233/TIN/blob/master/cuda_shift/rtc_wrap.py

### DIFF
--- a/mmcv/ops/tin_shift.py
+++ b/mmcv/ops/tin_shift.py
@@ -4,6 +4,7 @@
 # shaoh19@mails.tsinghua.edu.cn, sjqian@cse.cuhk.edu.hk, yuliu@ee.cuhk.edu.hk
 
 import torch
+import torch.nn as nn
 from torch.autograd import Function
 
 from ..utils import ext_loader
@@ -34,3 +35,12 @@ class TINShiftFunction(Function):
         ext_module.tin_shift_backward(grad_output, shift, data_grad_input)
 
         return data_grad_input, shift_grad_input
+
+
+tin_shift = TINShiftFunction.apply
+
+
+class TINShift(nn.Module):
+
+    def forward(self, input, shift):
+        return tin_shift(input, shift)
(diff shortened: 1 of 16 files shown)


mmcv/ops/tin_shift.py
```diff
@@ -0,0 +1,46 @@
...
+        shift_grad_input = shift.new(*shift.size()).zero_()
+        ext_module.tin_shift_backward(grad_output, shift, data_grad_input)
+
+        return data_grad_input, shift_grad_input
+
+
+tin_shift = TINShiftFunction.apply
+
+
+class TINShift(nn.Module):
```
hellock, line 43: Docstring.

dreamerlin:
off-pr: 
It seems the [pre-commit](https://github.com/open-mmlab/mmcv/blob/master/.pre-commit-config.yaml#L37-L44) use the clang of the local version to format the `cu`, `c++`, etc code. However, the [CI](https://github.com/open-mmlab/mmcv/blob/master/.github/workflows/build.yml#L27-L32) will use the [git action](https://github.com/DoozyX/clang-format-lint-action) (which uses clang-format-9) to format code. 
If pr submitter's clang is of < version 9, it may have conflict with the CI during linting. Bud it is a troublemaker to install Clang-9 especially for Unbuntu 16.04, I think it is better to make the pair (CI and pre-commit) corresponding with each other.

hellock:
> off-pr:
> It seems the [pre-commit](https://github.com/open-mmlab/mmcv/blob/master/.pre-commit-config.yaml#L37-L44) use the clang of the local version to format the `cu`, `c++`, etc code. However, the [CI](https://github.com/open-mmlab/mmcv/blob/master/.github/workflows/build.yml#L27-L32) will use the [git action](https://github.com/DoozyX/clang-format-lint-action) (which uses clang-format-9) to format code.
> If pr submitter's clang is of < version 9, it may have conflict with the CI during linting. Bud it is a troublemaker to install Clang-9 especially for Unbuntu 16.04, I think it is better to make the pair (CI and pre-commit) corresponding with each other.

If you are using vscode on ubuntu, there is no need to manually install clang-format. It is included in the c++ extension of vscode.

PR #494, opened 2020-08-14
Support for inheriting lists contents in addition to dictionaries in Config

anthracene:
Currently, child configurations are able to inherit and modify the contents of dictionaries in parent configurations. However, any attempt to modify a part of the configuration contained in a list replaces the contents of the parent list with the child list rather than merging them. The consequence of this is that child configs that need to change things contained in a list have to duplicate all of the unchanged items in the child config to avoid deleting them.

This pull requests adds the ability to inherit from lists and over-ride only the portions needed. Although the implementation of inheritance for dictionaries is changed, the behavior is unchanged.  For instance, this makes it possible to change the number of classes for DetecoRS in a child config like this:
```
model=dict(roi_head=dict(bbox_head=[dict(num_classes=2),dict(num_classes=2),dict(num_classes=2)]))
```
rather than having to do something like this
```
model=dict(roi_head=dict(bbox_head=[
            dict(
                type='Shared2FCBBoxHead',
                in_channels=256,
                fc_out_channels=1024,
                roi_feat_size=7,
                num_classes=2,
                bbox_coder=dict(
                    type='DeltaXYWHBBoxCoder',
                    target_means=[0., 0., 0., 0.],
                    target_stds=[0.1, 0.1, 0.2, 0.2]),
                reg_class_agnostic=True,
                loss_cls=dict(
                    type='CrossEntropyLoss',
                    use_sigmoid=False,
                    loss_weight=1.0),
                loss_bbox=dict(type='SmoothL1Loss', beta=1.0,
                               loss_weight=1.0)),
            dict(
                type='Shared2FCBBoxHead',
                in_channels=256,
                fc_out_channels=1024,
                roi_feat_size=7,
                num_classes=2,
           ...
```
As implemented here, the _delete_ key concept is extended such that if the first element of the list in the child config is _delete_, then list in the child config replaces the list in the parent config. Otherwise, elements in the child config list are merged with elements in the parent config list. When an element in the child config list is None, the corresponding element in the parent config list is left unchanged.
It would be possible to do type checking to make sure that lists appear in the child config only in places where there is a list in the parent config, but this was found to break some configs in mmdetection, which depend on child config lists overwriting parent config elements where the type is different (e.g. tuple), so that behavior has been maintained.


xvjiarui:
Hi @anthracene 
Thanks for your contribution! 
Please follow [contributing guide](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to pass CI. 
Btw, please rebase latest master and add unittests for this feature.

anthracene:
@xvjiarui I added unit tests and cleaned up the formatting. I think I appropriately squashed and rebased this, but I'm a little new to pull requests in git; please let me know if I did something wrong or need to do something else. Thanks!

xvjiarui:
Hi @anthracene 
The [diff review](https://github.com/open-mmlab/mmcv/pull/494/files) is not correct.
You may try to merge the current master to resolve it.

anthracene:
@xvjiarui: I merged with the master which seems to have cleaned up the diffs. Let me know if this gives you what you need.

anthracene:
Please don't merge yet -- I found a bug I need to fix.

xvjiarui:
Hi @anthracene 
Any update?

anthracene:
I believe I've got the bug fixed, but want to do a little more testing
before I push. Hopefully later today or tomorrow.

anthracene:
OK, @xvjiarui I think this is ready to go (unless I screwed up with git again). I really wanted the merge behavior of lists to parallel that of dicts -- merge by default, overwrite only if the delete key is included, as described in my first comment above. Unfortunately there are too many configs out in the wild that depend on the current behavior of child lists overwriting parent lists that would behave differently if lists merge by default. To avoid extensive breakage, I've had to keep the overwrite by default behavior and merge lists only when explicitly specified.

I've done this by creating a merge key, analogous to the delete key for dicts. If the first element of a list in a child config is `'_merge_'`, that element is deleted and the remainder of the elements are merged with the corresponding parent list; in all other cases the child config list replaces the parent config list. So if you want to change the number of classes in DetectoRS, you would do it like this
```
model=dict(roi_head=dict(bbox_head=['_merge_',dict(num_classes=2),dict(num_classes=2),dict(num_classes=2)]))
```
I'm not really happy with the difference between default behavior for dict and list, but unfortunately I think this is the only way to implement the feature without massive breakage in projects that depend on mmcv.
To make sure I'm not going to break things with this pull request, I ran print_config.py on all the configs in mmdetection with  and without my changes and diffed the output. The only differences were the ordering of a few keys in a few dicts, which shouldn't be significant.

anthracene:
@xvjiarui would love to see this get merged as I'm using it pretty extensively in my configs, have found it really useful and think others probably would as well. Anything I can do to facilitate review?

xvjiarui:
Hi @anthracene 
Sorry for the late reply. 
Currently, we don't have plan to merge this PR. We may make our config system more flexible and simpler in the future.

anthracene:
Hi @xvjiarui --
Really disappointing, but thanks for letting me know.
I really do hope that you'll look at overhauling the config system soon. The inheritance-based config system that you all have built into mmcv is extremely powerful and quite elegant, but it seems to have been designed with the assumption that essentially everything would be in hierarchies of dicts, and that any limited use of lists in config files would be nothing more than lists of numbers or strings. Looking at the config files you have for your dependent projects (e.g. mmdetection) that hasn't at all turned out to be the case, which means that the inheritance feature of your config system is effectively broken for anything that has dicts inside a list: in order to change anything there, you have to copy the whole block of config, largely defeating the purpose of inheritance.
Additionally, for the same reasons, the really important feature of being able to modify individual config file settings from the command line is completely broken for any setting nested in a list. This means, for instance, no modification of image augmentation parameters from the command line, since those are all in the ```pipeline``` list -- that really hurts for doing hyperparameter optimization. I have code that fixes this command line limitation as well, but it's based on the list merging code I already submitted, and since it looks like you're not going to go this way I won't bother packaging it up as a PR.
Please let me know if there's anything I can do to support revisiting the way that the config system deals with lists. I'm not wedded to the solution I came up with, but the current state is pretty glaringly broken, and really limits the ability to realize what I think is a quite beautiful software architectural vision.

Issue #253, opened 2020-04-28
May be Epoch(val)?


hust-nj:
Hi, when I use mmcv to train imagenet classification, it will print something like this(taken cifar10 example)
```text
2020-04-28 17:05:41,729 - INFO - Epoch [1][700/782]     lr: 0.10000, eta: 0:03:45, time: 0.049, data_time: 0.002, memory: 402, loss: 1.6594, acc_top1: 38.3125, acc_top5: 88.9062
2020-04-28 17:05:44,121 - INFO - Epoch [1][750/782]     lr: 0.10000, eta: 0:03:40, time: 0.048, data_time: 0.002, memory: 402, loss: 1.6392, acc_top1: 39.8750, acc_top5: 88.9062
2020-04-28 17:05:50,121 - INFO - Epoch(train) [1][157]  loss: 1.6123, acc_top1: 40.4100, acc_top5: 89.8100
2020-04-28 17:05:53,214 - INFO - Epoch [2][50/782]      lr: 0.10000, eta: 0:03:28, time: 0.062, data_time: 0.010, memory: 402, loss: 1.6114, acc_top1: 39.0312, acc_top5: 90.2812
2020-04-28 17:05:55,918 - INFO - Epoch [2][100/782]     lr: 0.10000, eta: 0:03:25, time: 0.054, data_time: 0.002, memory: 402, loss: 1.5901, acc_top1: 41.875 
```
I feel strange that in validation step it prints `Epoch(train)`, I thought it should print `Epoch(val)`.
Then I looked into code in logger hooks.
https://github.com/open-mmlab/mmcv/blob/f7de63fdd0b48a56eadb717cd9260260705f0632/mmcv/runner/hooks/logger/text.py#L92-L101

I insert the code `print(runner.mode)` and `print(mode)` in it, find that `print(mode)` always show train, even when `print(runner.mode)` show val.
Is there something wrong or a bug existing?

ecm200:
I have come across the exact same issue when using MMDetection. 

I used the training example script, setting the validation argument to false, and changing the cfg.workflow = [('train',1), ('val', 1)]

My understanding from MMCV runner.run() method is that the workflow is formatted to run these processes in order, for the number of epochs specified as int. So my interpretation of how I have set this up, is to that I will perform validation at the end of each training epoch.

This piece of code ensures that the validation dataset is loaded when a validation step is requested.
```python
datasets = [build_dataset(cfg.data.train)]
[...]

PR #495, opened 2020-08-14
Logging mode by using runner.mode

dreamerlin:
Log display fix for https://github.com/open-mmlab/mmcv/issues/253

### DIFF
--- a/mmcv/runner/hooks/logger/text.py
+++ b/mmcv/runner/hooks/logger/text.py
@@ -142,8 +142,7 @@ class TextLoggerHook(LoggerHook):
 
     def log(self, runner):
         log_dict = OrderedDict()
-        mode = runner.mode
-        log_dict['mode'] = mode
+        log_dict['mode'] = runner.mode
         log_dict['epoch'] = runner.epoch + 1
         if self.by_epoch:
             log_dict['iter'] = runner.inner_iter + 1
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/logger/text.py
```diff
@@ -142,8 +142,7 @@ def _round_float(self, items):
 
     def log(self, runner):
         log_dict = OrderedDict()
-        # training mode if the output contains the key "time"
-        mode = 'train' if 'time' in runner.log_buffer.output else 'val'
+        mode = runner.mode
```
hellock, line 145: The logic is not expected.

dreamerlin:
CI Failure is caused by `test_rgb2ycbcr`, which is not related with this pr.

mmcv/runner/hooks/logger/text.py
```diff
@@ -142,10 +142,22 @@ def _round_float(self, items):
...
+            # normal val mode
+            # runner.epoch += 1 has been done before val workflow
+            log_dict['mode'] = 'val'
+            log_dict['epoch'] = runner.epoch
+        else:
+            # Evaluation hook during train mode
+            # runner.epoch += 1 should be done since training has ended
+            log_dict['mode'] = 'val'
+            log_dict['epoch'] = runner.epoch + 1
+
```
yhcao6, line 160: It may be better to simplify the logic as follows:
```
        if runner.mode == 'train':
            log_dict['mode'] = 'train' if 'time' in runner.log_buffer.output \
                else 'val'
            log_dict['epoch'] = runner.epoch + 1
        elif runner.mode == 'val':
            # normal val mode
            # runner.epoch += 1 has been done before val workflow
            log_dict['mode'] = 'val'
            log_dict['epoch'] = runner.epoch
        else:
            raise ValueError

```

PR #497, opened 2020-08-14
Fix mmcv-dataparallel

ZwwWayne:
Fix https://github.com/open-mmlab/mmdetection/issues/1501
Fix https://github.com/open-mmlab/mmdetection/issues/3048

A RetinaNet is tested with three modes: multi-GPU distributed testing, single-GPU testing with MMDataParallel, and CPU testing with MMDataParallel. The model achieves similar AP under these three modes.

### DIFF
--- a/mmcv/parallel/_functions.py
+++ b/mmcv/parallel/_functions.py
@@ -19,8 +19,11 @@ def scatter(input, devices, streams=None):
         output = input.contiguous()
         # TODO: copy to a pinned buffer first (if copying from CPU)
         stream = streams[0] if output.numel() > 0 else None
-        with torch.cuda.device(devices[0]), torch.cuda.stream(stream):
-            output = output.cuda(devices[0], non_blocking=True)
+        if devices != [-1]:
+            with torch.cuda.device(devices[0]), torch.cuda.stream(stream):
+                output = output.cuda(devices[0], non_blocking=True)
+        # This is a workaround
+        output = output.unsqueeze(0)
         return output
     else:
         raise Exception(f'Unknown type {type(input)}.')
@@ -62,7 +65,7 @@ class Scatter:
     def forward(target_gpus, input):
         input_device = get_input_device(input)
         streams = None
-        if input_device == -1:
+        if input_device == -1 and target_gpus != [-1]:
             # Perform CPU to GPU copies in a background stream
             streams = [_get_stream(device) for device in target_gpus]
 
--- a/mmcv/parallel/_functions.py
+++ b/mmcv/parallel/_functions.py
@@ -22,8 +22,10 @@ def scatter(input, devices, streams=None):
         if devices != [-1]:
             with torch.cuda.device(devices[0]), torch.cuda.stream(stream):
                 output = output.cuda(devices[0], non_blocking=True)
-        # This is a workaround
-        output = output.unsqueeze(0)
+        else:
+            # unsquzee the first dimension thus the tensor's shape is the
+            # same as those scattered with GPU.
+            output = output.unsqueeze(0)
         return output
     else:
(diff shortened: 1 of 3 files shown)


mmcv/parallel/data_parallel.py
```diff
@@ -1,18 +1,90 @@
...
+
+    MMDataParallel has two main differences with PyTorch DataParallel:
+
+    - It supports a custom type :class:`DataContainer` which allows more
+      flexible control of input data during both GPU and CPU inference.
+    - It implement two more APIs ``train_step()`` and ``val_step()``.
+
+    Args:
+        module (:class:`nn.Module`): Module to be encapsulated.
+        device_ids (list[int]): Device IDS of modules to be scattered to.
```
xvjiarui, line 23: We may add docs about CPU cases as well.

xvjiarui: We may also check if the user input negative number, only `[-1]` is allowed?

mmcv/parallel/_functions.py
```diff
@@ -19,8 +19,11 @@ def scatter(input, devices, streams=None):
         output = input.contiguous()
         # TODO: copy to a pinned buffer first (if copying from CPU)
         stream = streams[0] if output.numel() > 0 else None
-        with torch.cuda.device(devices[0]), torch.cuda.stream(stream):
-            output = output.cuda(devices[0], non_blocking=True)
+        if devices != [-1]:
+            with torch.cuda.device(devices[0]), torch.cuda.stream(stream):
+                output = output.cuda(devices[0], non_blocking=True)
+        # This is a workaround
+        output = output.unsqueeze(0)
```
xvjiarui, line 26: This will affect original behavior.

mmcv/parallel/data_parallel.py
```diff
@@ -1,18 +1,93 @@
...
+                    'module must have its parameters and buffers '
+                    f'on device {self.src_device_obj} (device_ids[0]) but '
+                    f'found one of them on device: {t.device}')
+
+        inputs, kwargs = self.scatter(inputs, kwargs, self.device_ids)
+        if len(self.device_ids) == 1:
+            return self.module(*inputs[0], **kwargs[0])
+        replicas = self.replicate(self.module, self.device_ids[:len(inputs)])
+        outputs = self.parallel_apply(replicas, inputs, kwargs)
+        return self.gather(outputs, self.output_device)
```
hellock, line 81:
```python
if not self.device_ids:
    # our own codes
else:
    super().forward(xxx)
```

mmcv/parallel/data_parallel.py
```diff
@@ -1,18 +1,93 @@
...
+    """
+
+    def __init__(self, module, device_ids=None, output_device=None, dim=0):
+        super(DataParallel, self).__init__()
+
+        if not torch.cuda.is_available():
+            self.module = module
+            self.device_ids = []
+            # we add this line from the original DP to enable CPU inference.
+            self.dim = dim
```
hellock, line 36: Only this line is needed.

mmcv/parallel/scatter_gather.py
```diff
@@ -15,7 +15,11 @@ def scatter(inputs, target_gpus, dim=0):
 
     def scatter_map(obj):
         if isinstance(obj, torch.Tensor):
-            return OrigScatter.apply(target_gpus, None, dim, obj)
+            if target_gpus != [-1]:
+                return OrigScatter.apply(target_gpus, None, dim, obj)
+            else:
+                # for CPU inference we use self-implemented scatter
+                return Scatter.forward(target_gpus, obj.data)
```
hellock, line 22: Why `obj.data`?

Issue #499, opened 2020-08-15
Error: pickle.loads() with mmcv.Config object


loveunk:
I am trying to `pickle.dumps()` a `mmcv.Config` instance to a file, then `pickle.loads()` it back.

```python
import pickle, mmcv
config_file = '../configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
config = mmcv.Config.fromfile(config_file)

pickled_lambda = pickle.dumps(config)
with open("config_file.pkl", "wb") as f:
    f.write(pickled_lambda)

with open("config_file.pkl", "rb") as f:
    model_s = f.read()
config = pickle.loads(model_s)
```

ERROR happened in `pickle.loads`:
```shell
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
<ipython-input-4-df23a8c2d7c4> in <module>
      1 with open("config_file.pkl", "rb") as f:
      2     model_s = f.read()
----> 3 config = pickle.loads(model_s)

~/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/utils/config.py in __getattr__(self, name)
    306 
    307     def __getattr__(self, name):
--> 308         return getattr(self._cfg_dict, name)
    309 
    310     def __getitem__(self, name):

... last 1 frames repeated, from the frame below ...

~/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/utils/config.py in __getattr__(self, name)
    306 
    307     def __getattr__(self, name):
--> 308         return getattr(self._cfg_dict, name)
    309 
    310     def __getitem__(self, name):

RecursionError: maximum recursion depth exceeded while calling a Python object
```

loveunk:
Close this issue since PR #500 has been merged.

PR #500, opened 2020-08-15
Support pickle.loads (#499)

loveunk:


xvjiarui:
Hi @loveunk 
Thanks for your contribution. 
Please add more comments to elaborate on this PR. And we may also add unittest for pickle.loads.

loveunk:
Hi @xvjiarui , thank you for your comments.

This PR tries to solve one `pickle.loads()` issue as described in #499 .

It looks like the `mmcv.Config` object cannot be restored correctly by `pickle` in current mmcv implementation.

https://github.com/open-mmlab/mmcv/blob/7b18b97791167556643ed04ea8354f19c6306044/mmcv/utils/config.py#L356-L357

In `pickle.loads`, it tries to query `__setstate__` of the `mmcv.Config` instance, and not found. Then the query is captured by `__getattr__`.

However, in `getattr(self._cfg_dict, name)`, `self._cfg_dict` does not exist because the `__init__()` is not called when unpickling an instance.
Then `_cfg_dict` is queried with `__getattr__`. A recursion occurs.

xvjiarui:
Hi @loveunk 
Sorry for the late reply. 
You may add unittest for this.

loveunk:
> Hi @loveunk
> Sorry for the late reply.
> You may add unittest for this.

@xvjiarui Sure, I will commit some unit tests later. BTW, I am not familiar with the unit tests for mmcv. Is there any help documents to follow?

xvjiarui:
Hi @loveunk 
You may check [this](https://github.com/open-mmlab/mmcv/blob/master/tests/test_config.py) file to have an understanding of unittest.

loveunk:
@xvjiarui Thanks for your help. I have committed a new unittest.

xvjiarui:
Thx! LGTM

xvjiarui (approved):

PR #511, opened 2020-08-20
Support load with mmcls://

xvjiarui:


mmcv/runner/checkpoint.py
```diff
@@ -21,6 +23,7 @@
 ENV_MMCV_HOME = 'MMCV_HOME'
 ENV_XDG_CACHE_HOME = 'XDG_CACHE_HOME'
 DEFAULT_CACHE_DIR = '~/.cache'
+MMCLS_MODELZOO = 'https://raw.githubusercontent.com/open-mmlab/mmclassification/master/docs/model_zoo.md'  # noqa
```
hellock, line 26: This URL may not be accessible in some regions.

hellock (approved):

PR #514, opened 2020-08-21
Import modules from a string list

Johnson-Wang:


tests/test_misc.py
```diff
@@ -100,3 +100,25 @@ def func_c():
         ' please install them first.\n')
 
     assert func_c() == 1
+
+
+def test_custom_imports():
+    # multiple imports
+    runner, image = mmcv.import_modules_from_strings(
+        ['mmcv.runner', 'mmcv.image'])
```
hellock, line 108: `mmcv.runner` does not exist when pytorch is not installed. We can use other test cases like `os.path`.

hellock (approved):

PR #522, opened 2020-08-26
add Swish activation

xiaojieli0903:
add Swish activation, which can be used in EfficientNet architectures.


mmcv/cnn/bricks/swish.py
```diff
@@ -0,0 +1,24 @@
+import torch
+import torch.nn as nn
+
+from .registry import ACTIVATION_LAYERS
+
+
+@ACTIVATION_LAYERS.register_module()
+class Swish(nn.Module):
+    """ wish Module.
```
hellock, line 9: typo: Swish

xiaojieli0903: Thanks. It's fixed in the latest version.

hellock (approved):

PR #523, opened 2020-08-30
[feature] Add data pipeline

hellock:


mmcv/data_pipeline/data_pipeline.py
```diff
@@ -0,0 +1,69 @@
...
+
+class DataPipeline:
+
+    def __init__(self, transforms):
+        if not isinstance(transforms, Sequence):
+            raise TypeError('transforms must be a sequence')
+        self.transforms = []
+        for transform in transforms:
+            if isinstance(transform, dict):
+                transform = build_from_cfg(transform, TRANSFORMS)
```
ZwwWayne, line 15: Maybe we could still create a method `build_transform` and use that to create transforms.

PR #529, opened 2020-09-02
Implementation of 2D convolution in tensorflow with `padding` as "same"

xiaojieli0903:
Implementation of 2D convolution in tensorflow with `padding` as "same" (https://www.tensorflow.org/api_docs/python/tf/nn/conv2d).


xiaojieli0903:
This will be used for the EfficientNet pretrained models, which are converted from https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet. @hellock

mmcv/cnn/bricks/conv2d_samepadding.py
```diff
@@ -0,0 +1,61 @@
+import math
+
+from torch import nn
+from torch.nn import functional as F
+
+from .registry import CONV_LAYERS
+
+
+@CONV_LAYERS.register_module()
+class Conv2dSamePadding(nn.Conv2d):
```
hellock, line 10: Users may be confused with the class name. It sounds like Conv2d with "replicate" padding mode.

xiaojieli0903: How about using `Conv2dAdaptivePadding`? Because it will pad the image width (height) with zeros up to a multiple of the filter width (height) when the image width is not a multiple of the filter width or the image height is not a multiple of the filter height in an adaptive way.

hellock: Sounds good.

xiaojieli0903: The class mane has been changed to `Conv2dAdaptivePadding` in the latest version. Please have a check.

hellock (approved):

PR #531, opened 2020-09-02
Support reading video from url

dreamerlin:


mmcv/video/io.py
```diff
@@ -62,7 +62,13 @@ class VideoReader:
     """
 
     def __init__(self, filename, cache_capacity=10):
-        check_file_exist(filename, 'Video file not found: ' + filename)
+        try:
+            check_file_exist(filename, 'Video file not found: ' + filename)
+        except FileNotFoundError:
+            check_url_valid(filename)
```
hellock, line 68: Why not check the string starting with http?

hellock (approved):

PR #544, opened 2020-09-08
Reordered the hooks and use attributes rather than args.

SuX97:
Use attributes in `epoch_based_runner`, therefore hooks like MultiGridHook can take effect.

### DIFF
--- a/mmcv/runner/epoch_based_runner.py
+++ b/mmcv/runner/epoch_based_runner.py
@@ -21,8 +21,8 @@ class EpochBasedRunner(BaseRunner):
         self.model.train()
         self.mode = 'train'
         self.data_loader = data_loader
-        self._max_iters = self._max_epochs * len(self.data_loader)
         self.call_hook('before_train_epoch')
+        self._max_iters = self._max_epochs * len(self.data_loader)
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

mmcv/runner/epoch_based_runner.py
```diff
@@ -90,19 +91,25 @@ def run(self, data_loaders, workflow, max_epochs, **kwargs):
         assert len(data_loaders) == len(workflow)
 
         self._max_epochs = max_epochs
+
+        for i, flow in enumerate(workflow):
+            mode, epochs = flow
+            if mode == 'train':
+                self._max_iters = self._max_epochs * len(data_loaders[i])
+                break
```
ZwwWayne, line 98: Can we explain the necessity here? For example, which hook's `before_run` function requires the self._max_iters to be calculated.

SuX97: Never mind, I have rewind the change of order back. In some hooks such as `MomentumUpdaterHook`, self._max_iters is necessary, and the previous order changing will cause them to fail. I will not change the order now, and update the self._max_iters in my own hook.

Issue #541, opened 2020-09-06
Why is there seperate train() and val() in Runner


zyc-ai:
Hi, 

I'm wondering why is there two seperate and very similar function in runner.Runner, instead of a more modularised one like this:

```python
import time
import torch
import mmcv

from mmcv.runner import EpochBasedRunner as _EpochBasedRunner
from mmcv.runner.utils import get_host_info


class EpochBasedRunner(_EpochBasedRunner):
    def _model(self, data_batch, **kwargs):
        if self.batch_processor is None:
            outputs = self.model.train_step(data_batch, self.optimizer, **kwargs)
        else:
            outputs = self.batch_processor(self.model, data_batch, train_mode=self.train_mode, **kwargs)
        if not isinstance(outputs, dict):
            raise TypeError('"batch_processor()" or "model.train_step()"'
                    ' must return a dict')
        return outputs

    def _run(self, data_loader, mode, **kwargs):
        self.mode = mode
        self.model.train() if self.train_mode else self.model.eval()
        self.data_loader = data_loader
        self._max_iters = self._max_epochs * len(self.data_loader)
        self.call_hook(f'before_{mode}_epoch')
        time.sleep(2)  # Prevent possible deadlock during epoch transition
        for i, data_batch in enumerate(self.data_loader):
            self._inner_iter = i
            self.call_hook(f'before_{mode}_iter')
            outputs = None
            if self.train_mode:
                outputs = self._model(data_batch, **kwargs)
            else:
                with torch.no_grad():
                    outputs = self._model(data_batch, **kwargs)
            if outputs is None:
                raise ValueError(f'Unable to get any output, it is possible the mode {self.mode} is illegal')
            if 'log_vars' in outputs:
                self.log_buffer.update(outputs['log_vars'],
                                       outputs['num_samples'])
            self.outputs = outputs
            self.call_hook(f'after_{mode}_iter')
            self._iter += 1

        self.call_hook(f'after_{mode}_epoch')
        if self.train_mode:
            self._epoch += 1

    def run(self, data_loaders, workflow, max_epochs, **kwargs):
        """Start running.

        Args:
            data_loaders (list[:obj:`DataLoader`]): Dataloaders for training
                and validation.
            workflow (list[tuple]): A list of (phase, epochs) to specify the
                running order and epochs. E.g, [('train', 2), ('val', 1)] means
                running 2 epochs for training and 1 epoch for validation,
                iteratively.
            max_epochs (int): Total training epochs.
[...]

PR #553, opened 2020-09-14
merge the calling of train/val_step and batch_processor into run_iter

zyc-ai:
ref #541

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -173,10 +173,6 @@ class BaseRunner(metaclass=ABCMeta):
         return self._max_iters
 
     @abstractmethod
-    def run_iter(self, data_batch, **kwargs):
-        pass
-
-    @abstractmethod
     def train(self):
         pass
 
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -98,7 +98,6 @@ class BaseRunner(metaclass=ABCMeta):
         self.optimizer = optimizer
         self.logger = logger
         self.meta = meta
-        self.train_mode = False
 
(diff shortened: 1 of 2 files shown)


mmcv/runner/base_runner.py
```diff
@@ -172,6 +173,10 @@ def max_iters(self):
         """int: Maximum training iterations."""
         return self._max_iters
 
+    @abstractmethod
```
hellock, line 176: This method may be just implemented as private method of each subclass.

mmcv/runner/epoch_based_runner.py
```diff
@@ -17,29 +17,34 @@ class EpochBasedRunner(BaseRunner):
     This runner train models epoch by epoch.
     """
 
+    def run_iter(self, data_batch, **kwargs):
+        if self.batch_processor is not None:
+            outputs = self.batch_processor(
+                self.model, data_batch, train_mode=self.train_mode, **kwargs)
+        elif self.train_mode:
```
hellock, line 24: I suggest making it a argument of `run_iter()` instead of an attribute of class, which remains exactly the same with previous behavior.

zyc-ai: https://github.com/open-mmlab/mmcv/pull/553/commits/be7444b900dc8b14db0179137358ebe0ef058bbb

mmcv/runner/base_runner.py
zyc-ai: https://github.com/open-mmlab/mmcv/pull/553/commits/97e47f4c3f986d5d1a655e2fc01f58792b6f5d98

hellock (approved):

PR #554, opened 2020-09-14
remove the set of _max_iter in epoch based runner

zyc-ai:
`EpochBasedRunner._max_iter` should be set in line 24, `train()` of `epoch_based_runner.py`, and the setting between 93-97 is redundant and unnecessary.


zyc-ai:
Looks like 93-97

PR #557, opened 2020-09-16
Use copy instead of symlink on windows

hellock:


mmcv/runner/epoch_based_runner.py
```diff
@@ -163,7 +165,11 @@ def save_checkpoint(self,
         # in some environments, `os.symlink` is not supported, you may need to
         # set `create_symlink` to False
         if create_symlink:
-            mmcv.symlink(filename, osp.join(out_dir, 'latest.pth'))
+            dst_file = osp.join(out_dir, 'latest.pth')
+            if sys.platform != 'win32':
```
xvjiarui, line 169: In MMDetection, we use `platform.system()`. Maybe it is better to keep it consistent?

xvjiarui (approved):

PR #570, opened 2020-09-21
Add runner builder

daavoo:
This P.R. adds a new method `build_runner` and a new Registry object: `RUNNERS` for managing runner types.
This allows to specify runner type (and corresponding parameters) via configuration files.

The `max_iters` and `max_epochs` arguments have been moved from `run` to `__init__` (this change is backwards compatible with deprecation warning).
I don't know if there was a use case that needed to have them in `run` but in all downstream repos (mmdet, mmseg, etc) those args were parsed from config so i think it makes sense to set them during instantiation.

Discussed in https://github.com/open-mmlab/mmsegmentation/pull/118

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -43,6 +43,8 @@ class BaseRunner(metaclass=ABCMeta):
         meta (dict | None): A dict records some import information such as
             environment info and seed, which will be logged in logger hook.
             Defaults to None.
+        max_epochs (int, optional): Total training epochs.
+        max_iters (int, optional): Total training iterations.
     """
 
     def __init__(self,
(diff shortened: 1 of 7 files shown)


ZwwWayne (approved):
The PR looks good to me.

mmcv/runner/base_runner.py
```diff
@@ -121,8 +125,8 @@ def __init__(self,
         self._epoch = 0
         self._iter = 0
         self._inner_iter = 0
-        self._max_epochs = 0
-        self._max_iters = 0
+        self._max_epochs = max_epochs
+        self._max_iters = max_iters
```
hellock, line 129: We may assert that these two arguments cannot be both set.

hellock (approved):

PR #574, opened 2020-09-23
Update lr_updater.py

Kuro96:
since epoch/iteration in  runner starts with 0, we shouldn't leave the latter iteration to former (12th epoch for example, with first period equal to 12) period.

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -318,7 +318,7 @@ def get_position_from_periods(iteration, cumulative_periods):
         int: The position of the right-closest number in the period list.
     """
     for i, period in enumerate(cumulative_periods):
-        if iteration <= period:
+        if iteration < period:
             return i
     raise ValueError(f'Current iteration {iteration} exceeds '
                      f'cumulative_periods {cumulative_periods}')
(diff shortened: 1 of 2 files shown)


ZwwWayne:
This depends on the needs. Usually, the case in the boundary between two periods will not affect the performance much. However, since this is added and validated in [mmediting](https://github.com/open-mmlab/mmediting/blob/7721dfa67ebfe25f6dd209bc442d69fc5d19ec95/configs/restorers/esrgan/esrgan_psnr_x4c64b23g32_g1_1000k_div2k.py), I am afraid this PR will not be merged.

hejm37:
For `IterBasedRunner`, the difference that an iteration makes is minor. But for `EpochBasedRunner`, the case is difference. And I think the update is reasonable. @xinntao

mmcv/runner/hooks/lr_updater.py
```diff
@@ -318,7 +318,7 @@ def get_position_from_periods(iteration, cumulative_periods):
         int: The position of the right-closest number in the period list.
     """
     for i, period in enumerate(cumulative_periods):
-        if iteration <= period:
+        if iteration < period:
```
hejm37, line 321: You may need to modify the docstring.
```
if iteration == 300, return 3.
```

Kuro96: My fault, it should be done now.

nbei (approved):
I have no further comments.

Kuro96:
> You may need further modification to pass the unit test in CI.

I've updated `test_hooks.py` now

nbei (approved):
I have no further comments.

Issue #577, opened 2020-09-23
Bug feedback.


wdmwhh:
Thanks for your codes.
In [mmcv/runner/hooks/optimizer.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/optimizer.py#L63), which migrates from [mmdet/core/fp16/hooks.py](https://github.com/open-mmlab/mmdetection/blob/v2.1.0/mmdet/core/fp16/hooks.py#L39), 
`runner.optimizer.param_groups = copy.deepcopy(
            runner.optimizer.param_groups)` changes the ids of `runner.optimizer.param_groups` related to the the keys of  `runner.optimizer.state`. Hence, if `runner.optimizer.state` is not empty, e.g. resumed from a checkpoint, `runner.optimizer.state[param_after_copy] != runner.optimizer.state[param_orin]` is empty. For instance, if one trains the model resumed from a checkpoint in fp16 setting, the saved momentum_buffer will be ignored always.

wdmwhh:
In the official implementation of fp16 [apex](https://github.com/NVIDIA/apex/blob/master/apex/fp16_utils/fp16_optimizer.py#L52), it resets existing state dict key to the new master param.
`if param in self.optimizer.state:
                           self.optimizer.state[master_param] = self.optimizer.state.pop(param)`

ycxioooong:
Hi @wdmwhh sorry for a late reply. Apart from modifying the current implementation, we also adopt official fp16 using `torch.cuda.amp` in the coming PR (#791). Thanks for your suggestions.

PR #566, opened 2020-09-18
fix save checkpoint error with resume training at pytorch 1.6

halostorm:
Please refer to https://github.com/open-mmlab/mmsegmentation/issues/127 for details.

PR #580, opened 2020-09-25
Fixed fp16_optimizer state bug

wdmwhh:


yhcao6:
Can you give an example where the bug will happen?

wdmwhh:
> Can you give an example where the bug will happen?

When one trains the model resumed from a checkpoint in fp16 setting. 
https://github.com/open-mmlab/mmcv/issues/577

yhcao6:
It seems #566 is related to this Pr, could you have a check that?

wdmwhh:
> It seems #566 is related to this Pr, could you have a check that?

Yes,  they are related. Both the changes are able to fix the bug.

shinya7y:
Which implementation is better in terms of memory consumption?

yhcao6:
Thanks for your contribution. Could you fix the lint error?

yhcao6 (approved):
Will fix lint after merge

PR #582, opened 2020-09-25
register deconv in CONV_LAYERS

GT9505:
Deconv, which is used in flownet, is registered into CONV_LAYERS in this PR.

### DIFF
--- a/mmcv/cnn/bricks/conv.py
+++ b/mmcv/cnn/bricks/conv.py
@@ -6,7 +6,6 @@ CONV_LAYERS.register_module('Conv1d', module=nn.Conv1d)
 CONV_LAYERS.register_module('Conv2d', module=nn.Conv2d)
 CONV_LAYERS.register_module('Conv3d', module=nn.Conv3d)
 CONV_LAYERS.register_module('Conv', module=nn.Conv2d)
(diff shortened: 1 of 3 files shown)


hellock:
It has already been registered as one of `UPSAMPLE_LAYERS`.

mmcv/cnn/bricks/conv.py
```diff
@@ -6,6 +6,7 @@
 CONV_LAYERS.register_module('Conv2d', module=nn.Conv2d)
 CONV_LAYERS.register_module('Conv3d', module=nn.Conv3d)
 CONV_LAYERS.register_module('Conv', module=nn.Conv2d)
+CONV_LAYERS.register_module('ConvTranspose2d', module=nn.ConvTranspose2d)
```
ZwwWayne, line 9: I suggest to use `ConvTranspose2d` wrapper implemented in MMCV rather than `nn.ConvTranspose2d`. The wrapper performs normally as `nn.ConvTranspose2d` in normal cases and could handle empty tensors.

GT9505:
The deconv is registered into CONV_LAYERS in order to be built in ConvModule.

mmcv/cnn/bricks/conv.py
GT9505: fixed

hellock (approved):

PR #586, opened 2020-09-27
Misc updates regarding Windows and unittests

wangruohui:


mmcv/runner/epoch_based_runner.py
```diff
@@ -169,7 +169,7 @@ def save_checkpoint(self,
             if platform.system() != 'Windows':
                 mmcv.symlink(filename, dst_file)
             else:
-                shutil.copy(filename, dst_file)
+                shutil.copy(filepath, dst_file)
```
hellock, line 172: We may consider add a utility method named `symlink_or_copy()`.

hellock (approved):

PR #592, opened 2020-09-28
[Improvement] Add markdown linter and fix linting errors

hellock:


ZwwWayne (approved):

.pre-commit-config.yaml
```diff
@@ -29,16 +29,21 @@ repos:
...
         args: ["--in-place", "--wrap-descriptions", "79"]
-  - repo: local
-    hooks:
-      - id: clang-format
-        name: clang-format
-        description: Format files with ClangFormat
-        entry: clang-format -style=google -i
-        language: system
-        files: \.(c|cc|cxx|cpp|cu|h|hpp|hxx|cuh|proto)$
+  # - repo: local
```
ZwwWayne, line 42: Will these commented codes be used in the future? If not, we may clean them.

hellock: It can be used for local formatting. I comment these lines because of the incompatibility with github actions (requires installing clang-format locally).

PR #593, opened 2020-09-29
add onnx support to roi_align and roi_pool

grimoire:
Hi,
This PR add onnx support to roi_align and roi_pool. The results is a little bit different between mmcv implementation, but better than nothing.

By the way, may I add onnxruntime of some ops like [this](https://github.com/onnx/tutorials/tree/master/PyTorchCustomOperator)? Should be benefit to mmdet->onnx conversion.


hellock:
Thanks for the contribution! It is nice to have more ops to be exportable. Custom ops with onnxruntime support are also welcome. @drcut may take a look at this PR.

drcut:
Thanks for your contribution! I have looked through the code and it seems to make sense! However, we decide to add a unittest for onnx part in mmcv, so it may take some time to merge your PR. If you are interested, we highly welcome you for contributing to unittest part. Once again, thanks for your interesting in MMCV!

hellock:
#608

hellock:
blocked by #608

RunningLeon (approved):
It passed unit tests. Approved.

PR #597, opened 2020-10-01
revert dcn

yhcao6:


yhcao6:
Fix https://github.com/open-mmlab/mmdetection/issues/3859

PR #602, opened 2020-10-05
Fix wrappers version comparison

ZwwWayne:


mmcv/cnn/bricks/wrappers.py
```diff
@@ -44,7 +48,7 @@ def forward(self, x):
             else:
                 return empty
 
-        return super().forward(x)
+        return super(Conv2d, self).forward(x)
```
hellock, line 51: This modification is unnecessary.

hellock (approved):

PR #608, opened 2020-10-10
add unittest for onnx convert

drcut:
Add unit test for the ops that be converted from pytorch to onnx

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,6 +108,8 @@ jobs:
       - name: Install Pillow
         run: pip install Pillow==6.2.2
         if: ${{matrix.torchvision == '0.4.2'}}
+      - name: Install ONNX
+        run: pip install onnx==1.6.0 onnxruntime==1.2.0
       - name: Install PyTorch
         run: pip install torch==${{matrix.torch}}+cpu torchvision==${{matrix.torchvision}}+cpu -f https://download.pytorch.org/whl/torch_stable.html
       - name: Install unittest dependencies
@@ -171,6 +173,8 @@ jobs:
       - name: Install Pillow
         run: pip install Pillow==6.2.2
         if: ${{matrix.torchvision == '0.4.2'}}
+      - name: Install ONNX
+        run: pip install onnx==1.6.0 onnxruntime==1.2.0
       - name: Install PyTorch
         run: pip install torch==${{matrix.torch}} torchvision==${{matrix.torchvision}} -f https://download.pytorch.org/whl/torch_stable.html
       - name: Install system dependencies
@@ -220,6 +224,8 @@ jobs:
       - name: Install Pillow
         run: pip install Pillow==6.2.2
         if: ${{matrix.torchvision == '0.4.2'}}
+      - name: Install ONNX
+        run: pip install onnx==1.6.0 onnxruntime==1.2.0
       - name: Install PyTorch
         run: pip install torch==${{matrix.torch}} torchvision==${{matrix.torchvision}} --no-cache-dir
       - name: Build and install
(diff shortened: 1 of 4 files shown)


grimoire:
Hi
The gpu version of onnxruntime is `onnxruntime-gpu`, and different version of onnxruntime-gpu official package is built up on different cuda/cudnn version.  `run: pip install onnx==1.6.0 onnxruntime==1.2.0` might cause error if you want to test on gpu.

drcut:
> Hi
> The gpu version of onnxruntime is `onnxruntime-gpu`, and different version of onnxruntime-gpu official package is built up on different cuda/cudnn version. Is `run: pip install onnx==1.6.0 onnxruntime==1.2.0` ok on `build_cuda`?

Thanks for your suggestion. build_cuda will not effect onnx part as we only run onnxruntime on cpu to get the output and comparing with output from pytorch cpu/cuda.

hellock:
@grimoire Would you help review this PR?

grimoire:
@hellock of cause, it's my pleasure!

tests/test_ops/test_onnx.py
```diff
@@ -0,0 +1,60 @@
...
+        ]
+        net_feed_input = list(set(input_all) - set(input_initializer))
+        assert (len(net_feed_input) == 2)
+        sess = rt.InferenceSession(onnx_file)
+        onnx_dets, _ = sess.run(None, {
+            'scores': scores.detach().numpy(),
+            'boxes': boxes.detach().numpy()
+        })
+        onnx_score = onnx_dets[:, 4]
+        assert np.allclose(pytorch_score, onnx_score, atol=1e-3)
```
grimoire, line 60: should the output boxes(`onnx_dets[:, :4]`) be tested? We can normalize the boxes by the mean and std of `np_boxes` before test.
And, I know this might not related to this PR, could we add `offset` to input `bboxes[:, 2:]` in NMSop.symbolic? Should give the same results, Than we can add test about offset!=0.

drcut: I don't think the first 4 columns (corresponding to the bbox position) should be tested. As there may be a slight numerical difference between Pytorch and ONNX. For example, if two bbox has the same scores, which should be put in the front of the output is not guaranteed. So that's why we only comparing values.

grimoire (approved):
LGTM

.github/workflows/build.yml
```diff
@@ -108,6 +108,8 @@ jobs:
       - name: Install Pillow
         run: pip install Pillow==6.2.2
         if: ${{matrix.torchvision == '0.4.2'}}
+      - name: Install ONNX
+        run: pip install onnx==1.6.0 onnxruntime==1.2.0
```
hellock, line 112: We can move the installation to `Install unittest dependencies` (L116).

tests/test_ops/test_onnx.py
```diff
@@ -0,0 +1,60 @@
+from functools import partial
+
+import numpy as np
+import onnx
+import onnxruntime as rt
+import torch
+import torch.nn as nn
+
+onnx_file = 'tmp.onnx'
```
hellock, line 9: We need to remove this file after testing.

tests/test_ops/test_onnx.py
```diff
@@ -0,0 +1,62 @@
...
+        net_feed_input = list(set(input_all) - set(input_initializer))
+        assert (len(net_feed_input) == 2)
+        sess = rt.InferenceSession(onnx_file)
+        onnx_dets, _ = sess.run(None, {
+            'scores': scores.detach().numpy(),
+            'boxes': boxes.detach().numpy()
+        })
+        onnx_score = onnx_dets[:, 4]
+        assert np.allclose(pytorch_score, onnx_score, atol=1e-3)
+        os.remove(onnx_file)
```
grimoire, line 62: Should we put the remove code before assert? If the test failed, the file will not be removed.

hellock (approved):

PR #612, opened 2020-10-13
[Feature]: Add custom operators support for onnxruntime in mmcv

grimoire:
Hi
This repo is an onnxruntime prototype of softnms. Like I mentioned in https://github.com/open-mmlab/mmcv/pull/593

export following environment below:
```
export MMCV_WITH_ORT=1
export ONNXRUNTIME_DIR=<path/to/onnxruntime>
```

than build mmcv with ops.

This is a **prototype** version, might not suitable to merge, there are something I want to discuss

- onnxruntime python wheel did not provide the library that needed to build the ops. So I add the library path through environment variable. Is there a better way to find these library(such as build onnxruntime from source)?
- the custom op need to register like this
  ```
    so1 = rt.SessionOptions()
    so1.register_custom_ops_library("mmcv/mmcv/_ext_ort.cpython-37m-x86_64-linux-gnu.so")
    ...
    sess = rt.InferenceSession(output_file, so1)
  ```
  Better to add a variable that point to the path of _ext_ort library. Where should I put the variable?
- ~~pytorch did not support onnx conversion of funtion with tuple output. Read https://github.com/pytorch/pytorch/issues/13834 for detail. So I implement the softnms ort in a dirty way(concate the dets and inds then slice them outside the function). They have fix in [this](https://github.com/pytorch/pytorch/commit/afa489dea973a476a2382e0a5f171f8b58944964) commit. Guess it will be public in next release. Should I keep the dirty version or wait and write a better one after next torch release?~~ (update to torch 1.7)
- only support onnxruntime>=1.5.1


RunningLeon (commented):
Thanks for all the contributing, pls. be noted that it may take a bit longer time to review this PR.

RunningLeon (changes requested):
Since #593 has been merged into master, pls. merge with master at first.
Secondly, could you add test units for soft_nms ?
Appreciated.

RunningLeon:
@grimoire 
The pytest result of  `soft_nms` has error : 
```
FAILED tests/test_ops/test_onnx.py::test_softnms - RuntimeError: tuple appears in op that does not forward tuples, unsupported kind: prim::PythonOp
```
As it is reported in issue[ #13834](https://github.com/pytorch/pytorch/issues/13834),  PR[#31595](https://github.com/pytorch/pytorch/pull/31595).
We have to double-check whether upgrading Pytorch to 1.7 could solve this problem and also check the downwards compatibility.

grimoire:
[This commit](https://github.com/grimoire/mmcv/commit/d3e62234a9229aafa4dd739b5a17c30411b6c68c#diff-45849aea01186d6181641ef8bbd8c0a2be6ed809) is tested on 1.6.
The output is the concatenation of [dets, inds].
```python
    @staticmethod
    def forward(ctx, boxes, scores, iou_threshold, sigma, min_score, method,
                offset):
        dets, inds = ext_module.softnms(
            boxes.cpu(),
            scores.cpu(),
            iou_threshold=float(iou_threshold),
            sigma=float(sigma),
            min_score=float(min_score),
            method=int(method),
            offset=int(offset))
        # return dets, inds

        # this return should be deprecated after torch update
        return torch.cat([dets, inds.to(dets.dtype).unsqueeze(1)], dim=1)
```
and I have to slice and do dtype cast again.
```pytorch
        dets = nms_out[:, :5]
        inds = nms_out[:, 5].long()
```
This one works, at the cost of computation (origin one does not need concatenation and dtype cast) (dets and **inds** are cpu tensors, might take more time when doing the dtype cast)

I don't think this is a good solution(people who do not need onnxruntime will also suffer from the extra cost). Hope there is a better way.

RunningLeon:
Concat and Slice may be a workaround. Or we could add an Assert line that this op can only be exported to onnx with Pytorch>=1.7.0.

mmcv/ops/csrc/onnxruntime/cpu/onnxruntime_register.cpp
```diff
@@ -0,0 +1,23 @@
+#include "../onnxruntime_register.h"
```
RunningLeon, line 1: Here we could include with `#include "onnxruntime_register.h"` since our `include_dir` is `mmcv/ops/csrc/onnxruntime`.

mmcv/ops/nms.py
```diff
@@ -39,6 +39,37 @@ def symbolic(g, bboxes, scores, iou_threshold, offset):
...
+            sigma=float(sigma),
+            min_score=float(min_score),
+            method=int(method),
+            offset=int(offset))
+        return dets, inds
+
+    @staticmethod
+    def symbolic(g, boxes, scores, iou_threshold, sigma, min_score, method,
+                 offset):
+        nms_out = g.op(
```
RunningLeon, line 60: Here we could add assertation that it should be exported to ONNX with Pytorch>=1.7.0.

tests/test_ops/test_onnx.py
```diff
@@ -58,6 +58,77 @@ def test_nms():
...
+    if not torch.cuda.is_available():
+        return
+    from packaging import version
+    from mmcv.ops import soft_nms, get_onnxruntime_op_path
+
+    # only support pytorch >= 1.7.0
+    if version.parse(torch.__version__) < version.parse('1.7.0'):
+        return
+
+    # only support onnxruntime >= 1.5.0
```
RunningLeon, line 71: Actually, there is no onnxruntime of 1.5.0, check [here](https://github.com/microsoft/onnxruntime/tags). Please update the version.

RunningLeon (approved):
LGTM.

RunningLeon:
@ZwwWayne See if you have any word on this PR.

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
+# Custom operators for onnxruntime in mmcv
+
+## Introduction of onnxruntime
+
+**ONNX RUNTIME** is a cross-platform inferencing and training accelerator compatible with many popular ML/DNN frameworks. Check its [github](https://github.com/microsoft/onnxruntime) for more information.
+
+## Introduction of onnx
+
+**ONNX** stands for **Open Neural Network Exchange**, which acts as *IR(Intermediate Representation)* for ML/DNN models from many frameworks. Check its [github](https://github.com/onnx/onnx) for more information.
```
ZwwWayne, line 9: Intermediate Representation (IR)

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Why include custom operators for onnxruntime in mmcv
```
ZwwWayne, line 11: mmcv -> MMCV

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
+# Custom operators for onnxruntime in mmcv
```
ZwwWayne, line 1: Custom Operators for ONNX runtime in MMCV

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+
+**ONNX RUNTIME** is a cross-platform inferencing and training accelerator compatible with many popular ML/DNN frameworks. Check its [github](https://github.com/microsoft/onnxruntime) for more information.
+
+## Introduction of onnx
+
+**ONNX** stands for **Open Neural Network Exchange**, which acts as *IR(Intermediate Representation)* for ML/DNN models from many frameworks. Check its [github](https://github.com/onnx/onnx) for more information.
+
+## Why include custom operators for onnxruntime in mmcv
+
+- To support verify correctness and precision of exported onnx models in onnxruntime.
```
ZwwWayne, line 13: To help verify the correctness of exported ONNX models in ONNX runtime.

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+- To support deploy onnx models with custom operators from `mmcv.ops` in onnxruntime.
```
ZwwWayne, line 14: To ease the deployment of ONNX models with custom operators from `mmcv.ops` in ONNX runtime.

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Custom operator list for onnxruntime supported in mmcv
```
ZwwWayne, line 16: List of operators for ONNX runtime supported in MMCV

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Prerequisite
+
+- Clone repository
+
+```bash
+git clone https://github.com/open-mmlab/mmcv.git
+```
+
+- Download `onnxruntime-linux-x64-1.5.1.tgz` from onnxruntime [releases](https://github.com/microsoft/onnxruntime/releases/tag/v1.5.1), extract it, expose `ONNXRUNTIME_DIR` and finally add the lib path to `LD_LIBRARY_PATH`:
```
ZwwWayne, line 34: get into `ONNXRUNTIME_DIR`, and

tests/test_ops/test_onnx.py
```diff
@@ -58,6 +59,80 @@ def test_nms():
     assert np.allclose(pytorch_score, onnx_score, atol=1e-3)
 
 
+def test_softnms():
+    if not torch.cuda.is_available():
+        warnings.warn('test_softnms should be ran only if cuda is available.')
+        return
```
hellock, line 65:
```python
@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is unavailable for test_softnms")
def test_softnms():
    ...
```

tests/test_ops/test_onnx.py
```diff
@@ -58,6 +59,80 @@ def test_nms():
...
+    boxes = torch.from_numpy(np_boxes)
+    scores = torch.from_numpy(np_scores)
+
+    configs = [[0.3, 0.5, 0.01, 'linear'], [0.3, 0.5, 0.01, 'gaussian'],
+               [0.3, 0.5, 0.01, 'naive']]
+
+    session_options = rt.SessionOptions()
+    session_options.register_custom_ops_library(ort_custom_op_path)
+
+    for iou, sig, mscore, m in configs:
```
hellock, line 96: Use meaningful variable names like `_sigma` instead of `sig`.

tests/test_ops/test_onnx.py
```diff
@@ -58,6 +59,80 @@ def test_nms():
...
+        return
+    from packaging import version
+    from mmcv.ops import soft_nms, get_onnxruntime_op_path
+
+    # only support pytorch >= 1.7.0
+    if version.parse(torch.__version__) < version.parse('1.7.0'):
+        warnings.warn('test_softnms should be ran with pytorch >= 1.7.0')
+        return
+
+    # only support onnxruntime >= 1.5.1
```
hellock, line 74: We can just assert the version >= 1.5.1.

docs/onnxruntime_op.md
```diff
@@ -0,0 +1,118 @@
...
+
+```python
+import os
+
+import numpy as np
+import onnxruntime as ort
+
+from mmcv.ops import get_onnxruntime_op_path
+
+assert os.path.exists(ort_custom_op_path)
```
hellock, line 71: `ort_custom_op_path` is undefined.

mmcv/ops/csrc/pytorch/nms.cpp
```diff
@@ -177,16 +179,18 @@ Tensor softnms_cpu(Tensor boxes, Tensor scores, Tensor dets,
...
     }
   }
-  return inds_t.slice(0, 0, nboxes);
+  // return inds_t.slice(0, 0, nboxes);
+  return {dets.slice(0, 0, nboxes), inds_t.slice(0, 0, nboxes)};
 }
 
-Tensor softnms(Tensor boxes, Tensor scores, Tensor dets, float iou_threshold,
-               float sigma, float min_score, int method, int offset) {
+std::vector<Tensor> softnms(Tensor boxes, Tensor scores, float iou_threshold,
```
hellock, line 186: The API change may affect downstream codebases.

tests/test_ops/test_onnx.py
RunningLeon: Only pytorch>=1.7.0 works for soft_nms, which has the `return` value of **a tuple with multiple tensors** in its `forward` function.

hellock: This is the version of onnxruntime.

RunningLeon: Ok, warnings would be replaced with assertations.

mmcv/ops/nms.py
```diff
@@ -191,17 +226,10 @@ def soft_nms(boxes,
...
-        inds = ext_module.softnms(
-            boxes.cpu(),
-            scores.cpu(),
-            dets.cpu(),
-            iou_threshold=float(iou_threshold),
-            sigma=float(sigma),
-            min_score=float(min_score),
-            method=method_dict[method],
-            offset=int(offset))
-    dets = dets[:inds.size(0)]
```
RunningLeon, line 204: Here should be compatible with `parrots`, another backend  Sensetime. It seems wrong. `dets` from new pytorch soft_nms are sliced but, from parrots are not. We should add sliced dets `dets = dets[:inds.size(0)]`at the end of line 227.

mmcv/ops/csrc/onnxruntime/onnxruntime_register.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
RunningLeon, line 1: It would be better to have consistent c++  `guard defines` in headers', just like existing code in the repo. Please refer to [here](https://google.github.io/styleguide/cppguide.html#The__define_Guard).

mmcv/ops/csrc/onnxruntime/soft_nms.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
RunningLeon, line 1: Here too.

ZwwWayne (approved):

PR #613, opened 2020-10-13
fix wrappers when using parrots

RyanXLi:
This commits fix the exception in wrappers.py when using parrots as the torch backend.

### DIFF
--- a/mmcv/cnn/bricks/wrappers.py
+++ b/mmcv/cnn/bricks/wrappers.py
@@ -20,6 +20,10 @@ else:
     TORCH_VERSION = tuple(int(x) for x in torch.__version__.split('.')[:2])
 
 
+def obsolete_torch_version(torch_version, version_threshold):
+    return torch_version == 'parrots' or torch_version <= version_threshold
+
+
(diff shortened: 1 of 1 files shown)


mmcv/cnn/bricks/wrappers.py
```diff
@@ -34,7 +37,8 @@ def backward(ctx, grad):
 class Conv2d(nn.Conv2d):
 
     def forward(self, x):
-        if x.numel() == 0 and TORCH_VERSION <= (1, 4):
+        if x.numel() == 0 and (TORCH_VERSION == 'parrots' or TORCH_VERSION <=
```
Johnson-Wang, line 40: I suggest defining a `obsolete_torch_version(version_threshold)` function and change this line to `if x.numel() == 0 and obsolete_torch_version((1, 4)):`

ZwwWayne (approved):

ZwwWayne (commented):
Please test the CI of mmdetection in the environment with this version of MMCV installed.

Johnson-Wang (approved):

PR #616, opened 2020-10-15
Skip opencv requirement if it's already installed in the env (i.e. via conda)

daavoo:


setup.py
```diff
@@ -119,12 +119,15 @@ def gen_packages_items():
...
 
-# If first not installed install second package
-CHOOSE_INSTALL_REQUIRES = [('opencv-python-headless>=3', 'opencv-python>=3')]
-
 install_requires = parse_requirements()
-for main, secondary in CHOOSE_INSTALL_REQUIRES:
-    install_requires.append(choose_requirement(main, secondary))
+try:
+    # OpenCV installed via conda.
+    import cv2
```
hellock, line 125: We can check the minimum version here.

daavoo: I used RuntimeError but I'm not sure what error should be raised or how it should be handled if incorrect version is installed.

hellock: It is okay to me.

hellock (approved):

Issue #611, opened 2020-10-12
TypeError: symbolic() missing 2 required positional arguments: 'bias' and 'im2col_step' 


twmht:
Hi,

I want to convert deform convolution to onnx, as the following

```python
from mmcv.ops import DeformConv2dPack
import torch

model = DeformConv2dPack(3, 12, 3).cuda()
model.eval()
x = torch.randn(1,3,112,112).cuda()
torch.onnx.export(model, x, './test.onnx', input_names=['data'], output_names = ['out'])
```

but this would throw the errors

```bash
TypeError: symbolic() missing 2 required positional arguments: 'bias' and 'im2col_step' (occurred when translating DeformConv2dFunction)
```

any idea?

drcut:
Hi,
As far as I know, the official ONNX does not support deformConv, so we have defined a specific op ```MMCVDeformConv2d``` which is only suitable for our own backend device. So I highly suggest you to first figure out what is your target backend, whether it supports deformConv, what is the deformCONV ONNX format for this backend device?

twmht:
@drcut 

I know exactly what you mean. I am going to export dcn to tensorrt, so I have to include dcn symbol and op in the output onnx file.

although tensorrt does not support dcn for now, I can add it as a custom layer.

I think this error might be a bug since you have defined a symbolic function for that, but it can't work as expected.

drcut:
> @drcut
> 
> I know totally what you mean. I am going to export dcn to tensorrt, so I have to include dcn symbol and op in the output onnx file.
> 
> although tensorrt does not support dcn for now, I can add it as a custom layer.
> 
> I think this error might be a bug since you have defined a symbolic function for that, but it can't work as expected.

Yes you are right. To be honest, this code is written by some former person and I think they should be deprecated or just deleted or raise a NotImplement Error. We will fix this bug according to your suggestions! Thanks for your contribution!

drcut:
@RunningLeon Would you please fix this bug? Thanks

twmht:
@drcut 

I fix that and have submitted a PR (https://github.com/open-mmlab/mmcv/pull/624).

RunningLeon:
Fixed in PR ([#624](https://github.com/open-mmlab/mmcv/pull/624)).

PR #624, opened 2020-10-24
[fix] Missing arguments when converting dcn to onnx

twmht:
fix issue #611

### DIFF
--- a/mmcv/ops/deform_conv.py
+++ b/mmcv/ops/deform_conv.py
@@ -21,19 +21,18 @@ class DeformConv2dFunction(Function):
 
     @staticmethod
     def symbolic(g, input, offset, weight, stride, padding, dilation, groups,
-                 deform_groups, bias, im2col_step):
+                 deform_groups):
         return g.op(
             'MMCVDeformConv2d',
             input,
             offset,
             weight,
-            stride=stride,
-            padding=padding,
-            dilation=dilation,
-            groups=groups,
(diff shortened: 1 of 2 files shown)


mmcv/ops/deform_conv.py
```diff
@@ -21,19 +21,18 @@ class DeformConv2dFunction(Function):
 
     @staticmethod
     def symbolic(g, input, offset, weight, stride, padding, dilation, groups,
-                 deform_groups, bias, im2col_step):
+                 deform_groups):
```
RunningLeon, line 24: It would be better to keep `bias` and `im2col_step` for future extension. Pls. keep them as default arguments

mmcv/ops/deform_conv.py
```diff
@@ -44,21 +43,19 @@ def forward(ctx,
                 padding=0,
                 dilation=1,
                 groups=1,
-                deform_groups=1,
-                bias=False,
-                im2col_step=32):
+                deform_groups=1
+                ):
```
RunningLeon, line 47: It would be better to keep `bias` and `im2col_step` for future extension. Pls. keep them as default arguments

RunningLeon (changes requested):
Pls. keep `bias` and `im2col_step` for future extension. Could keep them as default arguments.
Thanks a lot.

RunningLeon (approved):

PR #635, opened 2020-11-03
[Feature] Add hook msg for checkpoint hook

xvjiarui:


hellock:
Task linked: [CU-f6rv80 MMCV hook share info](https://app.clickup.com/t/f6rv80)

mmcv/runner/hooks/checkpoint.py
```diff
@@ -60,6 +60,16 @@ def _save_checkpoint(self, runner):
             self.out_dir = runner.work_dir
         runner.save_checkpoint(
             self.out_dir, save_optimizer=self.save_optimizer, **self.args)
+        if runner.meta is not None:
+            if self.by_epoch:
+                cur_ckpt_filename = self.args.get(
+                    'filename_tmpl', 'epoch_{}.pth').format(runner.epoch + 1)
```
Johnson-Wang, line 66: Why `'filename_tmpl'` is passed using implicit `kwargs`?

hellock: Arguments of `runner.save_checkpoint()` are hidden from CheckpointHook, and different runners have different default values.

xvjiarui: It will be used for `runner.save_checkpoint(self.out_dir, save_optimizer=self.save_optimizer, **self.args)` also.

hellock (approved):

PR #642, opened 2020-11-06
update README.md for torch 1.7 installation

ChiangYintso:


hellock:
Thanks for your contribution! We have not released the prebuilt package with PyTorch 1.7, thus it it unavailable now. The readme will be updated as soon as the packages are uploaded.

hellock (approved):

PR #643, opened 2020-11-08
Skip backpropogation if grad_norm is Nan in check

ZwwWayne:


mmcv/runner/hooks/optimizer.py
```diff
@@ -31,6 +31,10 @@ def after_train_iter(self, runner):
                 # Add grad norm to the logger
                 runner.log_buffer.update({'grad_norm': float(grad_norm)},
                                          runner.outputs['num_samples'])
+            elif torch.isnan(grad_norm).any():
```
hellock, line 34: Will this slow down the training? By how much?

hellock:
The linting failed.

ZwwWayne:
The nan issue of loss is solved in an essential place in MMDet & MMDet3D. This PR is unnecessary because it still cannot protect the training. Therefore, this PR is closed. More essential fix of MMDet & MMDet3D can be proposed if necessary after discussion.

PR #648, opened 2020-11-10
[Enhance]: Better support for pytorch2onnx for detection models

RunningLeon:
Support pytorch2onnx conversion for RetinaNet, Yolov3, SSD, Faster-RCNN.
1. Support op `new_full`;
2. Fix _interger divide error_ for `carafe`;
3. Support `roi_align` exporting to onnxruntime;


ZwwWayne (approved):

ZwwWayne:
Looks good to me. Please @Johnson-Wang have a check.

mmcv/ops/nms.py
```diff
@@ -248,14 +247,14 @@ def batched_nms(boxes, scores, idxs, nms_cfg, class_agnostic=False):
         boxes_for_nms = boxes
     else:
         max_coordinate = boxes.max()
-        offsets = idxs.to(boxes) * (max_coordinate + 1)
+        offsets = idxs.to(boxes) * (max_coordinate + torch.tensor(1).to(boxes))
```
hellock, line 250: We may leave a comment here.

RunningLeon: Basically, it converts python scalar int variable to tensor for tracing. In case of converting models to onnx using cuda, it may be more friendly. It's also the same as what torchvision did in [here](https://github.com/pytorch/vision/blob/b2171653f83543d380715ddcffb5ebb2d05efa2e/torchvision/ops/boxes.py#L86)

hellock:
There are linting errors and conflicts.

RunningLeon:
> There are linting errors and conflicts.

Resolved.

hellock:
There are still linting errors.

ZwwWayne (approved):

ZwwWayne:
The PR can be merged after fixing linting errors.

Johnson-Wang (approved):

ZwwWayne (approved):

mmcv/ops/carafe.py
```diff
@@ -263,7 +263,7 @@ def init_weights(self):
     def kernel_normalizer(self, mask):
         mask = F.pixel_shuffle(mask, self.scale_factor)
         n, mask_c, h, w = mask.size()
-        mask_channel = int(mask_c / (self.up_kernel * self.up_kernel))
+        mask_channel = int(mask_c / float(self.up_kernel**2))
```
ZwwWayne, line 266: Is the change here necessary? If so, please add a comment in the code to indicate the reason.

mmcv/ops/nms.py
ZwwWayne: Please leave the comment in the code thus other developers will be notified when doing the related modifications.

PR #669, opened 2020-11-25
[feature] Add OP ball query

encore-zhou:
1. add op `ball_query`
2. add unittest for op `ball_query`


zhouzaida:
closed by #1332

PR #685, opened 2020-12-03
[Feature] Add clamp activation layer.

nbei:


ycxioooong (approved):

mmcv/cnn/bricks/activation.py
```diff
@@ -10,6 +11,39 @@
     ACTIVATION_LAYERS.register_module(module=module)
 
 
+@ACTIVATION_LAYERS.register_module(name='clip')
+@ACTIVATION_LAYERS.register_module(name='clamp')
```
hellock, line 15: Like other activation layers, we can name it Clamp or Clip.

mmcv/cnn/bricks/activation.py
```diff
@@ -10,6 +11,39 @@
...
+@ACTIVATION_LAYERS.register_module()
+class ClampLayer(nn.Module):
```
hellock, line 17: Maybe we just name it `Clamp`.

PR #692, opened 2020-12-05
Allow the initial value to be set in dynamic mode

ZwwWayne:
This PR modifies the `FP16OptimizerHook` to allow the initial scale value to be set in dynamic mode.
This modification is more practical since even in the dynamic mode we will not use 2**32 as an initial value sometimes. This also provides more flexible usages.


mmcv/runner/hooks/optimizer.py
```diff
@@ -48,8 +48,10 @@ class Fp16OptimizerHook(OptimizerHook):
     Refer to https://arxiv.org/abs/1710.03740 for more details.
 
     Args:
-        loss_scale (float | str): Scale factor multiplied with loss. If
-            'dynamic' is specified, then dynamic loss scaling will be used.
+        loss_scale (float | str | dict): Scale factor multiplied with loss.
```
hellock, line 51: Explain the behaviors when `loss_scale` are different types.

ycxioooong (approved):

Issue #700, opened 2020-12-08
scontrol: not found in slurm using pyxis container


minooei:
Hi
I'm trying to run mmdetection on slurm using pyxis container and getting this error:
```
  dist.init_process_group(backend=backend)
  File "/opt/conda/lib/python3.6/site-packages/torch/distributed/distributed_c10d.py", line 423, in init_process_group
    store = TCPStore(master_addr, master_port, world_size, start_daemon, timeout)
ValueError: host not found: Temporary failure in name resolution
```
I investigate the problem and I think the problem is related to using scontrol command since this command is not available in pyxis container. I tried a simple script with this [line](https://github.com/open-mmlab/mmcv/blob/50a33950a4b23c614152696e6f979ae978233432/mmcv/runner/dist_utils.py#L57)
```
addr = subprocess.getoutput(
        f'scontrol show hostname {node_list} | head -n1')
``` 

which results in: ```/bin/dash: 1: scontrol: not found```

what is the purpose of this [line](https://github.com/open-mmlab/mmcv/blob/50a33950a4b23c614152696e6f979ae978233432/mmcv/runner/dist_utils.py#L57)? 
is there an alternative not using scontrol command?

minooei:
after commenting this [line](https://github.com/open-mmlab/mmcv/blob/50a33950a4b23c614152696e6f979ae978233432/mmcv/runner/dist_utils.py#L67) my code works as intended.

maybe it's better to check os.environ['MASTER_ADDR'] before assignment like line 62:
```
    if 'MASTER_ADDR' in os.environ:
        pass
    else:
        os.environ['MASTER_ADDR'] = addr
```

xvjiarui:
HI @minooei 
Thanks for pointing out! 
Would you like to create a PR to fix it? 
As for the `scontrol`, we use it on our own slurm cluster. However, the slurm cluster may be configured differently. I think your proposal looks good to me.

minooei:
yes sure, I submitted a [PR](https://github.com/open-mmlab/mmcv/pull/705).

PR #705, opened 2020-12-11
check MASTER_ADDR in the environment variable before assignment

minooei:
addressing issue [700](https://github.com/open-mmlab/mmcv/issues/700)

### DIFF
--- a/mmcv/runner/dist_utils.py
+++ b/mmcv/runner/dist_utils.py
@@ -64,7 +64,10 @@ def _init_dist_slurm(backend, port=None):
     else:
         # 29500 is torch.distributed default port
         os.environ['MASTER_PORT'] = '29500'
-    os.environ['MASTER_ADDR'] = addr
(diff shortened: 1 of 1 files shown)


mmcv/runner/dist_utils.py
```diff
@@ -64,7 +64,10 @@ def _init_dist_slurm(backend, port=None):
     else:
         # 29500 is torch.distributed default port
         os.environ['MASTER_PORT'] = '29500'
-    os.environ['MASTER_ADDR'] = addr
+    if 'MASTER_ADDR' in os.environ:
+        pass  # use MASTER_ADDR in the environment variable
+    else:
+        os.environ['MASTER_ADDR'] = addr
```
hellock, line 70:
```python
# use MASTER_ADDR in the environment variable if it already exists
if 'MASTER_ADDR' not in os.environ:
    os.environ['MASTER_ADDR'] = addr
```

hellock (approved):

PR #707, opened 2020-12-12
[Refactor] remove the c implementation of flow_warp

nbei:
Fix bugs in `flow_warp` function in a way of replacing it with numpy implementation.


hellock:
You may need to check if the documentation needs to be updated accordingly.

mmcv/video/optflow.py
```diff
@@ -151,19 +150,43 @@ def flow_warp(img, flow, filling_value=0, interpolate_mode='nearest'):
...
-    img_float = img.astype(np.float64)
-
-    out = flow_warp_c(
-        img_float,
-        flow.astype(np.float64),
-        filling_value=filling_value,
-        interpolate_mode=interpolate_mode)
-
-    return out
+    height = flow.shape[0]
```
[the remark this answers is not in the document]
hellock: and add a warning that this method is only for prototyping

setup.py
```diff
@@ -1,15 +1,11 @@
 import glob
 import os
 import re
-import setuptools
 from pkg_resources import DistributionNotFound, get_distribution
 from setuptools import dist, find_packages, setup
 
 dist.Distribution().fetch_build_eggs(['Cython', 'numpy>=1.11.1'])
```
hellock, line 7: Cython is not needed.

docs/build.md
```diff
@@ -101,11 +101,11 @@ MMCV can be built in three ways:
 
 1. Lite version (without ops)
 
-   In this way, only `flow_warp` module will be compiled as a Cython extension.
+   In this way, there does not exist any compilation.
```
hellock, line 104: In this way, no custom ops are compiled and mmcv is a pure python package.

setup.py
```diff
@@ -1,14 +1,8 @@
...
-from setuptools import dist, find_packages, setup
-
-dist.Distribution().fetch_build_eggs(['Cython', 'numpy>=1.11.1'])
-
-import numpy  # NOQA: E402  # isort:skip
-from Cython.Build import cythonize  # NOQA: E402  # isort:skip
+from setuptools import find_packages, setup
 
 EXT_TYPE = ''
 try:
```
hellock, line 8: Since no cython compilation is needed, the following logic can be simplified.

nbei: Since we need to support non-torch compilation, I may think the logic should be kept as the previous one.

hellock: We can just avoid the dependency of cython like the current implementation.

hellock (approved):

ycxioooong (approved):

PR #710, opened 2020-12-14
[Feature]Support pavi modelcloud backend

LXXXXR:


mmcv/runner/checkpoint.py
```diff
@@ -217,6 +240,10 @@ def _load_checkpoint(filename, map_location=None):
         checkpoint = _process_mmcls_checkpoint(checkpoint)
     elif filename.startswith(('http://', 'https://')):
         checkpoint = load_url_dist(filename)
+    elif filename.startswith('pavimodelcloud://'):
```
ZwwWayne, line 243: pavimodelcloud -> pavi

mmcv/fileio/file_client.py
```diff
@@ -1,5 +1,7 @@
 import inspect
+import os
```
hellock, line 2: import os.path as osp

PR #724, opened 2020-12-16
[Feature]: add custom op roialign for onnxruntime

tangyanf:
add onnxruntime custom op MMCVRoIAlign,so we can use mmcvroialign in onnxruntime;
use mmcv domain as custom op's domain, this is required for torch to export custom op;
An ONNX opset consists of a domain name and a version number. If we are using existing ONNX operators (from the default ONNX domain), we do not need to add the domain name prefix. In mmcv, we want to use an op from mmcv custom opset. Therefore, we need to add the domain name as a prefix in the following format: "<domain_name>::<onnx_op>"


RunningLeon:
Please sign CLA and fix lint at first.

RunningLeon:
@tangyanf 
Since #612 has been merged into master and this PR is based on #612, we could merge with the newest master at first.

RunningLeon (approved):
LGTM

docs/onnxruntime_op.md
```diff
@@ -15,9 +15,10 @@
 
 ## List of operators for ONNX Runtime supported in MMCV
 
-| Operator |  CPU  |  GPU  | Note  |
-| :------: | :---: | :---: | :---: |
-| SoftNMS  |   Y   |   N   | None  |
+| Operator |  CPU  |  GPU  |       Note       |
+| :------: | :---: | :---: | :--------------: |
+| SoftNMS  |   Y   |   N   | commit `94810f2` |
```
ZwwWayne, line 20: Nit: May add link for commit `94810f2`

ZwwWayne (approved):

PR #731, opened 2020-12-18
Allow to replace nested tuple and list via options

Johnson-Wang:


mmcv/utils/config.py
```diff
@@ -506,12 +506,73 @@ def _parse_int_float_bool(val):
...
+
+        Examples:
+            >>> DictAction._parse_iterable('1,2,3')
+            [1, 2, 3]
+            >>> DictAction._parse_iterable('[a, b, c]')
+            ['a', 'b', 'c']
+            >>> DictAction._parse_iterable('[(1, 2, 3), [a, b], c]')
+            [(1, 2, 3), ['a', 'b], 'c']
+        """
+
```
xvjiarui, line 529: I suggest checking the balancing of brackets in the first place, then we don't need to do it anymore.

mmcv/utils/config.py
```diff
@@ -506,12 +506,73 @@ def _parse_int_float_bool(val):
...
     def __call__(self, parser, namespace, values, option_string=None):
         options = {}
         for kv in values:
             key, val = kv.split('=', maxsplit=1)
-            val = [self._parse_int_float_bool(v) for v in val.split(',')]
-            if len(val) == 1:
-                val = val[0]
-            options[key] = val
+            options[key] = self._parse_iterable(val)
```
xvjiarui, line 577: We may also update the docstring of this class.

xvjiarui (approved):

ZwwWayne (approved):

PR #737, opened 2020-12-20
fix grad_output uncontiguous

jshilong:
Some complex head architecture may cause grad_output uncontiguous, eg.dii head in sparsercnn

### DIFF
--- a/mmcv/ops/roi_align.py
+++ b/mmcv/ops/roi_align.py
@@ -94,7 +94,7 @@ class RoIAlignFunction(Function):
     def backward(ctx, grad_output):
         rois, argmax_y, argmax_x = ctx.saved_tensors
         grad_input = grad_output.new_zeros(ctx.input_shape)
-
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

mmcv/ops/roi_align.py
```diff
@@ -94,7 +94,7 @@ def forward(ctx,
     def backward(ctx, grad_output):
         rois, argmax_y, argmax_x = ctx.saved_tensors
         grad_input = grad_output.new_zeros(ctx.input_shape)
-
+        grad_output = grad_output.contiguous()
```
ZwwWayne, line 97: Add comments above this line in the code to indicate that in some cases the `grad_output` is incontiguous thus this modification is necessary.

kennymckormick:
Also met that bug

hellock (approved):

PR #738, opened 2020-12-21
[WIP] add common test function of MM repos

congee524:


dreamerlin:
Not only for mmaction2. It extract all common function for MM repos

PR #743, opened 2020-12-21
Add common testing function of MM repos

congee524:
Add common testing function of MM repos

### DIFF
--- a/mmcv/utils/testing.py
+++ b/mmcv/utils/testing.py
@@ -1,85 +1,116 @@
 # Copyright (c) Open-MMLab.
+from collections.abc import Iterable
+from typing import Any, Dict, List
 
 
-def check_dict(result_dict, key_list, value_list):
-    """Check if the result_dict is correct. The function does not support the
-    case that both values in result_dict and value_list are nonstandard python
-    data types.
+def _any(judge_result):
+    """Since built-in ``any`` works only when the element of iterable is not
+    iterable, implement the funciton."""
+    if not isinstance(judge_result, Iterable):
+        return judge_result
+
+    try:
+        for element in judge_result:
+            if _any(element):
+                return True
+    except TypeError:
+        # Maybe encouter the case: torch.tensor(True) | torch.tensor(False)
+        if judge_result:
+            return True
+    return False
+
+
+def assert_dict_contains_subset(dict_obj: Dict[Any, Any],
+                                expected_subset: Dict[Any, Any]) -> bool:
+    """Check if the dict_obj contains the expected_subset.
 
     Args:
-        result_dict(dict): Dict to be checked.
-        key_list(tuple): Tuple of checked keys.
-        value_list(tuple): Tuple of target values.
+        dict_obj (Dict[Any, Any]): Dict object to be checked.
+        expected_subset (Dict[Any, Any]): Subset expected to be contained in
+            dict_obj.
 
     Returns:
-        bool: Whether the result_dict is correct.
+        bool: Whether the dict_obj contains the expected_subset.
     """
-    for key, value in zip(key_list, value_list):
-        if result_dict[key] != value:
+
+    for key, value in expected_subset.items():
+        uneq = dict_obj[key] != value if key in dict_obj.keys() else True
+        if _any(uneq):
             return False
     return True
 
 
-def check_class_attr(obj, attr_list, value_list):
+def assert_attrs_equal(obj: Any, expected_attrs: Dict[str, Any]) -> bool:
     """Check if attribute of class object is correct.
 
     Args:
-        obj(object): Class object to be checked.
-        attr_list(tuple[str]): Tuple of inner attribute names ot be checked.
-        value_list(tuple): Tuple of target values.
+        obj (object): Class object to be checked.
+        expected_attrs (Dict[str, Any]): Dict of the expected attrs.
 
     Returns:
         bool: Whether the attribute of class object is correct.
     """
-    for attr, value in zip(attr_list, value_list):
-        if not hasattr(obj, attr) or getattr(obj, attr) != value:
+    for attr, value in expected_attrs.items():
+        if not hasattr(obj, attr):
             return False
-    return True
-
-
-def assert_keys_contain(result_keys, target_keys):
-    """Check if all elements in target_keys is in result_keys."""
-    return set(target_keys).issubset(set(result_keys))
-
-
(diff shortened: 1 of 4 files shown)


dreamerlin (approved):

mmcv/utils/testing.py
```diff
@@ -0,0 +1,91 @@
...
+    """Check if the module is a norm layer."""
+    from .parrots_wrapper import _BatchNorm, _InstanceNorm
+    from torch.nn import GroupNorm, LayerNorm
+    norm_layer_candidates = (_BatchNorm, _InstanceNorm, GroupNorm, LayerNorm)
+    return is_block(module, norm_layer_candidates)
+
+
+def is_all_zeros(module):
+    """Check if the weight (and bias) of the module is all zero."""
+    import torch
```
xvjiarui, line 81: We may remove the import by `module.weight.data.allclose(torch.zeros_like(module.weight.data)`

congee524: why we can remove the import, since we used `torch.zeros_like()`?

xvjiarui: Maybe `new_zeros`. But both works for me.

congee524: \(^o^)/~

mmcv/utils/testing.py
```diff
@@ -0,0 +1,92 @@
...
+    Args:
+        result_dict(dict): Dict to be checked.
+        key_list(tuple): Tuple of checked keys.
+        value_list(tuple): Tuple of target values.
+
+    Returns:
+        bool: Whether the result_dict is correct.
+    """
+    for key, value in zip(key_list, value_list):
+        if result_dict[key] != value:
```
ZwwWayne, line 16: What if the value in the dict is a Tensor? Is this supported?

congee524: No, it cannot support it. 
But the situation, that the elements of dict are tensors, seems to be rare. Should I deal with tensor alone, or should I say the case in the comments? Or is there a better way?

congee524: I cannot come up with the method that assert the equality of two tensors without `import torch`, unless we judge whether the comparison result is boolean type alone.

ZwwWayne: The key issue is when `result_dict[key] != value` is a tensor or array. Since both tensor and array are Iterable, we may use all() for iterable, like below:
```python
from collections.abc import Iterable
uneq = result_dict[key] != value
uneq = all(uneq) if isinstance(uneq, Iterable) else uneq

if uneq:
    xxxxx
```

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+            return False
+    return True
+
+
+def check_class_attr(obj, attr_list, value_list):
```
hellock, line 23: Suggested signature:
```python
def assert_attrs_equal(obj: Any, expected_attrs: dict[str, Any]) -> bool:
```

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) Open-MMLab.
+
+
+def check_dict(result_dict, key_list, value_list):
```
hellock, line 4: Suggested signature:
```python
def assert_dict_contains_subset(dict_obj: dict[Any, Any], expected_subset: dict[Any, Any]) -> bool:
```

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns:
+        bool: Whether the attribute of class object is correct.
+    """
+    for attr, value in zip(attr_list, value_list):
+        if not hasattr(obj, attr) or getattr(obj, attr) != value:
+            return False
+    return True
+
+
+def assert_keys_contain(result_keys, target_keys):
```
hellock, line 40:
```python
def assert_dict_has_keys(obj, expected_keys)
```

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+    """Check if all elements in target_keys is in result_keys."""
+    return set(target_keys).issubset(set(result_keys))
+
+
+def assert_keys_equal(result_keys, target_keys):
+    """Check if target_keys is equal to result_keys."""
+    return set(result_keys) == set(target_keys)
+
+
+def check_norm_state(modules, train_state):
```
hellock, line 50: This method is not so common.

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+    """Check if norm layer is in correct train state."""
+    from .parrots_wrapper import _BatchNorm
+    for mod in modules:
+        if isinstance(mod, _BatchNorm):
+            if mod.training != train_state:
+                return False
+    return True
+
+
+def is_block(module, block_candidates):
```
hellock, line 60: This method is unnecessary since it does not avoid redundant codes.

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+        module(nn.Module): The module to be checked.
+        block_candidates(tuple[nn.module]): Tuple of block candidates.
+
+    Returns:
+        bool: Whether the module is the specified block.
+    """
+    return isinstance(module, block_candidates)
+
+
+def is_norm(module):
```
hellock, line 73:
```python
def assert_is_norm_layer(module):
```

mmcv/utils/testing.py
```diff
@@ -0,0 +1,94 @@
...
+    """Check if the module is a norm layer."""
+    from .parrots_wrapper import _BatchNorm, _InstanceNorm
+    from torch.nn import GroupNorm, LayerNorm
+    norm_layer_candidates = (_BatchNorm, _InstanceNorm, GroupNorm, LayerNorm)
+    return is_block(module, norm_layer_candidates)
+
+
+def is_all_zeros(module):
```
hellock, line 81:
```python
def assert_params_all_zeros(module)
```

hellock:
Use unified naming conventions for methods with similar functionalities.

mmcv/utils/testing.py
congee524: Encountered some problems, using a less elegant way

congee524:
kindly ping :p

hellock (approved):

ZwwWayne (approved):

xvjiarui (approved):

PR #751, opened 2020-12-23
[Feature] add onnx simplify

RunningLeon:
This PR adds `mmcv.onnx.simplify` module, which is based on [onnx-simplifier](https://github.com/daquexian/onnx-simplifier), an useful tool to simplify and optimize ONNX models. 
Because we want to simplify models with MMCV's custom ops, which should be registered before simplifying. Thus, we have to modify source codes from [onnx-simplifier](https://github.com/daquexian/onnx-simplifier), and include them in mmcv repository.
This PR includes:
1. `mmcv.onnx.simplify` module;
2. Simple unit test;
3. Simple doc;

After this PR merged, `pytorch2onnx.py` in many projects such as mmdetection could be updated.


mmcv/onnx/simplify/core.py
```diff
@@ -0,0 +1,538 @@
...
+    """Map element type from ONNX to dtype of numpy.
+
+    Args:
+        elem_type (int): Element type index in ONNX.
+
+    Returns:
+        int: Data type in numpy.
+    """
+    # from https://github.com/onnx/onnx/blob/
+    # e5e9a539f550f07ec156812484e8d4f33fb91f88/onnx/onnx.proto#L461
```
hellock, line 93: For long urls, we can use `noqa` to disable the linter.
```python
# from https://github.com/onnx/onnx/blob/e5e9a539f550f07ec156812484e8d4f33fb91f88/onnx/onnx.proto#L461  # noqa: E501
```

tests/test_ops/test_onnx.py
```diff
@@ -182,3 +183,26 @@ def warpped_function(torch_input, torch_rois):
         # allclose
         os.remove(onnx_file)
         assert np.allclose(pytorch_output, onnx_output, atol=1e-3)
+
+
+def test_simplify():
+    try:
+        from mmcv.onnx import simplify
+    except ImportError:
```
hellock, line 191: What's this case?

tests/test_ops/test_onnx.py
```diff
@@ -182,3 +183,26 @@ def warpped_function(torch_input, torch_rois):
...
+    dummy_input = torch.randn(2, 3, 4, 5)
+    torch.onnx.export(net, dummy_input, onnx_file, input_names=['input'])
+    ori_onnx_model = onnx.load(onnx_file)
+
+    feed_input = [{'input': dummy_input.detach().cpu().numpy()}]
+    slim_onnx_model = simplify(ori_onnx_model, feed_input, onnx_file)
+    numel_before = len(ori_onnx_model.graph.node)
+    numel_after = len(slim_onnx_model.graph.node)
+    assert numel_before == 18 and numel_after == 1, 'Simplify failed.'
+    os.remove(onnx_file)
```
hellock, line 208: If the assertion failed, this line will not be executed.

mmcv/onnx/simplify/common.py
```diff
@@ -0,0 +1,43 @@
+import copy
+import warnings
+
+import onnx
+
+
+def add_suffix2name(ori_model, suffix='__', verify=True):
```
hellock, line 7: It seems that this method is only used once. Can it be merged into `core.py`?

docs/onnx.md
```diff
@@ -0,0 +1,40 @@
...
+
+opset_version = 11
+register_extra_symbolics(opset_version)
+```
+
+## ONNX simplify
+
+### Intention
+
+`mmcv.onnx.simplify` is based on [onnx-simplifier](https://github.com/daquexian/onnx-simplifier), which is a useful tool to make exported ONNX models slimmer by performing a series of optimization. However, for Pytorch models with custom op from `mmcv`, it would break down. Thus, custom op for ONNX Runtime should be registered.
```
hellock, line 21: Pytorch -> PyTorch
custom op -> custom ops

tests/test_ops/test_onnx.py
RunningLeon: Would be removed. Thought for backward compatibility.

mmcv/onnx/simplify/common.py
RunningLeon: The intention is that code by mmcv should be added into `common.py`.
While `core.py` would be kept updated with originally [onnx-simplifier](https://github.com/daquexian/onnx-simplifier).

hellock (approved):

daquexian:
Hi @RunningLeon , thanks for your excellent work! I'm the maintainer of onnx-simplifier (and onnx optimizer). Will it help if I support custom op registration in onnx-simplifier? mmcv will benefit from the future update of onnx-simplifier if it directly uses onnx-simplifier package.

RunningLeon:
Hi, @daquexian,
Very glad to hear from you. Totally agree with you! It would definitely benefit a lot if `onnx-simplifier` supports customer ops registration in the future.
Thanks.

PR #760, opened 2020-12-25
add model registry

xvjiarui:
This PR refactored `Registry` class in following  ways:

1. make `build` function an attribute of `Registry` class.

before:
```python
model = build_from_cfg(cfg, MODELS)
```
after:
```python
model = MODELS.build(cfg)
```
`build` function could be configured by passing argument.  

2. Added hierarchy structure for Registry. 

For example, we can define `MODEL` registry in both [MMDetection](https://github.com/open-mmlab/mmdetection) and [MMClassification](https://github.com/open-mmlab/mmclassification)  as followed:
In MMDetection:

```python
from mmcv.utils import Registry
from mmcv.cnn import MODELS as MMCV_MODELS
MODELS = Registry('model', parent=MMCV_MODELS)

@MODELS.register_module()
class NetA(nn.Module):
    def forward(self, x):
        return x
```

In MMClassification:

```python
from mmcv.utils import Registry
from mmcv.cnn import MODELS as MMCV_MODELS
MODELS = Registry('model', parent=MMCV_MODELS)

@MODELS.register_module()
class NetB(nn.Module):
    def forward(self, x):
        return x + 1
```

We could build either `NetA` or `NetB` by:

```python
from mmcv.cnn import MODELS as MMCV_MODELS
net_a = MMCV_MODELS.build(cfg=dict(type='mmdet.NetA'))
net_b = MMCV_MODELS.build(cfg=dict(type='mmcls.NetB'))
```

# Modification for MM repos:
before:
```python
from mmcv.utils import Registry, build_from_cfg
BACKBONES = Registry('backbone')
def build(cfg, registry, default_args=None):
    """Build a module.

    Args:
        cfg (dict, list[dict]): The config of modules, is is either a dict
            or a list of configs.
        registry (:obj:`Registry`): A registry the module belongs to.
        default_args (dict, optional): Default arguments to build the module.
            Defaults to None.

    Returns:
        nn.Module: A built nn module.
    """
    if isinstance(cfg, list):
        modules = [
            build_from_cfg(cfg_, registry, default_args) for cfg_ in cfg
        ]
        return nn.Sequential(*modules)
    else:
        return build_from_cfg(cfg, registry, default_args)

def build_backbone(cfg):
    """Build backbone."""
    return build(cfg, BACKBONES) 
```
after:
```python
from mmcv.cnn import ModelRegistry
from mmcv.cnn import MODELS as MMCV_MODELS
MODELS = ModelRegistry('models', parent=MMCV_MODELS)
def build_model(cfg):
    """Build model."""
    return MODELS.build(cfg)
```


hellock:
We can briefly describe the design in the PR description to involve more community discussions.

ZwwWayne:
The original goal of design seems to be achieved perfectly. One remaining question: is there a guide to bypass the 3-level hierarchy? Because from the code [here](https://github.com/open-mmlab/mmcv/pull/760/files#diff-4dd793cf824f1b0688049ed7ce700d16a0709e9f4eb1212c46e73371b5c70db0R106), it seems that only a 2-level hierarchy is accepted.

ZwwWayne:
We may also provide a guide for migrating from the old code to the new code (or what do we need to take care of to avoid BC-breaking) in the doc if necessary.

xvjiarui:
> The original goal of design seems to be achieved perfectly. One remaining question: is there a guide to bypass the 3-level hierarchy? Because from the code [here](https://github.com/open-mmlab/mmcv/pull/760/files#diff-4dd793cf824f1b0688049ed7ce700d16a0709e9f4eb1212c46e73371b5c70db0R106), it seems that only a 2-level hierarchy is accepted.

Multi-level is supported.

xvjiarui:
> We may also provide a guide for migrating from the old code to the new code (or what do we need to take care of to avoid BC-breaking) in the doc if necessary.

I will provide a illustration in this PR. And I will create a PR in model for demo.

mmcv/utils/registry.py
```diff
@@ -28,14 +96,43 @@ def __repr__(self):
...
 
+    @staticmethod
+    def infer_scope():
+        filename = inspect.getmodule(inspect.stack()[2][0]).__name__
+        split_filename = filename.split('.')
+        return split_filename[0]
+
+    @staticmethod
+    def split_scope_key(key):
+        # split_list = key.split('.')
```
ZwwWayne, line 107: clean commented code.

ZwwWayne:
We may also need to provide one more registry for DATASET, thus downstream tasks can use similar build function from MMCV to build model, dataset, and runner in the unified entry point/engine.

mmcv/utils/registry.py
```diff
@@ -5,16 +5,84 @@
...
 
-    def __init__(self, name):
+    def __init__(self, name, build_func=None, parent=None, scope=None):
         self._name = name
         self._module_dict = dict()
+        self._children = dict()
+        self._scope = self.infer_scope() if scope is None else scope
+        if parent is not None:
+            assert isinstance(parent, Registry)
+            parent._add_children(self)
```
hellock, line 78: We may do this step at last.

mmcv/utils/registry.py
```diff
@@ -5,16 +5,84 @@
...
+        obj_cls = obj_type
+    else:
+        raise TypeError(
+            f'type must be a str or valid type, but got {type(obj_type)}')
+
+    return obj_cls(**args)
+
+
 class Registry:
     """A registry to map strings to classes.
```
hellock, line 55: Documentation can be here.

mmcv/utils/registry.py
```diff
@@ -28,14 +96,36 @@ def __repr__(self):
                      f'items={self._module_dict})'
         return format_str
 
+    @staticmethod
+    def infer_scope():
```
hellock, line 100: Docstring

mmcv/utils/registry.py
```diff
@@ -45,7 +135,27 @@ def get(self, key):
         Returns:
             class: The corresponding class.
         """
-        return self._module_dict.get(key, None)
+        scope, real_key = self.split_scope_key(key)
+        if scope is not None:
+            return self._children[scope].get(real_key)
```
hellock, line 140: Does this satisfy the desired use case?

mmcv/utils/registry.py
```diff
@@ -45,7 +135,27 @@ def get(self, key):
...
+                # get from children
+                for registry in self._children.values():
+                    result = registry.get(real_key)
+                    if result is not None:
+                        return result
+
+    def build(self, *args, **kwargs):
+        return self.build_func(*args, **kwargs, registry=self)
+
+    def _add_children(self, registry):
```
hellock, line 155: Docstring.

mmcv/utils/registry.py
xvjiarui: Yep. The object could be built with inheritance in a recursive way.

docs/registry.md
```diff
@@ -66,5 +67,43 @@ If the module is successfully registered, you can use this converter through con
...
+class NetB(nn.Module):
+    def forward(self, x):
+        return x + 1
+```
+
+We could build either `NetA` or `NetB` by:
+
+```python
+from mmcv.cnn import MODELS as MMCV_MODELS
+net_a = MMCV_MODELS.build(cfg=dict(type='mmdet.NetA'))
```
ZwwWayne, line 107: There are two questions that might need to be answered  in the doc. 

1. In this case, can we still build net_a and net_b through

```python
net_a = MMCV_MODELS.build(cfg=dict(type='NetA'))
net_b = MMCV_MODELS.build(cfg=dict(type='NetB'))
```
2. If MMCls and MMDet both create a class named `NetA`, what will happen if we try

```python
net_a = MMCV_MODELS.build(cfg=dict(type='NetA'))
```
Does this depend on the order during registration or something else? Or the conflict can only be resolved by use `mmdet.NetA` and `mmcls.NetA` ?

ZwwWayne:
> We may also need to provide one more registry for DATASET, thus downstream tasks can use similar build function from MMCV to build model, dataset, and runner in the unified entry point/engine.

Ping this part. Are we going to add this in this PR or in the future when necessary?

xvjiarui:
> > We may also need to provide one more registry for DATASET, thus downstream tasks can use similar build function from MMCV to build model, dataset, and runner in the unified entry point/engine.
> 
> Ping this part. Are we going to add this in this PR or in the future when necessary?

The dataset build function is not unified yet. I suggest add DATASET in a new PR>

ZwwWayne:
> > > We may also need to provide one more registry for DATASET, thus downstream tasks can use similar build function from MMCV to build model, dataset, and runner in the unified entry point/engine.
> > 
> > 
> > Ping this part. Are we going to add this in this PR or in the future when necessary?
> 
> The dataset build function is not unified yet. I suggest add DATASET in a new PR>

OK.

docs/registry.md
xvjiarui: Currently, it is depend on the order.

ZwwWayne:
LGTM now. See if @hellock has any comments.

ZwwWayne (approved):

xvjiarui:
Kindly ping @hellock.

ZwwWayne:
May merge conflicts first.

mmcv/utils/registry.py
```diff
@@ -28,14 +107,57 @@ def __repr__(self):
...
+            >>> Registry.split_scope_key('mmdet.ResNet')
+            'mmdet', 'ResNet'
+            >>> Registry.split_scope_key('ResNet')
+            None, 'ResNet'
+
+        Return:
+            scope (str, None): The first scope.
+            key (str): The remaining key.
+        """
+        split_index = key.find('.')
```
hellock, line 139: What if there are more than one dot in `key`?

xvjiarui: Handled in a recursive way.

mmcv/utils/registry.py
```diff
@@ -45,7 +179,42 @@ def get(self, key):
         Returns:
             class: The corresponding class.
         """
-        return self._module_dict.get(key, None)
+        scope, real_key = self.split_scope_key(key)
+        if scope is not None:
+            if scope in self._children:
```
hellock, line 184: What if users define two registries in one codebases and make one registry be the parent of the other?

```python
A = Registry('models')
B = Registry('sub_models', parent=A)
```

xvjiarui: The scope of B could be inferred automatically.

mmcv/utils/registry.py
```diff
@@ -5,16 +5,98 @@
...
+            Default: None.
     """
 
-    def __init__(self, name):
+    def __init__(self, name, build_func=None, parent=None, scope=None):
         self._name = name
         self._module_dict = dict()
+        self._children = dict()
+        self._scope = self.infer_scope() if scope is None else scope
+        if build_func is None:
```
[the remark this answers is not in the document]
xvjiarui: Added. It also explained in the docstring.

xvjiarui:
There is a remaining issue, how could build modules in `mmcv.MODELS` with `mmdet.MODELS`? What prefix should be used? 
Currently, if `mmdet.MODELS.build` get `"ResNet"`, it will search `mmdet.MODELS` only. If `mmdet.MODELS.build` get `"mmaction.ResNet3D"` it will search `mmcv.MODELS` from top to bottom.

docs/registry.md
```diff
@@ -71,5 +72,63 @@ If the module is successfully registered, you can use this converter through con
 
 ```python
 converter_cfg = dict(type='Converter1', a=a_value, b=b_value)
-converter = build_converter(converter_cfg)
+converter = CONVERTERS.build(converter_cfg)
+```
+
+## Hierarchy Registry
```
hellock, line 78: This part needs refactoring. Users will get confused by this doc and do not know why and how.

docs/registry.md
```diff
@@ -71,5 +72,63 @@ If the module is successfully registered, you can use this converter through con
...
+@MODELS.register_module()
+class NetB(nn.Module):
+    def forward(self, x):
+        return x + 1
+```
+
+We could build two net in either MMDetection or MMClassification by:
+
+```python
+import mmcls # import mmcls to register models
```
hellock, line 111: Could this be omitted? It is unexpected that users need to modify the source codes apart from setting the configs.

mmcv/cnn/builder.py
```diff
@@ -0,0 +1,29 @@
+import torch.nn as nn
+
+from ..utils import Registry, build_from_cfg
+
+
+def build_model_from_cfg(cfg, registry, default_args=None):
+    """Build a module from config dict(s).
```
hellock, line 7: Explain the difference from `build_from_cfg`.

mmcv/utils/registry.py
```diff
@@ -5,16 +5,106 @@
...
         name (str): Registry name.
+        build_func(func, optional): Build function to construct instance from
+            Registry, func:`build_from_cfg` is used if neither ``parent`` or
+            ``build_func`` is specified. If ``parent`` is specified and
+            ``build_func`` is not given,  ``build_func`` will be inherited
+            from ``parent``. Default: None.
+        parent (Registry, optional): Parent registry. The class registered in
+            children registry could be built from parent. Default: None.
+        scope (str, optional): The scope of registry. It is the key to index
+            children. If not specified, scope will be the name of the package
```
hellock, line 81: What does it mean by "It is the key to index children"?

xvjiarui:
[ ] Refactor registry.md

xvjiarui:
@hellock doc refactor done.

ZwwWayne (approved):

xvjiarui:
@hellock another doc refactor done.

hellock (approved):

PR #775, opened 2021-01-05
[Feature] Allow register multi-name for a module simultaneously

nbei:


mmcv/utils/registry.py
```diff
@@ -54,10 +54,13 @@ def _register_module(self, module_class, module_name=None, force=False):
 
         if module_name is None:
             module_name = module_class.__name__
-        if not force and module_name in self._module_dict:
-            raise KeyError(f'{module_name} is already registered '
-                           f'in {self.name}')
-        self._module_dict[module_name] = module_class
+        if is_str(module_name):
```
ZwwWayne, line 57: We may add an assertion of the type of `module_name`?

ZwwWayne (approved):

mmcv/utils/registry.py
```diff
@@ -54,10 +54,18 @@ def _register_module(self, module_class, module_name=None, force=False):
...
```
hellock, line 57: `isinstance(module_name, str)` is okay. `is_str` will be deprecated since it is mainly used for py2.

hellock:
Unit tests failed.

PR #779, opened 2021-01-07
[Fix]: Set installation of onnx simplify's related packages to optional

RunningLeon:
After `mmcv.onnx.simplify` is added, when call `mmcv.ops`,  `onnx`,  `onnxruntime` and `onnxoptimizer` would be forced to install. This would be unfriendly for users who don't use onnx related feature.
Modification in this PR:
1. Set installation of onnx simplify's related packages to optional.

Thus, for users who want to use onnx related feature, you have to manually install  `onnx`,  `onnxruntime` and `onnxoptimizer`.

### DIFF
--- a/docs/onnx.md
+++ b/docs/onnx.md
@@ -27,7 +27,8 @@ import onnx
 import numpy as np
 
 import mmcv
-from mmcv.onnx import simplify
+from mmcv.onnx.simplify import simplify
+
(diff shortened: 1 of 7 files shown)


ZwwWayne (approved):

ZwwWayne (approved):

ZwwWayne:
Please update PR messages and the documentation to indicate that users need to install onnxoptimizer

ZwwWayne (approved):

PR #784, opened 2021-01-12
Add gradient cumulative optimizer

### LINKED ISSUE #190: Cumulative gradient？Using small BatchSize to simulate big BatchSize
Could you develop the "Cumulative gradient"？
It using small BatchSize to simulate big BatchSize.

eg: [PyTorch中在反向传播前为什么要手动将梯度清零? ](https://www.zhihu.com/question/303070254/answer/573037166)

ZhiyuanChen:
fixes #190

### DIFF
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -35,6 +35,30 @@ class OptimizerHook(Hook):
 
 
 @HOOKS.register_module()
+class GradientCumulativeOptimizerHook(OptimizerHook):
+
+    def __init__(self, grad_clip=None, cumulative_iters=1):
+        super(GradientCumulativeOptimizerHook, self).__init__(grad_clip)
+        self.cumulative_iters = cumulative_iters
+        self.steps = 0
+
+    def after_train_iter(self, runner):
+        self.steps += 1
+        if (runner.iter + 1) % cumulative_iters == 0 or runner.iter == max_iters:
+            runner.optimizer.zero_grad()
+            runner.outputs['loss'] = runner.outputs['loss'] / self.steps
+            runner.outputs['loss'].backward()
+            if self.grad_clip is not None:
+                grad_norm = self.clip_grads(runner.model.parameters())
+                if grad_norm is not None:
+                    # Add grad norm to the logger
+                    runner.log_buffer.update({'grad_norm': float(grad_norm)},
+                                             runner.outputs['num_samples'])
+            runner.optimizer.step()
+            self.steps = 0
+
+
+@HOOKS.register_module()
 class Fp16OptimizerHook(OptimizerHook):
     """FP16 optimizer hook.
 
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -40,11 +40,22 @@ class GradientCumulativeOptimizerHook(OptimizerHook):
     def __init__(self, grad_clip=None, cumulative_iters=1):
         super(GradientCumulativeOptimizerHook, self).__init__(grad_clip)
         self.cumulative_iters = cumulative_iters
+        self.divisible_ietrs = 0
+        self.remainder_iters = 0
+        self.initialized = False
+
+    def _init(self, runner):
+        self.divisible_ietrs = runner.max_iters // self.cumulative_iters * self.cumulative_iters
+        self.remainder_iters = runner.max_iters % self.cumulative_iters
+        self.initialized = True
 
     def after_train_iter(self, runner):
-        runner.outputs['loss'] = runner.outputs['loss'] / self.cumulative_iters
+        if not self.initialized:
+            self._init(runner)
+        loss_factor = self.cumulative_iters if runner.iter < self.divisible_ietrs else self.remainder_iters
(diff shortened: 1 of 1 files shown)


ZhiyuanChen:
Unit tests to be added

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,6 +34,27 @@ def after_train_iter(self, runner):
...
 
+@HOOKS.register_module()
+class GradientCumulativeOptimizerHook(OptimizerHook):
+
+    def __init__(self, grad_clip=None, cumulative_iters=1):
+        super(GradientCumulativeOptimizerHook, self).__init__(grad_clip)
+        self.cumulative_iters = cumulative_iters
+
+    def after_train_iter(self, runner):
+        runner.outputs['loss'] = runner.outputs['loss'] / self.cumulative_iters
```
ZhiyuanChen, line 45: Gradients will be scaled improperly with the current approach if the `runner.max_iters` is not divisible by `cumulative_iters`

nbei:
Please fix the linting error

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,6 +34,27 @@ def after_train_iter(self, runner):
         runner.optimizer.step()
 
 
+@HOOKS.register_module()
+class GradientCumulativeOptimizerHook(OptimizerHook):
+
```
nbei, line 39: Docs are missing.

mmcv/runner/hooks/optimizer.py
nbei, line 45: I'm confusing about the detailed implementation of this function. Could you please offer a reference for this implementation to show that it is a **general case**? 

In my opinion, the accumulative gradients are adopted to avoid large batch sizes. Is it right? If it's right, why should we divide the `self.cumulative_iters`?

nbei (commented):
1. Unit tests are missing.
2. The detailed implementation should be clarified. Please see my comments.

mmcv/runner/hooks/optimizer.py
ZhiyuanChen: Gradient accumulate is to achieve an equivalent larger batch size with small batch size. Therefore, the loss should be normalised. See more at https://discuss.pytorch.org/t/pytorch-gradient-accumulation/55955/2

nbei: Ok, got it. You may specify this in the docs.

In addition, please also fix the corner case where `total_iters % cumulative_iters != 0`

ggalan87: Hello, isn't BatchNormalization an issue as mentioned in here https://gist.github.com/thomwolf/ac7a7da6b1888c2eeac8ac8b9b05d3d3#gistcomment-3381285

ZhiyuanChen: @ggalan87 Yes the behavior of batch normal is different, however, not all network contain batchnorm

ZhiyuanChen:
As mentioned by @ggalan87 
We probably need to raise a warning when `runner.model` contains `nn.BatchNorm`.

nbei:
Hi @ZhiyuanChen , many thanks for your contribution. Please fix the linting error first. It seems that you have not adopted the pre-commit hook as requested in [CONTRIBUTING](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md).

mmcv/runner/hooks/optimizer.py
```diff
@@ -40,11 +40,22 @@ class GradientCumulativeOptimizerHook(OptimizerHook):
     def __init__(self, grad_clip=None, cumulative_iters=1):
         super(GradientCumulativeOptimizerHook, self).__init__(grad_clip)
         self.cumulative_iters = cumulative_iters
+        self.divisible_ietrs = 0
```
nbei, line 43: `self.divisible_iters`

mmcv/runner/hooks/optimizer.py
```diff
@@ -40,11 +40,22 @@ class GradientCumulativeOptimizerHook(OptimizerHook):
...
+        self.remainder_iters = 0
+        self.initialized = False
+
+    def _init(self, runner):
+        self.divisible_ietrs = runner.max_iters // self.cumulative_iters * self.cumulative_iters
```
nbei, line 48: There is another corner case where users resume from `iter=2` but the `cumulative_iters=4`. It seems this implementation will bring wrong gradients. If no good solutions, please put a warning here.

mmcv/runner/hooks/optimizer.py
```diff
@@ -40,11 +40,22 @@ class GradientCumulativeOptimizerHook(OptimizerHook):
     def __init__(self, grad_clip=None, cumulative_iters=1):
         super(GradientCumulativeOptimizerHook, self).__init__(grad_clip)
         self.cumulative_iters = cumulative_iters
+        self.divisible_ietrs = 0
+        self.remainder_iters = 0
+        self.initialized = False
+
+    def _init(self, runner):
```
nbei, line 47: Please put a warning for the usage of BN.

nbei (commented):
See comments.

zhouzaida:
closed by #1221

PR #786, opened 2021-01-13
[Feature] : Add ScatterND TensorRT Plugin

grimoire:
This PR add support to ONNX [ScatterND](https://github.com/onnx/onnx/blob/master/docs/Operators.md#ScatterND)
Support convert of pytorch code such as:
```python
tensor[:, -2:] += 1
```

### DIFF
--- a/mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
+++ b/mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
@@ -49,8 +49,6 @@ nvinfer1::DimsExprs ONNXScatterNDDynamic::getOutputDimensions(
 bool ONNXScatterNDDynamic::supportsFormatCombination(
     int pos, const nvinfer1::PluginTensorDesc *inOut, int nbInputs,
     int nbOutputs) {
-  // const auto *in = inOut;
-  // const auto *out = inOut + nbInputs;
   if (pos < nbInputs) {
     switch (pos) {
       case 0:
--- a/mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
+++ b/mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
@@ -1,0 +1,207 @@
+#include "trt_scatternd.hpp"
+
+#include <assert.h>
+
+#include <chrono>
+
+#include "trt_serialize.hpp"
+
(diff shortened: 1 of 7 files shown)


mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
```diff
@@ -0,0 +1,206 @@
...
+      case 1:
+        // indices
+        return inOut[pos].type == nvinfer1::DataType::kINT32 &&
+               inOut[pos].format == nvinfer1::TensorFormat::kLINEAR;
+      case 2:
+        // updates
+        return inOut[pos].type == inOut[0].type &&
+               inOut[pos].format == inOut[0].format;
+      default:
+        return true;
```
RunningLeon, line 69: Will there have any problem if `return true` is set by default?

mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
```diff
@@ -0,0 +1,206 @@
+#include "trt_scatternd.hpp"
```
RunningLeon, line 1: Would add unit test for  `scatternd plugin`.

mmcv/ops/csrc/tensorrt/plugins/trt_scatternd.cpp
grimoire: Theoretically, switch() include all possible case. Default branch will never be reached.

mmcv/ops/csrc/tensorrt/plugins/trt_scatternd_kernel.cu
```diff
@@ -0,0 +1,93 @@
...
+using mmcv::TensorDesc;
+
+template <typename T>
+__global__ void onnx_scatternd_kernel(const int n, const int* indices,
+                                      const T* update, T* output,
+                                      TensorDesc tensor_desc,
+                                      TensorDesc indice_desc) {
+  const int indice_cols = indice_desc.shape[indice_desc.dim - 1];
+  const int copy_stride = tensor_desc.stride[indice_cols - 1];
+  const int* shape = &(tensor_desc.shape[0]);
```
RunningLeon, line 20: `shape` is unused. Could be removed.

RunningLeon (approved):
LGTM

hellock (approved):

PR #787, opened 2021-01-13
[Feature] : Add NonMaxSuppression TensorRT Plugin 

grimoire:
This PR add support of convert [NonMaxSuppression](https://github.com/onnx/onnx/blob/master/docs/Operators.md#NonMaxSuppression)
Note: this is based on https://github.com/open-mmlab/mmcv/pull/786

Because TensorRT does not support dynamic graph, the output of nms layer will be filled with `-1` if the box is invalid. This might cause some unexpected result.


mmcv/ops/csrc/tensorrt/plugins/trt_nms.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+const char *ONNXNonMaxSuppressionDynamic::getPluginNamespace() const {
+  return mNamespace.c_str();
+}
+
+////////////////////// creator /////////////////////////////
+
+ONNXNonMaxSuppressionDynamicCreator::ONNXNonMaxSuppressionDynamicCreator() {
+  mPluginAttributes.clear();
+  // mPluginAttributes.emplace_back(nvinfer1::PluginField("center_point_box"));
```
RunningLeon, line 199: Could this line be removed if it is not a comment?

mmcv/ops/csrc/tensorrt/plugins/trt_nms.cpp
```diff
@@ -0,0 +1,252 @@
...
+    switch (pos - nbInputs) {
+      case 0:
+        // selected_indices
+        return inOut[pos].type == nvinfer1::DataType::kINT32 &&
+               inOut[pos].format == nvinfer1::TensorFormat::kLINEAR;
+      default:
+        return true;
+    }
+  }
+  return true;
```
RunningLeon, line 105: We may consider if `return true` is OK.

mmcv/ops/csrc/tensorrt/plugins/trt_nms_kernel.cu
```diff
@@ -0,0 +1,317 @@
...
+  int* index_template = (int*)workspace;
+  workspace = static_cast<char*>(workspace) + spatial_dimension * sizeof(int);
+  thrust::sequence(thrust::cuda::par.on(stream), index_template,
+                   index_template + spatial_dimension, 0);
+
+  int* index_cache = (int*)workspace;
+  workspace = static_cast<char*>(workspace) +
+              spatial_dimension * num_batches * num_classes * sizeof(int);
+
+  // this is dirty, should we create another nms_cuda kernel with float*
```
RunningLeon, line 158: We may add TODO here and optimize it with another PR.

mmcv/ops/csrc/tensorrt/plugins/trt_nms.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+void ONNXNonMaxSuppressionDynamic::configurePlugin(
+    const nvinfer1::DynamicPluginTensorDesc *inputs, int nbInputs,
+    const nvinfer1::DynamicPluginTensorDesc *outputs, int nbOutputs) {
+  mNumberInputs = nbInputs;
+}
+
+size_t ONNXNonMaxSuppressionDynamic::getWorkspaceSize(
```
RunningLeon, line 114: We should set max number of input boxes to nms plugin.

mmcv/ops/csrc/tensorrt/plugins/trt_nms.cpp
grimoire: There is an optional attribute [center_point_box](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#attributes-138) in onnx nms.
TensorRT can not parse the attribute if it is not explicit in the model. with error log:
```
IndexError: Attribute not found: center_point_box
```
That's why I comment this line to disable this attribute. Should I add a TODO here?

RunningLeon: Is there a way to solve this? say, set a default attribute?

grimoire: Yes, I have already add the default attribute to the node when preprocess onnx.

grimoire:
Remove code in commit https://github.com/open-mmlab/mmcv/pull/787/commits/9e32c3d0caa7924227fcf422cbefabc8734c2b05 might not be a good idea.
Nms might return less inds than max_output_size. Say, if we have 100 boxes and only 10 boxes have score > threshold. Top 10 inds will be valid, all other inds will be filled with -1. Indexing boxes and scores with -1 will give us the last row boxes and scores, which might not be valid.

RunningLeon:
> Remove code in commit [9e32c3d](https://github.com/open-mmlab/mmcv/commit/9e32c3d0caa7924227fcf422cbefabc8734c2b05) might not be a good idea.
> Nms might return less inds than max_output_size. Say, if we have 100 boxes and only 10 boxes have score > threshold. Top 10 inds will be valid, all other inds will be filled with -1. Indexing boxes and scores with -1 will give us the last row boxes and scores, which might not be valid.

Because` inds` and `dets` are both returned from `NMS`,  adding dummy data to boxes and scores inside `NMS` can only guarantee `dets` are correct. This is an issue we have to solve.

RunningLeon (approved):
LGTM

mmcv/ops/nms.py
```diff
@@ -285,10 +287,12 @@ def batched_nms(boxes, scores, idxs, nms_cfg, class_agnostic=False):
...
 
     split_thr = nms_cfg_.pop('split_thr', 10000)
-    if boxes_for_nms.shape[0] < split_thr:
+    # Won't split to multiple nms nodes when exporting to onnx
+    if boxes_for_nms.shape[0] < split_thr or torch.onnx.is_in_onnx_export():
         dets, keep = nms_op(boxes_for_nms, scores, **nms_cfg_)
         boxes = boxes[keep]
-        scores = dets[:, -1]
+        # Keep tracing when export to ONNX.
+        scores = scores[keep]
```
ZwwWayne, line 295: This will cause bug when using soft nms as soft nms will modify the scores and we should use `det[:, -1]`.

RunningLeon: `-1` indexing works abnormally in TensorRT, changed to `4`

RunningLeon (approved):

mmcv/ops/csrc/tensorrt/plugins/trt_nms_kernel.cu
```diff
@@ -0,0 +1,271 @@
...
+   *             center_point_box == 1.
+   * scores_sorted: [spatial_dimension], the workspace used to sort scores.
+   * dev_mask: [spatial_dimension, col_blocks], intersect mask,
+   *           where col_block = DIVUP(spatial_dimension, threadsPerBlock)
+   * index_template: [spatial_dimension], sequence (0,1,2,3,...), workspace
+   *                 for argsort.
+   * index_cache: [num_batches, spatial_dimension], argsort of
+   *              each classes.
+   *
+   */
```
hellock, line 163: We can use the docstring style compatible with doxygen.

mmcv/tensorrt/tensorrt_utils.py
```diff
@@ -1,7 +1,91 @@
...
+
+    nrof_node = len(nodes)
+    for idx in range(nrof_node):
+        node = nodes[idx]
+        node_attributes = node.attribute
+        node_inputs = node.input
+        node_outputs = node.output
+        node_name = node.name
+        # process NonMaxSuppression node
+        if node.op_type == 'NonMaxSuppression':
```
hellock, line 47: Any better design than hardcode?

mmcv/tensorrt/tensorrt_utils.py
```diff
@@ -1,7 +1,91 @@
+import numpy as np
+import onnx
 import tensorrt as trt
 import torch
 
 
+def preprocess_onnx(onnx_model):
+    """Modify onnx model to match with TensorRT plugins in mmcv.
```
hellock, line 8: Add a section of detailed description of what was modified in this method.

mmcv/tensorrt/tensorrt_utils.py
grimoire: We can use some `Registry` and `build` mechanism to map the onnx node name and process code, just like mmdetection.
But I am not sure if it worth to do. Most official onnx node does not need to do any process. NonMaxSuppression is a rare case.

mmcv/ops/nms.py
```diff
@@ -295,7 +298,8 @@ def batched_nms(boxes, scores, idxs, nms_cfg, class_agnostic=False):
     if boxes_for_nms.shape[0] < split_thr or torch.onnx.is_in_onnx_export():
         dets, keep = nms_op(boxes_for_nms, scores, **nms_cfg_)
         boxes = boxes[keep]
-        scores = dets[:, -1]
+        # -1 indexing works abnormal in TensorRT
```
ZwwWayne, line 301: Add comments here:
```python
This assumes `dets` has 5 dimensions where the last dimension is score.
TODO: more elegant way to handle the dimension issue.

```

ZwwWayne:
LGTM if the comments can be resolved. See if @hellock  have any more comments.

mmcv/tensorrt/tensorrt_utils.py
RunningLeon: I think we can just leave it this way. Currently,  we only have one special node to process. Besides, `preprocess_onnx` could be removed when the interface between ONNXRuntime and TensorRT is aligned.

ZwwWayne:
This PR will be merged after the V1.2.7 release.

hellock:
LGTM

PR #795, opened 2021-01-18
[Bug] fix for norm attribute in ConvModule

cvluca:
The attribute self.norm_name only exists when self.with_norm is True.
Calling getattr will throw error if self.norm_name is not a attribute,

### DIFF
--- a/mmcv/cnn/bricks/conv_module.py
+++ b/mmcv/cnn/bricks/conv_module.py
@@ -162,7 +162,7 @@ class ConvModule(nn.Module):
 
     @property
     def norm(self):
-        return getattr(self, self.norm_name)
+        return getattr(self, self.norm_name) if self.with_norm else None
 
     def init_weights(self):
         # 1. It is mainly for customized conv layers with their own


nbei:
Could you please detail the cases where you have to depend on this modification? In the `forward` function, only if `self.with_norm` is `True`, we will try to use the property of `norm`. Thus, it seems that `AttributeError` will not be raised as long as you call the `ConvModule` correctly.

cvluca:
Some libraries (I found this error in pytorch) use `dir()` to get a list of valid attributes, `norm` will also appear in that list.
Maybe I should fix it by adding `hasattr` to test if the attribute is in the list.

OpenMMLab-Coodinator:
Hi @cvluca  ！We are grateful for your efforts in helping improve this open-source project during your personal time.
Welcome to join OpenMMLab Special Interest Group (SIG) , where you can share your experiences, ideas, and build connections with like-minded peers.

To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA

PR #803, opened 2021-01-25
add ort nms op

tangyanf:
add onnxruntime custom op: nms

### DIFF
--- a/mmcv/ops/nms.py
+++ b/mmcv/ops/nms.py
@@ -25,7 +25,7 @@ class NMSop(torch.autograd.Function):
         has_custom_op = is_custom_op_loaded()
         if has_custom_op:
             return g.op(
-                'mmcv::NonMaxSupression',
+                'mmcv::NonMaxSuppression',
                 bboxes,
                 scores,
                 iou_threshold_f=float(iou_threshold),
--- a/mmcv/ops/nms.py
+++ b/mmcv/ops/nms.py
@@ -21,22 +21,32 @@ class NMSop(torch.autograd.Function):
 
     @staticmethod
     def symbolic(g, bboxes, scores, iou_threshold, offset):
-        from torch.onnx.symbolic_opset9 import select, squeeze, unsqueeze
-        boxes = unsqueeze(g, bboxes, 0)
-        scores = unsqueeze(g, unsqueeze(g, scores, 0), 0)
-        max_output_per_class = g.op(
-            'Constant', value_t=torch.tensor([sys.maxsize], dtype=torch.long))
-        iou_threshold = g.op(
-            'Constant',
-            value_t=torch.tensor([iou_threshold], dtype=torch.float))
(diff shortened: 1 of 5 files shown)


hellock:
@grimoire Would you help review this PR?

grimoire:
@hellock Sure, I will do it.

grimoire:
Hi @tangyanf, thanks for your contribution!
As long as onnxruntime already has an official nms implementation, I think we might not have to create custom one.
The diff between pytorch and onnxruntime is caused by `offset`, use the code below could fix this:
```python
        if torch.onnx.is_in_onnx_export() and offset == 1:
            # ONNX only support offset == 1
            boxes[:, -2:] += 1
        inds = NMSop.apply(boxes, scores, iou_threshold, offset)
        if torch.onnx.is_in_onnx_export() and offset == 1:
            # ONNX only support offset == 1
            boxes[:, -2:] -= 1
```

tangyanf:
> Hi @tangyanf, thanks for your contribution!
> As long as onnxruntime already has an official nms implementation, I think we might not have to create custom one.
> The diff between pytorch and onnxruntime is caused by `offset`, use the code below could fix this:
> 
> ```python
>         if torch.onnx.is_in_onnx_export() and offset == 1:
>             # ONNX only support offset == 1
>             boxes[:, -2:] += 1
>         inds = NMSop.apply(boxes, scores, iou_threshold, offset)
>         if torch.onnx.is_in_onnx_export() and offset == 1:
>             # ONNX only support offset == 1
>             boxes[:, -2:] -= 1
> ```

mmcv/ops/nms.py
```diff
@@ -108,6 +120,9 @@ def nms(boxes, scores, iou_threshold, offset=0):
     """
     assert isinstance(boxes, (torch.Tensor, np.ndarray))
     assert isinstance(scores, (torch.Tensor, np.ndarray))
+    print('save input data')
+    np.save('boxes.bin', boxes.detach().cpu().numpy())
+    np.save('scores.bin', scores.detach().cpu().numpy())
```
grimoire, line 125: Please remove these code.

tests/test_ops/test_onnx.py
```diff
@@ -46,12 +46,19 @@ def test_nms():
             opset_version=11)
     onnx_model = onnx.load(onnx_file)
 
+    ort_custom_op_path = get_onnxruntime_op_path()
+    if not os.path.exists(ort_custom_op_path):
+        pytest.skip('softnms for onnxruntime is not compiled.')
```
grimoire, line 51: Please change the log from softnms to nms.

mmcv/ops/nms.py
tangyanf: Sorry, I forgot to delete

tests/test_ops/test_onnx.py
tangyanf: ok

hellock (approved):

PR #790, opened 2021-01-15
Refactor _load_checkpoint fn

hhaAndroid:
Refactor _load_checkpoint fn

PR #804, opened 2021-01-25
Add map_location in load_url_dist

innerlee:
Fix https://github.com/open-mmlab/mmpose/issues/201


hellock:
This PR conflicts with the refactoring #790. It would be better to merge #790 first and then this one.

hellock:
@innerlee This PR can be updated accordingly.

innerlee:
Cool

PR #805, opened 2021-01-26
Hande exif orientation with `pillow` backend in `imread`

daavoo:
The default behavior of `imread` with the `pillow` backend ignores the exif orientation tag, resulting in inconsistencies with the default backend (`cv2`) which properly handles it. This P.R. adds a regression test and updates `_pillow2array` to match the behavior of the `cv2` backend.


hellock (approved):

hellock:
Thanks for the nice fix!

PR #808, opened 2021-01-29
[Feature] Add file mode for logger

nbei:
Default implementation: https://docs.python.org/3/library/logging.handlers.html

### DIFF
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -5,7 +5,7 @@ import torch.distributed as dist
 logger_initialized = {}
 
 
-def get_logger(name, log_file=None, log_level=logging.INFO):
+def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
     """Initialize and get a logger by name.
 
     If the logger has not been initialized, this method will initialize the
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

mmcv/utils/logging.py
```diff
@@ -45,7 +45,7 @@ def get_logger(name, log_file=None, log_level=logging.INFO):
 
     # only rank 0 will add a FileHandler
     if rank == 0 and log_file is not None:
-        file_handler = logging.FileHandler(log_file, 'w')
```
ZwwWayne, line 48: May add comments to indicate the reason of modification

mmcv/utils/logging.py
```diff
@@ -5,7 +5,7 @@
 logger_initialized = {}
 
 
-def get_logger(name, log_file=None, log_level=logging.INFO):
+def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
```
ZwwWayne, line 8: Please also update the docstring accordingly.

ZwwWayne (changes requested):

ZwwWayne (approved):

Issue #792, opened 2021-01-16
MMDataParallel Implementation inconsistency between CPU / GPU 


24hours:
It seems that [MMDataParallel](https://github.com/open-mmlab/mmcv/blob/master/mmcv/parallel/data_parallel.py) support CPU only mode. 

However, https://github.com/open-mmlab/mmcv/blob/master/mmcv/parallel/data_parallel.py#L52 will prompt error if `self.device_ids` is falsy. 
```python
 return self.module.train_step(*inputs, **kwargs)
```

In GPU mode, the following code is executed instead https://github.com/open-mmlab/mmcv/blob/master/mmcv/parallel/data_parallel.py#L67 with no error message. 
```python
return self.module.train_step(*inputs[0], **kwargs[0])
```

modifying line 52 to `(*inputs[0], *kwargs[0])` seems to resolve  the error with no other issue. Is there a reason why there is difference between line 52 and line 67 ? 

In addition, 

why CPU mode unsqueeze additional dimension ? https://github.com/open-mmlab/mmcv/blob/master/mmcv/parallel/_functions.py#L28
`output = output.unsqueeze(0)`

The comment mention
```python
# unsquzee the first dimension thus the tensor's shape is the
# same as those scattered with GPU.
```

But it is not clear how GPU will have addition dimension

24hours:
I have found another inconsistency between GPU and GPU
```python
from mmcv.parallel import scatter_kwargs, DataContainer
import torch 

inputs = (torch.zeros([20, 3, 128, 128]), )
output, _ = scatter_kwargs(inputs, {}, [-1], 0)
print('CPU', output[0][0].size())
output, _ = scatter_kwargs(inputs, {}, [0], 0)
print('GPU', output[0][0].size())

print('------------')
inputs = (DataContainer([torch.zeros([20, 3, 128, 128])]),)
output, _ = scatter_kwargs(inputs, {}, [-1], 0)
print('CPU', output[0][0].size())
output, _ = scatter_kwargs(inputs, {}, [0], 0)
print('GPU', output[0][0].size())
```
```
CPU torch.Size([20, 3, 128, 128])
GPU torch.Size([20, 3, 128, 128])
---------------------
CPU torch.Size([1, 20, 3, 128, 128])
GPU torch.Size([20, 3, 128, 128])
```

you can see that CPU return wrong size when compare to GPU implementation

ycxioooong:
[...]

PR #814, opened 2021-02-03
add cpu check on DataContainer scattering

24hours:
fix issue related to open-mmlab/mmcv#792


hellock:
There may also be inconsistency between different PyTorch versions, @ZwwWayne can have a check.

24hours:
Im closing this with the reason below. 

https://github.com/open-mmlab/mmcv/issues/792

PR #823, opened 2021-02-06
fix: imshow_bboxes fails with incontiguous img

### LINKED ISSUE #801: imshow_bboxes does not show bboxes if img is incontiguous
As [DKDKDDK](https://stackoverflow.com/questions/57586449/why-cv2-rectangle-sometimes-return-np-ndarray-while-sometimes-cv2-umat) asked, cv2.rectangle seems unable to draw inplacely on incontiguous arrays. When calling `mmcv.imshow_bboxes` or `mmcv.imshow_det_bboxes`, the contiguousness of argument `img` is consistent with what the user passed in. Would it be convenient to add `img = np.ascontiguousarray(img)` inside `mmcv.imshow_bboxes` and `mmcv.imshow_det_bboxes`?

LutingWang:
Fix #801

### DIFF
--- a/mmcv/visualization/image.py
+++ b/mmcv/visualization/image.py
@@ -53,6 +53,7 @@ def imshow_bboxes(img,
         ndarray: The image with bboxes drawn on it.
     """
     img = imread(img)
+    img = np.ascontiguousarray(img)
 
     if isinstance(bboxes, np.ndarray):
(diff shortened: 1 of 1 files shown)


ycxioooong (approved):
LGTM. @hellock pls have a look.

hellock (approved):

PR #824, opened 2021-02-07
Fix init

MeowZheng:
add kwargs instead of same arguments of `BaseInit` and fix docstring and error msg


ycxioooong:
Seems like the commit history is a little bit messy.

PR #825, opened 2021-02-07
fix weight_init.py

MeowZheng:
use kwargs instead of the same arguments in `BaseInit`, and fix docstring and error msg.

### DIFF
--- a/mmcv/cnn/utils/weight_init.py
+++ b/mmcv/cnn/utils/weight_init.py
@@ -77,7 +77,7 @@ def bias_init_with_prob(prior_prob):
 
 class BaseInit(object):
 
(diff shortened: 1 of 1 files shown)


mmcv/cnn/utils/weight_init.py
```diff
@@ -77,7 +77,7 @@ def bias_init_with_prob(prior_prob):
 
 class BaseInit(object):
 
-    def __init__(self, bias, bias_prob, layer):
+    def __init__(self, bias=0, bias_prob=None, layer=None):
```
hellock, line 80: We can force keyword arguments by
```python
def __init__(self, *, bias=0, bias_prob=None, layer=None):
```

hellock (approved):

PR #826, opened 2021-02-08
map_location for all

innerlee:
replacement of https://github.com/open-mmlab/mmcv/pull/804

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -276,15 +276,16 @@ def load_from_http(filename, map_location=None, model_dir=None):
     Returns:
         dict or OrderedDict: The loaded checkpoint.
     """
-
-    rank = int(os.environ.get('LOCAL_RANK', get_dist_info()[0]))
+    rank, world_size = get_dist_info()
+    rank = int(os.environ.get('LOCAL_RANK', rank))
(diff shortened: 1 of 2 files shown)


mmcv/runner/checkpoint.py
```diff
@@ -277,14 +277,12 @@ def load_from_http(filename, map_location=None, model_dir=None):
...
     """
 
-    rank, world_size = get_dist_info()
-    rank = int(os.environ.get('LOCAL_RANK', rank))
+    rank = int(os.environ.get('LOCAL_RANK', get_dist_info()[0]))
     if rank == 0:
-        checkpoint = model_zoo.load_url(filename, model_dir=model_dir)
-    if world_size > 1:
+        checkpoint = model_zoo.load_url(filename, model_dir=model_dir, map_location=map_location)
+    else:
```
innerlee, line 283: Please double check this change

hellock: For the rank-0 worker, `torch.distributed.barrier()` is also needed in distributed training to wait other processes.

hellock: L272 may also be updated since the docstring is wrong.

hellock (approved):

PR #828, opened 2021-02-09
Initial implementation and design for OpenMMLab model hub

ZwwWayne:
This PR adds some basic functions for the OpenMMLab model hub, which will be used in all the downstream codebases.

### DIFF
--- a/mmcv/model_zoo/hub.py
+++ b/mmcv/model_zoo/hub.py
@@ -7,7 +7,21 @@ from mmcv import dump, load
 
 def update_model_zoo(name: str, config: str, details: dict,
                      checkpoint_path: str, json_path: str) -> None:
-    """Update the model zoo information in the json file.
+    r"""Update the model zoo information in the json file.
+
+    The format of json file storing the model information should be as below:
+
+    .. code-block:: json
+
+    [
+        {
+            'name': str,
+            'config': str,
+            'details': str,
+            'path': str
+        }
+        ...
+    ]
 
     Args:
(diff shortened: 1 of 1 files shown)


mmcv/model_zoo/hub.py
```diff
@@ -0,0 +1,69 @@
...
+
+    The format of json file storing the model information should be as below:
+
+    .. code-block:: json
+
+    [
+        {
+            'name': str,
+            'config': str,
+            'details': str,
```
jshilong, line 20: 'details' is a dict

mmcv/model_zoo/hub.py
```diff
@@ -0,0 +1,117 @@
+import subprocess
+from typing import Union
+
+import torch
+
+from mmcv import dump, load
+from mmcv.utils import Config
+
+
+def update_model_zoo(name: str, config: str, details: dict,
```
jshilong, line 10: Maybe we have to add information about the name of the repo in JSON, we need this information to get the corresponding builder.

mmcv/model_zoo/hub.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns:
+        tuple: (model, details)
+    """
+
+    # TODO: support to load the json for only once
+    #
+
+
+def load_model(
+        repo_or_dir: str,
```
jshilong, line 95: Maybe task name + model name would be a better choice, for example when we want to build a model from a self_develop codebase,   we may not be able to infer which builder should be used.

PR #829, opened 2021-02-10
[Feature] Support revise_keys in load_checkpoint().

yaochaorui:
This would be useful when one would like to incorporate other models into open-mmlab series, e.g. when using simple encoder-decoder structure in mmediting, the keyword backbone may need to be added for the checkpoint. Then one can set the addword='backbone'.

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -45,6 +45,8 @@ class BaseRunner(metaclass=ABCMeta):
             Defaults to None.
         max_epochs (int, optional): Total training epochs.
         max_iters (int, optional): Total training iterations.
+        revise_keys (list): Customized keywords to modify for compatibility
+            between model and checkpoint. Default: remove 'module.'.
     """
 
     def __init__(self,
@@ -55,7 +57,8 @@ class BaseRunner(metaclass=ABCMeta):
                  logger=None,
                  meta=None,
                  max_iters=None,
-                 max_epochs=None):
+                 max_epochs=None,
+                 revise_keys=[(r'^module.', '')]):
         if batch_processor is not None:
             if not callable(batch_processor):
                 raise TypeError('batch_processor must be callable, '
@@ -102,7 +105,7 @@ class BaseRunner(metaclass=ABCMeta):
         self.optimizer = optimizer
         self.logger = logger
         self.meta = meta
-
+        self.revise_keys = revise_keys
         # create work_dir
         if mmcv.is_str(work_dir):
             self.work_dir = osp.abspath(work_dir)
@@ -307,9 +310,10 @@ class BaseRunner(metaclass=ABCMeta):
             getattr(hook, fn_name)(self)
 
     def load_checkpoint(self, filename, map_location='cpu', strict=False):
+
         self.logger.info('load checkpoint from %s', filename)
         return load_checkpoint(self.model, filename, map_location, strict,
-                               self.logger)
+                               self.logger, self.revise_keys)
 
     def resume(self,
                checkpoint,
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -45,7 +45,6 @@ class BaseRunner(metaclass=ABCMeta):
             Defaults to None.
         max_epochs (int, optional): Total training epochs.
         max_iters (int, optional): Total training iterations.
-
     """
 
     def __init__(self,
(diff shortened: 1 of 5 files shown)


mmcv/runner/checkpoint.py
```diff
@@ -503,6 +504,8 @@ def load_checkpoint(model,
                     filename,
                     map_location=None,
                     strict=False,
+                    rmword='module',
```
innerlee, line 507: Change this to
```python
revise_key=(r'^module.', '')
```

`revise_key` accepts both a tuple of two regexes, or a list of tuples.

innerlee: emm maybe a dict is also acceptable.

```python
revise_key={r'^module.': ''}
```

yaochaorui: Hi, thanks for your suggestion! Modified. BTW, it seems that this one can cover the _load_checkpoint_with_prefix() function in the sense of functionality.  While the latter may be more user-friendly.

innerlee:
Tests are failing. You can click on the details link to see what casued the error

innerlee:
ping @yaochaorui

mmcv/runner/checkpoint.py
```diff
@@ -514,6 +516,8 @@ def load_checkpoint(model,
         map_location (str): Same as :func:`torch.load`.
         strict (bool): Whether to allow different params for the model and
             checkpoint.
+        revise_keys (list): Customized keywords to modify for compatibility
+            between model and checkpoint. Default: remove 'module.'.
```
innerlee, line 520: Default: strip the prefix `module.` by `[(r'^module.', '')])`.

mmcv/runner/base_runner.py
```diff
@@ -55,7 +57,8 @@ def __init__(self,
                  logger=None,
                  meta=None,
                  max_iters=None,
-                 max_epochs=None):
+                 max_epochs=None,
+                 revise_keys=[(r'^module.', '')]):
```
innerlee, line 61: No need to change the signature of runner

yaochaorui: Hi, modified. This function can be called with the help of the 'load_from' in config.

tests/test_runner/test_checkpoint.py
```diff
@@ -188,6 +189,30 @@ def __init__(self):
...
+
+    # add prefix
+    torch.save(model.state_dict(), chkpt_path)
+    load_checkpoint(pmodel, chkpt_path, revise_keys=[(r'^', 'backbone.')])
+
+    # strip prefix
+    torch.save(pmodel.state_dict(), chkpt_path)
+    load_checkpoint(model, chkpt_path, revise_keys=[('backbone.', '')])
+
+    os.remove(chkpt_path)
```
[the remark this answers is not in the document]
yaochaorui: Hi, thanks for your comments! Will modify them.

mmcv/runner/base_runner.py
yaochaorui: Correction: add revise_keys=[] in config to call this func.

mmcv/runner/checkpoint.py
```diff
@@ -503,7 +504,8 @@ def load_checkpoint(model,
                     filename,
                     map_location=None,
                     strict=False,
-                    logger=None):
+                    logger=None,
+                    revise_keys=[(r'^module.', '')]):
```
innerlee, line 508: @MeowZheng Do you prefer a dict `{r'^module.': ''}` over the current list of tuple?

mmcv/runner/checkpoint.py
```diff
@@ -530,8 +535,8 @@ def load_checkpoint(model,
     else:
         state_dict = checkpoint
     # strip prefix of state_dict
-    if list(state_dict.keys())[0].startswith('module.'):
-        state_dict = {k[7:]: v for k, v in state_dict.items()}
+    for (p, s) in revise_keys:
```
innerlee, line 538: Unnecessary brackets for `p, s`

mmcv/runner/checkpoint.py
MeowZheng: Some previous version python doesn't preserve the order for `dict`, and it may be not safe to revise state_dict, but key-value stores are more understandable. In this way, I prefer `OrderedDict`. :)

innerlee: `OrderedDict` is too long...

Let's keep the list

mmcv/runner/checkpoint.py
yaochaorui: Ok.

MeowZheng (approved):

mmcv/runner/checkpoint.py
```diff
@@ -515,6 +517,9 @@ def load_checkpoint(model,
         strict (bool): Whether to allow different params for the model and
             checkpoint.
         logger (:mod:`logging.Logger` or None): The logger for error message.
+        revise_keys (list): Customized keywords to modify for compatibility
```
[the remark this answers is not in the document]
yaochaorui: Polished. Hope this version is more clear.

hellock (approved):

PR #844, opened 2021-02-20
Fix the iter error when the number of GPUs is different during resume

hhaAndroid:
reslove: https://github.com/open-mmlab/mmdetection/issues/3676

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -337,9 +337,12 @@ class BaseRunner(metaclass=ABCMeta):
             config = mmcv.Config.fromstring(
                 checkpoint['meta']['config'], file_format='.py')
             previous_gpu_ids = config.get('gpu_ids', None)
-            if previous_gpu_ids and len(previous_gpu_ids) > 0:
+            if previous_gpu_ids and len(previous_gpu_ids) > 0 and len(
+                    previous_gpu_ids) != self.world_size:
                 self._iter = int(self._iter * len(previous_gpu_ids) /
                                  self.world_size)
+                self.logger.info('the iteration number is changed due to '
+                                 'change of GPU number')
 
         if 'optimizer' in checkpoint and resume_optimizer:
             if isinstance(self.optimizer, Optimizer):
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -2,7 +2,6 @@
 import copy
 import logging
 import os.path as osp
-import tempfile
 import warnings
 from abc import ABCMeta, abstractmethod
 
@@ -332,17 +331,13 @@ class BaseRunner(metaclass=ABCMeta):
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
 
+        # Recalculate iter when the number of GPUs is different during resume
         if 'config' in checkpoint['meta']:
-            temp_file = tempfile.NamedTemporaryFile()
-            config_path = f'{temp_file.name}.py'
-            with open(config_path, 'w') as f:
-                f.write(checkpoint['meta']['config'])
-            config = mmcv.Config.fromfile(config_path)
+            config = mmcv.Config.fromstring(checkpoint['meta']['config'])
             previous_gpu_ids = config.get('gpu_ids', None)
             if previous_gpu_ids and len(previous_gpu_ids) > 0:
                 self._iter = int(self._iter * len(previous_gpu_ids) /
                                  self.world_size)
-            temp_file.close()
 
         if 'optimizer' in checkpoint and resume_optimizer:
             if isinstance(self.optimizer, Optimizer):
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -2,6 +2,7 @@
 import copy
 import logging
 import os.path as osp
+import tempfile
 import warnings
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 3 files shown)


mmcv/runner/base_runner.py
```diff
@@ -330,6 +331,19 @@ def resume(self,
 
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
+
+        if 'config' in checkpoint['meta']:
```
hellock, line 335: Add a comment for the purpose of this code block.

mmcv/runner/base_runner.py
```diff
@@ -330,6 +331,19 @@ def resume(self,
...
+            temp_file = tempfile.NamedTemporaryFile()
+            config_path = f'{temp_file.name}.py'
```
hellock, line 337: Seems that you just want to get a temporary directory name instead of creating a NamedTemporaryFile.

mmcv/runner/base_runner.py
```diff
@@ -330,6 +331,19 @@ def resume(self,
...
+            with open(config_path, 'w') as f:
+                f.write(checkpoint['meta']['config'])
+            config = mmcv.Config.fromfile(config_path)
```
hellock, line 340: Add a staticmethod `Config.fromstring` can simplify the above logic.

mmcv/utils/config.py
```diff
@@ -253,6 +253,32 @@ def fromfile(filename,
             import_modules_from_strings(**cfg_dict['custom_imports'])
         return Config(cfg_dict, cfg_text=cfg_text, filename=filename)
 
+    @staticmethod
+    def fromstring(cfg_str, suffix=None, is_pretty_text=True):
```
hellock, line 257: The argument `is_pretty_text` is unnecessary. Setting `suffix='py'` or `is_pretty_text=True` are of the same cost for users.

hhaAndroid: If remove `is_pretty_text`, when the user input `suffix='.json'` and cfg_str not from pretty_text, the program will go wrong.

mmcv/runner/base_runner.py
```diff
@@ -330,6 +330,16 @@ def resume(self,
...
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
+
+        # Recalculate iter when the number of GPUs is different during resume
+        if 'config' in checkpoint['meta']:
+            config = mmcv.Config.fromstring(
+                checkpoint['meta']['config'], file_format='.py')
+            previous_gpu_ids = config.get('gpu_ids', None)
+            if previous_gpu_ids and len(previous_gpu_ids) > 0:
+                self._iter = int(self._iter * len(previous_gpu_ids) /
```
ZwwWayne, line 340: Log the information when change self._iter to indicate users that the iteration number is changed due to change of GPU number

mmcv/utils/config.py
```diff
@@ -253,6 +253,31 @@ def fromfile(filename,
...
+        if file_format not in ['.py', '.json', '.yaml', '.yml']:
+            raise IOError('Only py/yml/yaml/json type are supported now!')
+        with tempfile.NamedTemporaryFile('w', suffix=file_format) as temp_file:
+            temp_file.write(cfg_str)
+            temp_file.flush()
+            try:
+                cfg = Config.fromfile(temp_file.name)
+            except Exception as e:
+                raise SyntaxError(
+                    f'{e}\nPlease check file_format, If cfg_str comes from '
```
ZwwWayne, line 277: "cfg_str" "file_format"

tests/test_utils/test_config.py
```diff
@@ -161,6 +161,29 @@ def test_fromfile():
...
+        out_cfg = Config.fromstring(cfg_str, file_format)
+        assert in_cfg._cfg_dict == out_cfg._cfg_dict
+
+    cfg_file = osp.join(data_path, 'config', 'b.json')
+    in_cfg = Config.fromfile(cfg_file)
+    with pytest.raises(SyntaxError):
+        Config.fromstring(in_cfg.pretty_text, '.json')
+
+    cfg_str = open(cfg_file, 'r').read()
+    with pytest.raises(SyntaxError):
```
ZwwWayne, line 183: Add comments on what error you are testing.

mmcv/utils/config.py
```diff
@@ -253,6 +253,32 @@ def fromfile(filename,
...
+            obj:`Config`: Config obj.
+        """
+        if file_format not in ['.py', '.json', '.yaml', '.yml']:
+            raise IOError('Only py/yml/yaml/json type are supported now!')
+        with tempfile.NamedTemporaryFile('w', suffix=file_format) as temp_file:
+            temp_file.write(cfg_str)
+            temp_file.flush()
+            try:
+                cfg = Config.fromfile(temp_file.name)
+            except Exception as e:
```
hellock, line 275: This exception handling is too wild. Other errors will also be catched and a misleading message will be printed.
You may use some trivial rules to check if this string represents python source codes. It may work for 90% cases though not perfect.
```python
if suffix != '.py':
    # check if users specify a wrong suffix for python
    if 'dict(' in cfg_str:
        warnings.warn('xxxx')
```

hhaAndroid: I did want to trivial the exception rules but considering that the error type is difficult to define, soI did not do it.

hellock (approved):

PR #853, opened 2021-02-24
[Feature] Support save last checkpoint

xvjiarui:


mmcv/runner/hooks/checkpoint.py
```diff
@@ -35,6 +36,7 @@ def __init__(self,
                  out_dir=None,
                  max_keep_ckpts=-1,
                  sync_buffer=False,
+                 save_last=True,
```
hellock, line 39: Better to put `save_last` together with `max_keep_ckpts`.

mmcv/runner/hooks/checkpoint.py
```diff
@@ -26,6 +26,7 @@ class CheckpointHook(Hook):
             Default: -1, which means unlimited.
         sync_buffer (bool): Whether to synchronize buffers in different
             gpus. Default: False.
+        keep_last (bool): Whether to save last checkpoint.
```
hellock, line 29: save_last (bool): Whether to force the last checkpoint to be saved regardless of interval.

mmcv/runner/hooks/checkpoint.py
```diff
@@ -99,3 +102,10 @@ def after_train_iter(self, runner):
         if self.sync_buffer:
             allreduce_params(runner.model.buffers())
         self._save_checkpoint(runner)
+
+    def after_run(self, runner):
+        if self.save_last:
+            if not self.out_dir:
+                self.out_dir = runner.work_dir
```
hellock, line 109: Maybe this two lines can be moved to the init method?

xvjiarui: But the checkpoint hook can't access runner during `__init__`. Moved to `before_run` instead.

hellock (approved):

hellock:
@xvjiarui Unit tests failed. There might be bugs.

xvjiarui:
@hellock Note that I set the `save_last` default to be true, so the original behavior is changed.

hellock:
> @hellock Note that I set the `save_last` default to be true, so the original behavior is changed.

This should be okay.

PR #858, opened 2021-02-28
[Feature] : Add Deformable Conv2d TensorRT Plugin

grimoire:
This PR add deformable conv2d support. Note that mDCN and deformable pooling is not include.


mmcv/ops/csrc/tensorrt/trt_cuda_helper.cuh
```diff
@@ -13,4 +13,8 @@
     }                                                          \
   }
 
+template <class scalar_t>
+void memcpyPermute(scalar_t *dst, const scalar_t *src, int *src_size,
```
RunningLeon, line 17: Could we add a short docstring for common functions in trt_cuda_helper?

RunningLeon (approved):
Tested and LGTM

ZwwWayne (approved):

PR #864, opened 2021-03-03
[Feature] Add adjust sharpness

LXXXXR:
Hi, 
In this pull request, `adjust_sharpness` is added for data augmentation. Pls kindly take a look. Thx.

### DIFF
--- a/mmcv/image/photometric.py
+++ b/mmcv/image/photometric.py
@@ -250,6 +250,12 @@ def adjust_sharpness(img, factor=1., kernel=None):
         kernel (np.ndarray, optional): Filter kernel to be applied on the img
             to obtain the degenerated img. Defaults to None.
 
+    Notes:
+        No value sanity check is enforced on the kernel set by users. So with
+        an inappropriate kernel, the `adjust_sharpness` may fail to perform
+        the function its name indicates but end up performing whatever
+        transform determined by the kernel.
+
     Returns:
         ndarray: The sharpened image.
     """
--- a/mmcv/image/photometric.py
+++ b/mmcv/image/photometric.py
@@ -97,9 +97,10 @@ def posterize(img, bits):
 
 
 def adjust_color(img, alpha=1, beta=None, gamma=0):
-    """It blends the source image and its gray image:
+    r"""It blends the source image and its gray image:
 
-    ``output = img * alpha + gray_img * beta + gamma``
+    .. math::
+        output = img * alpha + gray\_img * beta + gamma
 
     Args:
         img (ndarray): The input source image.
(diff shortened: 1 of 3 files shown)


mmcv/image/photometric.py
```diff
@@ -227,6 +227,33 @@ def adjust_contrast(img, factor=1.):
...
+    ``output = img * factor + degenerated * (1 - factor)``
+
+    Args:
+        img (ndarray): Image to be sharpened. BGR order.
+        factor (float): Same as :func:`mmcv.adjust_brightness`.
+
+    Returns:
+        ndarray: The sharpened image.
+    """
+    kernel = np.array([[1., 1., 1.], [1., 5., 1.], [1., 1., 1.]]) / 13
```
MeowZheng, line 248: Could you please give an explanation about kernel value?

MeowZheng (approved):

mmcv/image/photometric.py
```diff
@@ -227,6 +227,34 @@ def adjust_contrast(img, factor=1.):
...
+def adjust_sharpness(img, factor=1.):
+    """Adjust image sharpness.
+
+    This function controls the sharpness of an image. An
+    enhancement factor of 0.0 gives a blurred image. A
+    factor of 1.0 gives the original image. And a factor
+    of 2.0 gives a sharpened image. It blends the source
+    image and the degenerated mean image:
+
+    ``output = img * factor + degenerated * (1 - factor)``
```
ZwwWayne, line 239: May use code block or math block for this sentence.

mmcv/image/photometric.py
```diff
@@ -227,6 +227,34 @@ def adjust_contrast(img, factor=1.):
...
+
+    Args:
+        img (ndarray): Image to be sharpened. BGR order.
+        factor (float): Same as :func:`mmcv.adjust_brightness`.
+
+    Returns:
+        ndarray: The sharpened image.
+    """
+    # adopted from PIL.ImageFilter.SMOOTH
+    kernel = np.array([[1., 1., 1.], [1., 5., 1.], [1., 1., 1.]]) / 13
```
ZwwWayne, line 249: Should we make the kernel a settable  argument and by default using this value?

LXXXXR: Revised accordingly. However, one thing to note is that there is no way to check if the kernel users set can actually perform the function of adjusting sharpness, so the function `adjust_sharpness` will not always performs the function its name indicates but performs whatever transform determined by the kernel.

ZwwWayne: This can be put into the Notification section in the docstring as a reminder.

ZwwWayne (approved):
LGTM

PR #868, opened 2021-03-04
add tifffile backend support for image io

sshuair:
The input image data is usually 16bit and multi-channels(>4 channel) tiffile in the field of remote sensing. The default `cv2`, `pillow` image io backend can not handle this file.

So, this PR intended to support this feature.


tests/test_image/test_io.py
```diff
@@ -28,6 +28,7 @@ def setup_class(cls):
         cls.palette_img_path = osp.join(cls.data_dir, 'palette.gif')
         cls.exif_img_path = osp.join(cls.data_dir, 'color_exif.jpg')
         cls.img = cv2.imread(cls.img_path)
+        cls.tiff_path = osp.join(cls.data_dir, 'uint16-5channel.tif')
```
hellock, line 31: The file is 1.15MB large. Is it possible to use a smaller image?

sshuair: ok, I’ll replace it with smaller one.

sshuair: It's now 208Kb.

sshuair:
@hellock the new test tiff file size is 208kb with 150px x 200px. Is it ok? It's difficult to reduce the file size even smaller.

hellock (approved):

hellock:
> @hellock the new test tiff file size is 208kb with 150px x 200px. Is it ok? It's difficult to reduce the file size even smaller.

It is okay now. Thanks!

PR #882, opened 2021-03-09
fix scatter in pytorch18

jshilong:
This PR adds a method named `to_kwargs`  to `MMDistributedDataParallel`.

`DistributedDataParallel` in PyTorch 1.8 use `to_kwargs` instead of `scatter` to move all tensor in dataflow to 
 the device. it is necessary to overload this function as `scatter`.

### breaking change
None

Fixes open-mmlab/mmdetection#4737

### DIFF
--- a/mmcv/parallel/distributed.py
+++ b/mmcv/parallel/distributed.py
@@ -18,6 +18,11 @@ class MMDistributedDataParallel(DistributedDataParallel):
     - It implement two APIs ``train_step()`` and ``val_step()``.
     """
 
+    def to_kwargs(self, inputs, kwargs, device_id):
+        #  Use `self.to_kwargs` instead of `self.scatter` in pytorch1.8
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

mmcv/parallel/distributed.py
```diff
@@ -18,6 +18,11 @@ class MMDistributedDataParallel(DistributedDataParallel):
     - It implement two APIs ``train_step()`` and ``val_step()``.
     """
 
+    def to_kwargs(self, inputs, kwargs, device_id):
+        #  Use `self.to_kwargs` instead of `self.scatter` in pytorch1.8
```
yeliudev, line 22: Are there too many blanks between `#` and `Use`?

jshilong: sure

PR #884, opened 2021-03-12
[Refactoring]Add Sequential with init_weight

MeowZheng:
As developers always used `nn.Sequential` to implement a submodule, I add `Sequential` with `init_weight` for parameters initialization with `init_cfg`.  In addition, I override `__repr__` to display initialization information.

example
```python
seq_model_cfg = [
        dict(
            type='FooConv1d', init_cfg=dict(type='Constant', val=0., bias=1.)),
        dict(
            type='FooConv2d', init_cfg=dict(type='Constant', val=2., bias=3.)),
    ]
layers = [build_from_cfg(cfg, COMPONENTS) for cfg in seq_model_cfg]
seq_model = Sequential(*layers)
seq_model.init_weight()


ZwwWayne:
Should we also handle `nn.ModuleList`?

hellock (approved):

PR #889, opened 2021-03-15
fix bug of convmodule

### LINKED ISSUE #871: 'ConvModule' object has no attribute 'norm' when using torch.jit.trace
environment: python3.6.6 pytorch1.7.0

code:

```
import torch
from mmcv.cnn.bricks import ConvModule

conv = ConvModule(3, 8, 2, norm_cfg=dict(type='BN'))
input_example = torch.randn((1, 3, 224, 224))

ts = torch.jit.trace(func=conv.eval(), example_inputs=input_example)
torch.jit.save(ts, 'conv_module.ts')
```

It work well, but when set `norm_cfg=None`, failed.

```
torch.nn.modules.module.ModuleAttributeError: 'ConvModule' object has no attribute 'norm'
```

Any help?

jshilong:
This PR fixes the potential bug of `ConvModule` when specifying the `norm_cfg` as None.
The original implementation would raise the `ConvModule' object has no attribute 'norm` when you try to call
`self.norm` when passing `norm_cfg` as None.

I believe the original design is not reasonable, because you can find an attribute named `norm` by `dir(instance)` or `instance.__dict__`,  but when you access it by `instance.norm`, it will raise an error. 

Closes #871

### DIFF
--- a/mmcv/cnn/bricks/conv_module.py
+++ b/mmcv/cnn/bricks/conv_module.py
@@ -146,6 +146,8 @@ class ConvModule(nn.Module):
                 norm_channels = in_channels
             self.norm_name, norm = build_norm_layer(norm_cfg, norm_channels)
             self.add_module(self.norm_name, norm)
+        else:
+            self.norm_name = None
 
         # build activation layer
         if self.with_activation:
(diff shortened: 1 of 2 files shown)


mmcv/cnn/bricks/conv_module.py
```diff
@@ -145,7 +145,11 @@ def __init__(self,
             else:
                 norm_channels = in_channels
             self.norm_name, norm = build_norm_layer(norm_cfg, norm_channels)
+            assert self.norm_name != 'norm', 'norm has been used as' \
```
nbei, line 148: add unit test for this assertion or just remove this assertion?

nbei (commented):
LGTM

nbei (approved):

PR #893, opened 2021-03-17
[Refactoring] Revise function of layers and override keys in init_cfg

MeowZheng:
Hi,

As developers would like to only initialize parameters with attribute names, I just revise the initialization mechanism. If we don't define `layer` key or `override` key, it will not initialize anything; if we define `override` but don't define `layer`, it will initialize parameters with the attribute name in `override`; if we only define `layer`, it just initialize the layer in `layer` key; if we define `override` and `layer`, `override` has higher priority and will override initialization mechanism.

Moreover, for ease of use, if we only define `override`, and `override` without `type` key, the `type` key and other kwargs will merge in `override`.

### DIFF
--- a/mmcv/cnn/utils/weight_init.py
+++ b/mmcv/cnn/utils/weight_init.py
@@ -344,6 +344,9 @@ class PretrainedInit(object):
 
 def _initialize(module, cfg, wholemodule=False):
     func = build_from_cfg(cfg, INITIALIZERS)
+    # wholemodule flag is for override mode, there is no layer key in override
+    # and initializer will give init values for the whole module with the name
+    # in override.
     func.wholemodule = wholemodule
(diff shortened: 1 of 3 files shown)


zhouzaida (approved):

mmcv/cnn/utils/weight_init.py
```diff
@@ -78,6 +80,7 @@ def bias_init_with_prob(prior_prob):
 class BaseInit(object):
 
     def __init__(self, *, bias=0, bias_prob=None, layer=None):
+        self.overmodule = False
```
ZwwWayne, line 83: The name of overmodule is a little bit weird. May not get the meaning at the first glance.

ZwwWayne (approved):

PR #899, opened 2021-03-18
add more detail of warning in basemodule

jshilong:
Add the class name of the module to the warning in ``init_weight`` of BaseModule.

### DIFF
--- a/mmcv/runner/base_module.py
+++ b/mmcv/runner/base_module.py
@@ -47,8 +47,8 @@ class BaseModule(nn.Module, metaclass=ABCMeta):
                     module.init_weight()
             self._is_init = True
         else:
-            warnings.warn('This module has bee initialized, \
-                please call initialize(module, init_cfg) to reinitialize it')
+            warnings.warn(f'init_weight of {self.__class__.__name__} has '
+                          f'been called more than once.')
 
     def __repr__(self):
         s = super().__repr__()


mmcv/runner/base_module.py
```diff
@@ -47,8 +47,8 @@ def init_weight(self):
                     module.init_weight()
             self._is_init = True
         else:
-            warnings.warn('This module has bee initialized, \
-                please call initialize(module, init_cfg) to reinitialize it')
+            warnings.warn(f'init_weight of {self.__class__.__name__} has '
+                          f'been called more than once.')
```
zhouzaida, line 51: Agree, but "please call initialize(module, init_cfg) to reinitialize it"  should be kept to give a tip for users what they could do.

jshilong: Actually, I am confused about why we should reinitialize it?

MeowZheng: It should not be initialized again, so I leave a warning as a reminder.

MeowZheng (approved):

mmcv/runner/base_module.py
jshilong: So  what dose ```please call initialize(module, init_cfg) to reinitialize it``` mean?

ZwwWayne (approved):

PR #913, opened 2021-03-29
[Fix] Fix wandb logger drop result bug

shenmishajing:
You can find details at #911.


zhouzaida:
Advice:
fix wandb logger drop result bug by delete step param
->
[Fix] Fix wandb logger drop result bug

zhouzaida:
> You can find details at #911.

Although it works, the step don't equals self._iter anymore.

image

xvjiarui:
Hi @shenmishajing 
Will `commit` argumentation do the trick?

zhouzaida:
> Hi @shenmishajing
> Will `commit` argumentation do the trick?

"You can also set commit=False in wandb.log to accumulate metrics, just be sure to call wandb.log without the commit flag to persist the metrics."

zhouzaida:
[For example, if you have training and validation steps you'd like to align, pass us your own step counter: wandb.log({"acc":1, "global_step":1}). Then in the graphs choose "global_step" as the x-axis.](https://docs.wandb.ai/library/log#custom-x-axis)

```python
    @master_only
    def log(self, runner):
        # if tags:
        #     self.wandb.log(
        #         tags, step=self.get_iter(runner), commit=self.commit)
        if tags:
            tags['global_step'] =self.get_iter(runner)
            self.wandb.log(
                tags, commit=self.commit)
```

shenmishajing:
> > You can find details at #911.
> 
> Although it works, the step don't equals self._iter anymore.
> 
> image

Does it matter? I didn't pay attention to the meaning of step var when I use wandb.
And the step var minus self._iter is always equal to number of val epoch we have passed. So, I think we can ignore this.

> > Hi @shenmishajing
> > Will `commit` argumentation do the trick?
> 
> "You can also set commit=False in wandb.log to accumulate metrics, just be sure to call wandb.log without the commit flag to persist the metrics."

yep, if we use the `commit` argumentation to do the trick. We may need do some change like follows:
In mmcv/runner/hooks/logger/wandb.py
log func from:
```python
    @master_only
    def log(self, runner):
        if tags:
            self.wandb.log(
                tags, step=self.get_iter(runner), commit=self.commit)
```
change to
```python
    @master_only
    def log(self, runner, commit=True):
        if tags:
            self.wandb.log(
                tags, step=self.get_iter(runner) + 0 if commit else 1, commit=commit)
```
And we have to rewrite after_val_epoch func like follows:
```python
    def after_val_epoch(self, runner):
        runner.log_buffer.average()
        self.log(runner, commit=False)
        if self.reset_flag:
            runner.log_buffer.clear_output()
```

zhouzaida:
@shenmishajing 
```python
    @master_only
    def log(self, runner):
        # if tags:
        #     self.wandb.log(
        #         tags, step=self.get_iter(runner), commit=self.commit)
        if tags:
            tags['global_step'] =self.get_iter(runner)
            self.wandb.log(
                tags, commit=self.commit)
```
We can choose "global_step" as the x-axis in the graphs or do nothing.

shenmishajing:
> [For example, if you have training and validation steps you'd like to align, pass us your own step counter: wandb.log({"acc":1, "global_step":1}). Then in the graphs choose "global_step" as the x-axis.](https://docs.wandb.ai/library/log#custom-x-axis)
> 
> ```python
>     @master_only
>     def log(self, runner):
>         tags = self.get_loggable_tags(runner)
>         # if tags:
>         #     self.wandb.log(
>         #         tags, step=self.get_iter(runner), commit=self.commit)
>         if tags:
>             tags['global_step'] =self.get_iter(runner)
>             self.wandb.log(
>                 tags, commit=self.commit)
> ```

I approve this.

zhouzaida (approved):

xvjiarui:
If user doesn't use 'val' pipeline, this PR will change the **original behavior**.

shenmishajing:
> If user doesn't use 'val' pipeline, this PR will change the **original behavior**.

If user doesn't use 'val' pipeline, wandb logger will log once every train iter, so wandb step will always equal to self._iter var. By the way, wandb step is equal to self._iter +1 now.
So, I think we can ignore this.
Or, if you persist in this. The solution I mentioned above will not change the **original behavior**.

> yep, if we use the `commit` argumentation to do the trick. We may need do some change like follows:
> In mmcv/runner/hooks/logger/wandb.py
> log func from:
> 
> ```python
>     @master_only
>     def log(self, runner):
>         tags = self.get_loggable_tags(runner)
>         if tags:
>             self.wandb.log(
>                 tags, step=self.get_iter(runner), commit=self.commit)
> ```
> 
> change to
> 
> ```python
>     @master_only
>     def log(self, runner, commit=True):
>         tags = self.get_loggable_tags(runner)
>         if tags:
>             self.wandb.log(
>                 tags, step=self.get_iter(runner) + 0 if commit else 1, commit=commit)
> ```
> 
> And we have to rewrite after_val_epoch func like follows:
> 
> ```python
>     def after_val_epoch(self, runner):
>         runner.log_buffer.average()
>         self.log(runner, commit=False)
>         if self.reset_flag:
>             runner.log_buffer.clear_output()
> ```
> 
> Well, I think it is a little bit ugly, how about you? We can use it if you gays can accept the magic `+ 0 if commit else 1`

xvjiarui:
> > If user doesn't use 'val' pipeline, this PR will change the **original behavior**.
> 
> If user doesn't use 'val' pipeline, wandb logger will log once every train iter, so wandb step will always equal to self._iter var. By the way, wandb step is equal to self._iter +1 now.
> So, I think we can ignore this.
> Or, if you persist in this. The solution I mentioned above will not change the **original behavior**.
> 
> > yep, if we use the `commit` argumentation to do the trick. We may need do some change like follows:
> > In mmcv/runner/hooks/logger/wandb.py
> > log func from:
> > ```python
> >     @master_only
> >     def log(self, runner):
> >         tags = self.get_loggable_tags(runner)
> >         if tags:
> >             self.wandb.log(
> >                 tags, step=self.get_iter(runner), commit=self.commit)
> > ```
> > 
> > 
> > change to
> > ```python
> >     @master_only
> >     def log(self, runner, commit=True):
> >         tags = self.get_loggable_tags(runner)
> >         if tags:
> >             self.wandb.log(
> >                 tags, step=self.get_iter(runner) + 0 if commit else 1, commit=commit)
> > ```
> > 
> > 
> > And we have to rewrite after_val_epoch func like follows:
> > ```python
> >     def after_val_epoch(self, runner):
> >         runner.log_buffer.average()
> >         self.log(runner, commit=False)
> >         if self.reset_flag:
> >             runner.log_buffer.clear_output()
> > ```
> > 
> > 
> > Well, I think it is a little bit ugly, how about you? We can use it if you gays can accept the magic `+ 0 if commit else 1`

Hi @shenmishajing 
What will happen if we just set `commit=True` and change nothing else?

shenmishajing:
> > > If user doesn't use 'val' pipeline, this PR will change the **original behavior**.
> > 
> > 
> > If user doesn't use 'val' pipeline, wandb logger will log once every train iter, so wandb step will always equal to self._iter var. By the way, wandb step is equal to self._iter +1 now.
> > So, I think we can ignore this.
> > Or, if you persist in this. The solution I mentioned above will not change the **original behavior**.
> > > yep, if we use the `commit` argumentation to do the trick. We may need do some change like follows:
> > > In mmcv/runner/hooks/logger/wandb.py
> > > log func from:
> > > ```python
> > >     @master_only
> > >     def log(self, runner):
> > >         tags = self.get_loggable_tags(runner)
> > >         if tags:
> > >             self.wandb.log(
> > >                 tags, step=self.get_iter(runner), commit=self.commit)
> > > ```
> > > 
> > > 
> > > change to
> > > ```python
> > >     @master_only
> > >     def log(self, runner, commit=True):
> > >         tags = self.get_loggable_tags(runner)
> > >         if tags:
> > >             self.wandb.log(
> > >                 tags, step=self.get_iter(runner) + 0 if commit else 1, commit=commit)
> > > ```
> > > 
> > > 
> > > And we have to rewrite after_val_epoch func like follows:
> > > ```python
> > >     def after_val_epoch(self, runner):
> > >         runner.log_buffer.average()
> > >         self.log(runner, commit=False)
> > >         if self.reset_flag:
> > >             runner.log_buffer.clear_output()
> > > ```
> > > 
> > > 
> > > Well, I think it is a little bit ugly, how about you? We can use it if you gays can accept the magic `+ 0 if commit else 1`
> 
> Hi @shenmishajing
> What will happen if we just set `commit=True` and change nothing else?

Do you mean we change the log func like follows?
```python
    @master_only
    def log(self, runner):
        if tags:
            self.wandb.log(
                tags, step=self.get_iter(runner), commit=True)
```
In fact, self.commit is True by default. In other word, this do not change anything.

ZwwWayne:
Lint and CI failed. Could @shenmishajing help fix that?

shenmishajing:
> Lint and CI failed. Could @shenmishajing help fix that?

e, I have no idea, in fact. Anyone want to help me? Or where can I find some doc about this?

ZwwWayne:
> > Lint and CI failed. Could @shenmishajing help fix that?
> 
> I have no idea, in fact. Anyone want to help me? Or where can I find some doc about this?

1. You can see why it fails [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135274). And you can use tools like yapf/flake8 and pre-commit to fix the lint issues.
2. You can see why the unit tests fail [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135516).

shenmishajing:
> > > Lint and CI failed. Could @shenmishajing help fix that?
> > 
> > 
> > I have no idea, in fact. Anyone want to help me? Or where can I find some doc about this?
> 
> 1. You can see why it fails [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135274). And you can use tools like yapf/flake8 and pre-commit to fix the lint issues.
> 2. You can see why the unit tests fail [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135516).

I can fix the unit tests bug.
In tests/test_runner/test_hooks.py, the log res we expected defined as
```python
    hook.wandb.log.assert_called_with({
        'learning_rate': 0.02,
        'momentum': 0.95
    },
                                      step=6,
                                      commit=True)
```
But, we have changed the format of log, so I think we have to change it to
```python
    hook.wandb.log.assert_called_with({
        'learning_rate': 0.02,
        'momentum': 0.95,
        'global_step': 6
    },
                                      commit=True)
```
As the lint issus, I have no idea at all. It's ok when I run `python -m flake8 mmcv` locally. I need some help :).

shenmishajing:
> > > > Lint and CI failed. Could @shenmishajing help fix that?
> > > 
> > > 
> > > I have no idea, in fact. Anyone want to help me? Or where can I find some doc about this?
> > 
> > 
> > 
> > 1. You can see why it fails [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135274). And you can use tools like yapf/flake8 and pre-commit to fix the lint issues.
> > 2. You can see why the unit tests fail [here](https://github.com/open-mmlab/mmcv/pull/913/checks?check_run_id=2224135516).
> 
> I can fix the unit tests bug.
> In tests/test_runner/test_hooks.py, the log res we expected defined as
> 
> ```python
>     hook.wandb.log.assert_called_with({
>         'learning_rate': 0.02,
>         'momentum': 0.95
>     },
>                                       step=6,
>                                       commit=True)
> ```
> 
> But, we have changed the format of log, so I think we have to change it to
> 
> ```python
>     hook.wandb.log.assert_called_with({
>         'learning_rate': 0.02,
>         'momentum': 0.95,
>         'global_step': 6
>     },
>                                       commit=True)
> ```
> 
> As the lint issus, I have no idea at all. It's ok when I run `python -m flake8 mmcv` locally. I need some help :).

Well, I may fix it :).

ZwwWayne:
Hi @shenmishajing ，
Thanks for the cooperation. Do we have some visualization results to see how it looks before and after this PR?

shenmishajing:
> Hi @shenmishajing ，
> Thanks for the cooperation. Do we have some visualization results to see how it looks before and after this PR?

Do you mean the visualization of the log result on wandb site, like this?

> > You can find details at #911.
> 
> Although it works, the step don't equals self._iter anymore.
> 
> image

After this PR, it will look like above. Before this PR, it also looks like above. The only difference is the step var on wandb site is not equal to self._iter in mmcv runner code any more. After this PR, a var named global_step will be added, which will equal to self._iter.

ZwwWayne:
LGTM. This PR will be merged after approval by @xvjiarui

xvjiarui:
The `step` argument is useful under some cases. 
For example, when we training by epochs. The wandb logger log training info after each iteration. Then user terminate the job at the [500/1000]  of Epoch[5] and resume from the previous checkpoint Epoch[4].
Before this PR, the new logs from [1/1000] to [500/1000] will be ignored. Starting from [501/1000], logger will log new train infos. This could keep the log consistent and of the same total step (the x axis of the plot). 
After this PR,  the new logs from [1/1000] to [500/1000] will still be logged after the old  [1/1000] to [500/1000], which is not ideal.

So I suggest make this change optional. Such that the original behavior could also be preserved. 
We could keep the `global_step` as it does not affect other behaviors. We may add an argument `with_step` default to `True`. If `with_step == True`, we will log the `step` from `get_iters`. Otherwise, we will not log `step`.

shenmishajing:
> The `step` argument is useful under some cases.
> For example, when we training by epochs. The wandb logger log training info after each iteration. Then user terminate the job at the [500/1000] of Epoch[5] and resume from the previous checkpoint Epoch[4].
> Before this PR, the new logs from [1/1000] to [500/1000] will be ignored. Starting from [501/1000], logger will log new train infos. This could keep the log consistent and of the same total step (the x axis of the plot).
> After this PR, the new logs from [1/1000] to [500/1000] will still be logged after the old [1/1000] to [500/1000], which is not ideal.
> 
> So I suggest make this change optional. Such that the original behavior could also be preserved.
> We could keep the `global_step` as it does not affect other behaviors. We may add an argument `with_step` default to `True`. If `with_step == True`, we will log the `step` from `get_iters`. Otherwise, we will not log `step`.

approve

xvjiarui (approved):

shenmishajing:
When will I get the pre-build version mmcv in pip with this pr?

PR #918, opened 2021-03-31
[Fix]: Fix initialization bug of override cfg

ZwwWayne:
This PR fixes the bug in the `initialize` function, which will change the original init_cfg after override some keys.
This behavior will cause an error when one config is used in many other modules as the config will be changed after it has been used once.
One test case that will cause the failure is [here](https://github.com/hhaAndroid/mmdetection/blob/refactor_init/tests/test_models/test_backbones/test_resnet.py#L622).


MeowZheng (approved):
Many thanks for fixing this bug.

PR #927, opened 2021-04-07
Remove onnx-simplify source code(deprecated)

QingChuanWS:
1. The purpose of this PR: The [onnx-simplify](https://github.com/daquexian/onnx-simplifier) 3rd party package already supports importing custom operator.
2. For the repo that is still using the onnx simplify function in `mmcv.onnx`,  firstly you need to install the onnx-simplify 3rd party package with the version greater than 0.3.0,  then remove the `mmcv.onnx.simplify` and implement the onnx simplify according to the `simplify()` in onnxsim.
3. This PR is associated with [mmdetection #4913](https://github.com/open-mmlab/mmdetection/pull/4913),  and `mmcv.onnx.simplify` is used in mmdetection to complete the simplify of the onnx model.


tests/test_ops/test_onnx.py
```diff
@@ -345,32 +345,6 @@ def warpped_function(torch_input, torch_rois):
         assert np.allclose(pytorch_output, onnx_output, atol=1e-3)
```
RunningLeon, line 347: update the onnx [doc](https://github.com/grimoire/mmcv/blob/ops_document/docs/onnx.md) as well.

RunningLeon:
> This PR focus on remove the custom simplify model in mmcv.

Please update the description of this PR:
1. clarify why we have to remove this feature
2. what impact it would have

RunningLeon (approved):
LGTM

PR #929, opened 2021-04-08
[Fix] OneCycleLrUpdaterHook adds total_steps

MeowZheng:
Add total_steps, the total number of steps in the cycle. Note that if a value is not provided here, it will be the max_iter of runner.


mmcv/runner/hooks/lr_updater.py
```diff
@@ -469,6 +478,15 @@ def __init__(self,
         super(OneCycleLrUpdaterHook, self).__init__(**kwargs)
 
     def before_run(self, runner):
+        if hasattr(self, 'total_steps'):
+            total_steps = self.total_steps
+        else:
+            total_steps = runner.max_iters
+        if runner.max_iters > total_steps:
```
zhouzaida, line 485: It may be better if replacing `runner.max_iters > total_steps` with`if total_steps < runner.max_iters`

mmcv/runner/hooks/lr_updater.py
```diff
@@ -433,6 +436,7 @@ class OneCycleLrUpdaterHook(LrUpdaterHook):
 
     def __init__(self,
                  max_lr,
+                 total_steps=None,
```
ZwwWayne, line 439: Suggest adding unit tests to test that.

tests/test_runner/test_hooks.py
```diff
@@ -317,6 +317,55 @@ def test_one_cycle_runner_hook():
     hook.writer.add_scalars.assert_has_calls(calls, any_order=True)
 
 
+def test_onecycle_lr_updater_hook():
```
zhouzaida, line 320: Maybe  `def test_onecycle_lr_updater_hook` can be merged into `def test_one_cycle_runner_hook()`

tests/test_runner/test_hooks.py
```diff
@@ -316,6 +316,52 @@ def test_one_cycle_runner_hook():
     ]
     hook.writer.add_scalars.assert_has_calls(calls, any_order=True)
 
+    # Test OneCycleLrUpdaterHook
```
zhouzaida, line 319: Could we use `pytest.mark.parametrize` to remove redundant code like https://github.com/open-mmlab/mmcv/blob/master/tests/test_runner/test_hooks.py#L398

ZwwWayne:
lint failed.

zhouzaida (approved):
LGTM

ZwwWayne:
CI failed

PR #937, opened 2021-04-11
[Feature] Add PyTorch Profiler to analyze training process

zhouzaida:
Profiler is a tool that allows the collection of the performance metrics during the training and inference. More details on Profiler can be found at https://pytorch.org/docs/1.8.1/profiler.html#torch.profiler.profile

```python
import torch
import torch.nn as nn
import torch.nn.functional as F
import torch.optim as optim
from torch.utils.data import DataLoader
import torchvision
from torchvision.datasets import CIFAR10
import torchvision.transforms as transforms

from mmcv.parallel import MMDataParallel
from mmcv.runner import EpochBasedRunner
from mmcv.utils import get_logger


class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()
        self.conv1 = nn.Conv2d(3, 6, 5)
        self.pool = nn.MaxPool2d(2, 2)
        self.conv2 = nn.Conv2d(6, 16, 5)
        self.fc1 = nn.Linear(16 * 5 * 5, 120)
        self.fc2 = nn.Linear(120, 84)
        self.fc3 = nn.Linear(84, 10)
 
    def forward(self, x):
        x = self.pool(F.relu(self.conv1(x)))
        x = self.pool(F.relu(self.conv2(x)))
        x = x.view(-1, 16 * 5 * 5)
        x = F.relu(self.fc1(x))
        x = F.relu(self.fc2(x))
        x = self.fc3(x)
        return x
 
    def train_step(self, data, optimizer):  # optimizer is unused and reversed.
        images, labels = data
        predicts = self(images)  # -> self.__call__() -> self.forward()
        loss = F.cross_entropy(predicts, labels)
        return {'loss': loss}


# instantiate a model
model = Model()
# wrap model by MMDataParallel
model = MMDataParallel(model.cuda())

# dataset
transform = transforms.Compose(
    [transforms.ToTensor(),
    transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
trainset = CIFAR10(root='data', train=True, download=True, transform=transform)
trainloader = DataLoader(trainset, batch_size=128, shuffle=True, num_workers=4)

# runner
optimizer = optim.SGD(model.parameters(), lr=0.001, momentum=0.9)
runner = EpochBasedRunner(model, optimizer=optimizer, work_dir='./work_dir',
                          logger=get_logger('mmcv'), max_epochs=1)

# optimizer hook
optimizer_config = dict(grad_clip=None)
# log hook
lr_config = dict(policy='step', step=[2, 3])
checkpoint_config = dict(interval=1)
log_config = dict(interval=100, hooks=[dict(type='TextLoggerHook')])
runner.register_training_hooks(lr_config=lr_config,
                               optimizer_config=optimizer_config,
                               checkpoint_config=checkpoint_config,
                               log_config=log_config)
# profiler hook
trace_config = dict(type='tb_trace', dir_name='./work_dir')
profiler_config = dict(on_trace_ready=trace_config, record_shapes=True)
runner.register_profiler_hook(profiler_config)

# training
runner.run(data_loaders=[trainloader], workflow=[('train', 1)])

```

```bash
> tensorboard --logdir=work_dir
```

-  num_worker=2

localhost_6006_

localhost_6006_ (1)

Modify the `num_worker` according to the `Performance Recommendation`.

- num_worker=4

localhost_6006_ (2)

- num_worker=8

image

As we can see, the consumption time of `DataLoader` is reduced.

Reference:

- https://pytorch.org/blog/introducing-pytorch-profiler-the-new-and-improved-performance-tool/
- https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html
- https://pytorch.org/docs/stable/profiler.html?highlight=profiler#module-torch.profiler


zhouzaida:
ping @xvjiarui

mmcv/runner/hooks/profiler.py
```diff
@@ -0,0 +1,160 @@
...
+
+import torch
+
+from ..dist_utils import master_only
+from .hook import HOOKS, Hook
+
+
+@HOOKS.register_module()
+class ProfilerHook(Hook):
+    """Profiler context manager.
```
xvjiarui, line 13: `"context manager"` may not be accurate. Maybe profiler to analyze training process.

mmcv/runner/hooks/profiler.py
```diff
@@ -0,0 +1,160 @@
...
+            activities=self.activities,
+            schedule=self.schedule,
+            on_trace_ready=_on_trace_ready,
+            record_shapes=self.record_shapes,
+            profile_memory=self.profile_memory,
+            with_stack=self.with_stack,
+            with_flops=self.with_flops)
+
+        self.profiler.__enter__()
+
```
xvjiarui, line 146: We may log something before profiler begins.

mmcv/runner/hooks/profiler.py
```diff
@@ -0,0 +1,160 @@
...
+        self.with_stack = with_stack
+        self.with_flops = with_flops
+        self.json_trace_path = json_trace_path
+
+    @master_only
+    def before_run(self, runner):
+        if callable(self.on_trace_ready):  # handler
+            _on_trace_ready = self.on_trace_ready
+        elif isinstance(self.on_trace_ready, dict):  # config of handler
+            trace_type = self.on_trace_ready.pop('type')  # log_trace handler
```
xvjiarui, line 105: Maybe we should do something like this:
```suggestion
            on_trace_ready = self.on_trace_ready.copy()
            trace_type = on_trace_ready.pop('type')  # log_trace handler
```
Avoiding change the dict in-place.

mmcv/runner/hooks/profiler.py
```diff
@@ -0,0 +1,160 @@
...
+            _on_trace_ready = None  # type: ignore
+        else:
+            raise ValueError('on_trace_ready should be handler, dict or None, '
+                             f'but got {type(self.on_trace_ready)}')
+
+        if runner.max_epochs > 1:
+            warnings.warn(f'epoch should be 1, but got {runner.max_epochs}. '
+                          'profiler will slow down your training, so it is '
+                          'not advised to set epoch greater than 1 if you use '
+                          'ProfilerHook.')
```
xvjiarui, line 134:
```
Profiler will profile {runner.max_epochs} epochs instead of 1 epoch
Since profiler will slow down the training, it is recommended to train 1 epoch with ProfilerHook and adjust your setting according to the profiler summary. During normal training (epoch > 1), you may disable the ProfilerHook. 
```

mmcv/runner/hooks/profiler.py
zhouzaida: got it

xvjiarui (approved):

mmcv/runner/hooks/profiler.py
```diff
@@ -0,0 +1,154 @@
...
+                raise ValueError('trace_type should be "log_trace" or '
+                                 f'"tb_trace", but got {trace_type}')
+        elif self.on_trace_ready is None:
+            _on_trace_ready = None  # type: ignore
+        else:
+            raise ValueError('on_trace_ready should be handler, dict or None, '
+                             f'but got {type(self.on_trace_ready)}')
+
+        if runner.max_epochs > 1:
+            warnings.warn(f'profiler will profile {runner.max_epochs} epochs '
```
ZwwWayne, line 122: Shall we simply add an option that allows the users to only profile first few epochs or the first few iterations? This might be more useful as we do not need to profile it by training 1 extra epoch but can do that during the normal training process.

mmcv/runner/hooks/profiler.py
```diff
@@ -10,14 +10,18 @@
...
-    metrics during the training and inference. More details on Profiler can
-    be found at
+    metrics during the training. More details on Profiler can be found at
     https://pytorch.org/docs/1.8.1/profiler.html#torch.profiler.profile
 
     Args:
+        by_epoch (bool): Profile performance by epoch or by iteration.
+            Default: True.
+        profile_iters (int): Number of iterations for profiling.
+            If ``by_epoch=True``, num_iters indicates epochs, otherwise it
```
ZwwWayne, line 23: 1. num_iters -> profile_iters
2. Indicate that they are the first xxx iterations/epochs at the beginning of the training.

PR #945, opened 2021-04-14
Compile mmcv with custom op when not using cuda

tangyanf:
When I compile mmcv with custom op on a machine that does not have cuda installed, I will encounter the following error：
image
So I modified the code to get the cuda header file path when compiling Ort custom op in setup.py
image


zhouzaida (approved):
LGTM

PR #791, opened 2021-01-16
Support auto_fp16 using torch.cuda.amp

ycxioooong:
Testing using mmdet faster_rcnn_r50_fpn_fp16_1x_coco.

Ubuntu 18.04
torch1.7.1
CUDA10.2
GPU: 2080Ti

Train speed 0.3s/iter;
Memory: 2.8GB;
Inference: 99fps;
mAP 37.4 (compared with previous fp16 implementation 37.5)

PR #951, opened 2021-04-15
Support auto_fp16 using torch.cuda.amp when PyTorch >= 1.6.0

ycxioooong:
This PR enables PyTorch official implementation for automatic mixed-precision training.
It replaces the original [Pull Request](https://github.com/open-mmlab/mmcv/pull/791) due to some accident.

### DIFF
--- a/mmcv/runner/fp16_utils.py
+++ b/mmcv/runner/fp16_utils.py
@@ -13,6 +13,8 @@ from .dist_utils import allreduce_grads as _allreduce_grads
 try:
     # If PyTorch version >= 1.6.0, torch.cuda.amp.autocast would be imported
     # and used; otherwise, auto fp16 will adopt mmcv's implementation.
+    # Note that when PyTorch >= 1.6.0, we still cast tensor types to fp16
+    # manually, so the behavior may not be consistant with real amp.
     from torch.cuda.amp import autocast
 except ImportError:
     pass
--- a/mmcv/runner/fp16_utils.py
+++ b/mmcv/runner/fp16_utils.py
@@ -118,8 +118,8 @@ def auto_fp16(apply_to=None, out_fp32=False):
                         new_kwargs[arg_name] = arg_value
             # apply converted arguments to the decorated method
             if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
-                output = autocast(enabled=True)(old_func)(*new_args,
-                                                          **new_kwargs)
+                with autocast(enabled=True):
+                    output = old_func(*new_args, **new_kwargs)
             else:
                 output = old_func(*new_args, **new_kwargs)
             # cast the results back to fp32 if necessary
@@ -203,8 +203,8 @@ def force_fp32(apply_to=None, out_fp16=False):
                         new_kwargs[arg_name] = arg_value
             # apply converted arguments to the decorated method
             if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
-                output = autocast(enabled=False)(old_func)(*new_args,
-                                                           **new_kwargs)
+                with autocast(enabled=False):
+                    output = old_func(*new_args, **new_kwargs)
             else:
                 output = old_func(*new_args, **new_kwargs)
             # cast the results back to fp32 if necessary
--- a/mmcv/runner/fp16_utils.py
+++ b/mmcv/runner/fp16_utils.py
@@ -7,7 +7,15 @@ import numpy as np
 import torch
 import torch.nn as nn
 
+from mmcv.utils import TORCH_VERSION
 from .dist_utils import allreduce_grads as _allreduce_grads
+
+try:
+    # If PyTorch version >= 1.6.0, torch.cuda.amp.autocast would be imported
+    # and used; otherwise, auto fp16 will adopt mmcv's implementation.
+    from torch.cuda.amp import autocast
+except ImportError:
+    pass
 
 
 def cast_tensor_type(inputs, src_type, dst_type):
@@ -45,7 +53,8 @@ def auto_fp16(apply_to=None, out_fp32=False):
     This decorator is useful when you write custom modules and want to support
     mixed precision training. If inputs arguments are fp32 tensors, they will
     be converted to fp16 automatically. Arguments other than fp32 tensors are
-    ignored.
+    ignored. If you are using PyTorch >= 1.6, torch.cuda.amp is used as the
+    backend, otherwise, original mmcv implementation will be adopted.
 
     Args:
         apply_to (Iterable, optional): The argument names to be converted.
@@ -82,6 +91,7 @@ def auto_fp16(apply_to=None, out_fp32=False):
                                 'method of nn.Module')
             if not (hasattr(args[0], 'fp16_enabled') and args[0].fp16_enabled):
                 return old_func(*args, **kwargs)
+
             # get the arg spec of the decorated method
             args_info = getfullargspec(old_func)
             # get the argument names to be casted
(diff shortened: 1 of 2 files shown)


mmcv/runner/fp16_utils.py
```diff
@@ -107,7 +117,11 @@ def new_func(*args, **kwargs):
                     else:
                         new_kwargs[arg_name] = arg_value
             # apply converted arguments to the decorated method
-            output = old_func(*new_args, **new_kwargs)
+            if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
```
zhouzaida, line 120:
```python
with autocast(enabled=False):
      output = old_func(*new_args,  **new_kwargs)
```
will be more clear.

another question: If we use `autocast`, do we need to manually convert the `args` and `kwargs`  by invoking `cast_tensor_type`?

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,128 +42,234 @@ def after_train_iter(self, runner):
...
+                If loss_scale is a float, static loss scaling will be used with
+                the specified scale. If loss_scale is a string, it must be
+                'dynamic', then dynamic loss scaling will be used.
+                It can also be a dict containing arguments of LossScaler.
+                Defaults to 512. For Pytorch >= 1.6, mmcv uses official
+                implementation of GradScaler. If you use a dict version of
+                loss_scale to create GradScaler, plese refer to:
+                https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler
+                for the parameters. Here is an example:
+                loss_scale = dict(
```
ZwwWayne, line 65: Should use `Examples:`?

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,128 +42,234 @@ def after_train_iter(self, runner):
...
-        5. Copy back the params from fp32 weight copy to the fp16 model.
+    @HOOKS.register_module()
+    class Fp16OptimizerHook(OptimizerHook):
+        """FP16 optimizer hook (using PyTorch's implementation).
+
+        If you are using PyTorch >= 1.6, torch.cuda.amp is used as the backend,
+        to take care of the optimization procedure.
+
+        Args:
+            loss_scale (float | str | dict): Scale factor multiplied with loss.
```
ZwwWayne, line 55: The docstring is not consistent with the real arguments.

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,128 +42,234 @@ def after_train_iter(self, runner):
...
+        1. Scale the loss value.
+        2. BP in the fp16 model.
+        2. Copy gradients from fp16 model to fp32 weights.
+        3. Update fp32 weights.
+        4. Copy updated parameters from fp32 weights to fp16 model.
+
+        Refer to https://arxiv.org/abs/1710.03740 for more details.
+
+        Args:
+            loss_scale (float | str | dict): Scale factor multiplied with loss.
```
ZwwWayne, line 158: Update the arguments.

mmcv/runner/fp16_utils.py
```diff
@@ -207,16 +227,25 @@ def allreduce_grads(params, coalesce=True, bucket_size_mb=-1):
...
+    If you are using PyTorch >= 1.6, torch.cuda.amp is used as the
+    backend, otherwise, original mmcv implementation will be adopted.
+
+    For PyTorch >= 1.6, this function will
+    1. Set fp16 flag inside the model to True.
+
+    Otherwise:
     1. Convert FP32 model to FP16.
     2. Remain some necessary layers to be FP32, e.g., normalization layers.
+    3. Set fp16 flag inside the model to True.
```
ZwwWayne, line 239: Set fp16 flag `fp16_enabled` inside the modules of model to True.

mmcv/runner/fp16_utils.py
```diff
@@ -186,7 +202,11 @@ def new_func(*args, **kwargs):
                     else:
                         new_kwargs[arg_name] = arg_value
             # apply converted arguments to the decorated method
-            output = old_func(*new_args, **new_kwargs)
+            if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
+                output = autocast(enabled=False)(old_func)(*new_args,
```
ZwwWayne, line 206: It seems that the key difference is to use  `autocast` if we can. So shall we set the case in the header first like
```python
try:
    from torch.cuda.amp import autocast
except ImportError:
    autocast = None
```

Then simply use the condition like `if iscallable(autocast)` here, rather than checking the version of pytorch

ZwwWayne:
I am benchmarking the mmdetection3d and mmdetection with the new FP16 feature. Will update here after finishing them (roughly 4.23)

mmcv/runner/fp16_utils.py
```diff
@@ -82,6 +91,7 @@ def new_func(*args, **kwargs):
                                 'method of nn.Module')
             if not (hasattr(args[0], 'fp16_enabled') and args[0].fp16_enabled):
```
nbei, line 92: I may suggest that we can enforce the input tensor to be float32 in this case, so that `auto_fp16` can be used in intermediate layers in truly mixed-precision training.

mmcv/runner/fp16_utils.py
ycxioooong: fixed using with. Checked.

Also added docstring to explain why we still use cast_tensor_type

mmcv/runner/fp16_utils.py
ycxioooong: fixed

mmcv/runner/hooks/optimizer.py
ycxioooong: fixed

mmcv/runner/hooks/optimizer.py
ycxioooong: fixed

mmcv/runner/hooks/optimizer.py
ycxioooong: fixed

mmcv/runner/fp16_utils.py
ycxioooong: Actually in previous PyTorch versions (e.g. 1.5), `torch.cuda.amp` is also implemented but we will see warnings that showing the `torch.cuda.amp` is experimental. So using version as the indicator is more reliable here in our case.

mmcv/runner/fp16_utils.py
ycxioooong: leave it later

mmcv/runner/fp16_utils.py
ZwwWayne: OK

ZwwWayne (approved):

xvjiarui (approved):
Verified in MMSeg

mmcv/runner/hooks/optimizer.py
```diff
@@ -34,128 +42,237 @@ def after_train_iter(self, runner):
...
-            if fp16_param.grad is not None:
-                if fp32_param.grad is None:
-                    fp32_param.grad = fp32_param.data.new(fp32_param.size())
-                fp32_param.grad.copy_(fp16_param.grad)
-
-    def copy_params_to_fp16(self, fp16_net, fp32_weights):
-        """Copy updated params from fp32 weight copy to fp16 model."""
-        for fp16_param, fp32_param in zip(fp16_net.parameters(), fp32_weights):
-            fp16_param.data.copy_(fp32_param.data)
+if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
```
ZwwWayne, line 45: Last comment: The indentation is a little bit annoying. Can we create two hooks first, namely `PT16Fp16OptimizerHook` and `PT15Fp16OptimizerHook`, then assign one of them to `Fp16OptimizerHook` according to the version?

PR #916, opened 2021-03-31
[feature]:add onnxruntime custom op grid_sample

tangyanf:
this pr add a onnxruntime custom op grid_sample() based on this pr #865

PR #953, opened 2021-04-15
[Feature] enable exporting to onnx for PointRend

DmitriySidnev:
This PR enables exporting to ONNX for `PointRend` based models.


mmcv/ops/point_sample.py
```diff
@@ -70,46 +120,57 @@ def rel_roi_point_to_abs_img_point(rois, rel_roi_points):
...
+
+def abs_img_point_to_rel_img_point(abs_img_points, img, spatial_scale=1.):
     """Convert image based absolute point coordinates to image based relative
     coordinates for sampling.
 
     Args:
         abs_img_points (Tensor): Image based absolute point coordinates,
             shape (N, P, 2)
-        img_shape (tuple): (height, width) of image or feature map.
+        img (tuple): (height, width) of image or feature map.
```
RunningLeon, line 148: Could update the docstring here.

mmcv/ops/point_sample.py
```diff
@@ -70,46 +120,57 @@ def rel_roi_point_to_abs_img_point(rois, rel_roi_points):
...
+    else:
+        scale = get_shape_from_feature_map(img)
+
+    return abs_img_points / scale * spatial_scale
 
 
 def rel_roi_point_to_rel_img_point(rois,
                                    rel_roi_points,
-                                   img_shape,
+                                   img,
```
RunningLeon, line 173: Could update the docstring for the argument.

mmcv/ops/point_sample.py
```diff
@@ -153,8 +214,12 @@ def point_sample(input, points, align_corners=False, **kwargs):
     if points.dim() == 3:
         add_dim = True
         points = points.unsqueeze(2)
-    output = F.grid_sample(
-        input, denormalize(points), align_corners=align_corners, **kwargs)
+    if torch.onnx.is_in_onnx_export():
+        output = bilinear_grid_sample(
```
RunningLeon, line 218: Could you explain why we have this difference for exporting and the normal Pytorch pipeline?

mmcv/ops/point_sample.py
```diff
@@ -181,25 +246,32 @@ def __init__(self, output_size, spatial_scale, aligned=True):
...
-            if inds.any():
-                rel_img_points = rel_roi_point_to_rel_img_point(
-                    rois[inds], rel_roi_points[inds], feat.shape[2:],
-                    self.spatial_scale).unsqueeze(0)
-                point_feat = point_sample(
-                    feat, rel_img_points, align_corners=not self.aligned)
-                point_feat = point_feat.squeeze(0).transpose(0, 1)
-                point_feats.append(point_feat)
+        if torch.onnx.is_in_onnx_export():
+            feat = features[0].unsqueeze(0)
```
RunningLeon, line 255: Do we support exporting ONNX with batch here?

mmcv/ops/point_sample.py
DmitriySidnev: Fixed type.

mmcv/ops/point_sample.py
DmitriySidnev: Fixed type.

mmcv/ops/point_sample.py
DmitriySidnev: ONNX does not support pytorch's operation `grid_sample` and during export we should not use it, so I implemented this operation in supported operations that can be exported and inferenced without any issues. [Here](https://github.com/pytorch/pytorch/issues/27212) is related issue in pytorch.

mmcv/ops/point_sample.py
DmitriySidnev: No, it works only with batch size = 1. But as I understand, according to [this](https://github.com/open-mmlab/mmdetection/blob/master/tools/deployment/pytorch2onnx.py#L15) export script in `mmdetection`, batch > 1 is impossible here.

RunningLeon: Hi, currently, we are supporting exporting ONNX with batch and dynamic shape on [onnx branch](https://github.com/open-mmlab/mmdetection/tree/onnx) and we are planning to merge the onnx branch into master branch. It would be great if you could try to support `PointRend` model with batch and dynamic shape if possible.

mmcv/ops/point_sample.py
RunningLeon: There's a `grid_sampler` op for ONNXRuntime as custom op from this PR #916, could you check with that?

mmcv/ops/point_sample.py
RunningLeon: Line 182 should be updated as well.

mmcv/ops/point_sample.py
DmitriySidnev: I tried the custom implementation. Model converted to onnx but I had the following runtime error during loading a model:
```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from mmcv_grid_sample.onnx failed:Fatal error: grid_sampler is not a registered function/op
```
My implementation converted and works in onnx runtime. [Here](https://github.com/DmitriySidnev/mmcv/blob/ds/grid_sample_test/grid_sample_test.py) is the script used for test. B.t.w. I did not manage to install `mmcv-full` module with onnx runtime ops via pip. Only from sources.

mmcv/ops/point_sample.py
DmitriySidnev: Oh, yes. Fixed.

mmcv/ops/point_sample.py
DmitriySidnev: The original implementation converted with batch > 1. My bad. But in my case this fix is necessary. I added one more condition where is checked batch size. If it is possible I would like to remain code under `if torch.onnx.is_in_onnx_export() and num_imgs == 1`.

mmcv/ops/point_sample.py
RunningLeon: Did you install mmcv-full by following the instruction from [the doc](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md#how-to-build-custom-operators-for-onnx-runtime)?

RunningLeon:
@DmitriySidnev To fix lint, could you merge master and push again?

mmcv/ops/point_sample.py
RunningLeon: Could you start with ONNX branch and support exporting ONNX with batch and dynamic shape if possible?

RunningLeon: Supporting Point Rend with batch and dynamic shape may need much work. We may do this together once the time is available.

mmcv/ops/point_sample.py
DmitriySidnev: Yes, I used the instruction. I meant that command `pip install mmcv-full` with environment variable `MMCV_WITH_ORT=1` does not work without `onnxruntime` library that should be configured before.

mmcv/ops/point_sample.py
DmitriySidnev: So, the code supports batch and dynamic shapes in onnx. [Here](https://github.com/DmitriySidnev/mmcv/blob/ds/grid_sample_test/grid_sample_test.py) is my test with `SimpleRoiAlign` class as an exporting model with the following dynamic axes: batch, height and width for input image and objects number for RoIs. But in my case, onnx model is an intermediate one and I need to avoid the loop through batch and get rid of indexing, it brakes my target graph in `OpenVINO`.

mmcv/ops/point_sample.py
RunningLeon: You could try to install mmcv-full from source following the instruction from [the doc](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md#how-to-build-custom-operators-for-onnx-runtime). `grid_sampler` should work for onnxruntime backend.

DmitriySidnev: Yes, it works and seems that my implementation is slower in onnx runtime. But I propose to use my implementation during exporting when custom ops not installed.

mmcv/ops/point_sample.py
RunningLeon: Yes, once we support exporing to ONNX with batch, then there's no loop on batch dim in the ONNX file.

DmitriySidnev: In `OpenVINO` it leads to unsupported dynamic ops. Avoiding this loop fixes the result graph.

RunningLeon: Do you mean the exported PointRend ONNX file from the master branch would have `Loop` op in the graph?

mmcv/ops/point_sample.py
RunningLeon: It could be added. You could add the unit tests for your implementation for both PyTorch and pytorch2onnx.

mmcv/ops/point_sample.py
DmitriySidnev: No, I mean that `OpenVINO` wrong processes indexing generated in loop. I do not know how exactly it works but removing the loop helps.

mmcv/ops/point_sample.py
DmitriySidnev: Do you mean tests for `bilinear_grid_sample` function?

RunningLeon: Yes.

DmitriySidnev: Tests added.

mmcv/ops/point_sample.py
```diff
@@ -181,25 +255,32 @@ def __init__(self, output_size, spatial_scale, aligned=True):
...
-            if inds.any():
-                rel_img_points = rel_roi_point_to_rel_img_point(
-                    rois[inds], rel_roi_points[inds], feat.shape[2:],
-                    self.spatial_scale).unsqueeze(0)
-                point_feat = point_sample(
-                    feat, rel_img_points, align_corners=not self.aligned)
-                point_feat = point_feat.squeeze(0).transpose(0, 1)
-                point_feats.append(point_feat)
+        if torch.onnx.is_in_onnx_export() and num_imgs == 1:
+            feat = features[0].unsqueeze(0)
```
RunningLeon, line 264: Isn't `features` here is a Tensor of NCHW? It seems `features[0].unsqueeze(0)` makes no sense here.

DmitriySidnev: Yes, you are right. Removed.

mmcv/ops/point_sample.py
```diff
@@ -1,9 +1,80 @@
...
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.modules.utils import _pair
+from torch.onnx.operators import shape_as_tensor
+
+from mmcv.ops import get_onnxruntime_op_path
+
+
+def bilinear_grid_sample(im, grid, align_corners=False):
```
RunningLeon, line 14: We could add docstring for the function `bilinear_grid_sample`

mmcv/ops/point_sample.py
```diff
@@ -70,46 +141,57 @@ def rel_roi_point_to_abs_img_point(rois, rel_roi_points):
         if rois.size(1) == 5:
             rois = rois[:, 1:]
         abs_img_points = rel_roi_points.clone()
-        abs_img_points[:, :, 0] = abs_img_points[:, :, 0] * (
-            rois[:, None, 2] - rois[:, None, 0])
-        abs_img_points[:, :, 1] = abs_img_points[:, :, 1] * (
-            rois[:, None, 3] - rois[:, None, 1])
-        abs_img_points[:, :, 0] += rois[:, None, 0]
-        abs_img_points[:, :, 1] += rois[:, None, 1]
+        xs = abs_img_points[:, :, 0] * (rois[:, None, 2] - rois[:, None, 0])
```
RunningLeon, line 144: Add comment here on why we have to do this update here.

mmcv/ops/point_sample.py
```diff
@@ -70,46 +141,57 @@ def rel_roi_point_to_abs_img_point(rois, rel_roi_points):
...
+        xs += rois[:, None, 0]
+        ys += rois[:, None, 1]
+        abs_img_points = torch.stack([xs, ys], dim=2)
     return abs_img_points
 
 
-def abs_img_point_to_rel_img_point(abs_img_points,
-                                   img_shape,
-                                   spatial_scale=1.):
+def get_shape_from_feature_map(x):
```
RunningLeon, line 152: we could add docstring here.

mmcv/ops/point_sample.py
```diff
@@ -153,8 +235,12 @@ def point_sample(input, points, align_corners=False, **kwargs):
     if points.dim() == 3:
         add_dim = True
         points = points.unsqueeze(2)
-    output = F.grid_sample(
-        input, denormalize(points), align_corners=align_corners, **kwargs)
+    if is_in_onnx_export_without_custom_ops():
```
RunningLeon, line 238: could add comment on why do this change here.

RunningLeon (changes requested):

mmcv/ops/point_sample.py
DmitriySidnev: Added.

mmcv/ops/point_sample.py
DmitriySidnev: Added.

mmcv/ops/point_sample.py
DmitriySidnev: Added.

mmcv/ops/point_sample.py
DmitriySidnev: Added.

mmcv/ops/point_sample.py
```diff
@@ -45,15 +45,16 @@ def bilinear_grid_sample(im, grid, align_corners=False):
...
-    y1[y1 < 0] = 0
-    y1[y1 > padded_h - 1] = padded_h - 1
+    x0 = (torch.clamp(x0.float() / (padded_w - 1), 0, 1) *
+          (padded_w - 1)).long()
+    x1 = (torch.clamp(x1.float() / (padded_w - 1), 0, 1) *
+          (padded_w - 1)).long()
+    y0 = (torch.clamp(y0.float() / (padded_h - 1), 0, 1) *
+          (padded_h - 1)).long()
+    y1 = (torch.clamp(y1.float() / (padded_h - 1), 0, 1) *
+          (padded_h - 1)).long()
```
DmitriySidnev, line 72: @RunningLeon, this fix leads to a metrics drop. My variation works with the same quality as custom ort op.

RunningLeon: This change is to support dynamic shape. If it results in a performance drop, you could post your tested results here.

RunningLeon: Kindly ping @DmitriySidnev

DmitriySidnev: @RunningLeon, I am collecting metrics in onnx runtime on cpu. It is very slow.

RunningLeon: @DmitriySidnev, Yes. It's quite slow with onnxruntime on cpu.

DmitriySidnev: My results with `clamp` function:
| Config                                      | Runtime | Bbox mAP | Mask mAP |
|---------------------------------------------|---------|----------|----------|
| point_rend_r50_caffe_fpn_mstrain_3x_coco.py | Pytorch | 41.0     | 38.0     |
| point_rend_r50_caffe_fpn_mstrain_3x_coco.py | ONNX    | 40.6     | 36.1     |

Here is a huge drop for segmentation mAP. Also I will collect metrics without `clamp` function to compare results.

RunningLeon: @DmitriySidnev,  Hi, if the performace of with `clamp` is worse than without `clamp`, please also post tested results here to verify this issue.

DmitriySidnev: Here is metrics without the `clamp` function (sorry for the delay, on cpu it works too slow):
| Config                                      | Runtime | Bbox mAP | Mask mAP |
|---------------------------------------------|---------|----------|----------|
| point_rend_r50_caffe_fpn_mstrain_3x_coco.py | ONNX    | 40.6     | 36.7     |

So, mAP for masks is better compared to the model with `clamp` function in python implementation of `grid_sampler`. But there is one more point from my side: presence of `clamp` function does not fix dynamic batch (nothing to fix). Dynamic batch works without it. I propose to retrieve my code.

RunningLeon: @DmitriySidnev, thanks for the effort of testing. Actually `clamp` is for dynamic shape with different input height and width.
If there's no need for dynamic, it could be replaced with the original style.

RunningLeon: @DmitriySidnev You may need to update the code accordingly.

DmitriySidnev: Done.

zhouzaida:
hi @DmitriySidnev , CI failed.

DmitriySidnev:
@zhouzaida, hi! I have merged master branch. Hope it helps. With `onnxruntime==1.5.1` all onnx tests passed successfully.

zhouzaida:
> @zhouzaida, hi! I have merged master branch. Hope it helps. With `onnxruntime==1.5.1` all onnx tests passed successfully.

got it

DmitriySidnev:
In the failed job `build / build_cpu (3.7, 1.7.0) (pull_request)` I can see the following lines:
```
Collecting onnxruntime==1.4.0
  Downloading onnxruntime-1.4.0-cp37-cp37m-manylinux2010_x86_64.whl (4.4 MB)
```
But this version of onnxruntime does not satisfy [this statement](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md#how-to-build-custom-operators-for-onnx-runtime).

RunningLeon:
@DmitriySidnev You could find the reason why `test_bilinear_grid_sample` failed with `torch==1.7.0` on cpu.

DmitriySidnev:
@RunningLeon, `bilinear_grid_sample` function fixed. In onnxruntime 1.5.1 onnx models obtained from pytorch 1.7.0 and 1.8.1 now works.

RunningLeon:
@DmitriySidnev Thanks for the effort. Could you fix CI error? Thanks again.

DmitriySidnev:
@RunningLeon, the `bilinear_grid_sample` function fixed for pytorch < 1.8.

RunningLeon:
kindly ping @ZwwWayne

mmcv/ops/point_sample.py
```diff
@@ -1,9 +1,95 @@
...
 
+from os import path as osp
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.modules.utils import _pair
+from torch.onnx.operators import shape_as_tensor
+
+from mmcv.ops import get_onnxruntime_op_path
```
RunningLeon, line 11: We could move this import to line 90(inside is_in_onnx_export_without_custom_ops) to avoid import the module when there is no onnx exporting.

DmitriySidnev: Fixed.

RunningLeon (approved):

RunningLeon:
LGTM. Kindly ping @ZwwWayne

RangiLyu (approved):
LGTM

PR #957, opened 2021-04-16
[Feature] NMS update

SemyonBevzuk:
Added support for "score_threshold" and "max_num" for NMS operation when exporting to ONNX.
These parameters are supported in ONNX: [link](https://github.com/onnx/onnx/blob/master/docs/Operators.md#NonMaxSuppression).

These changes significantly speed up the export time:
|  Model name | Export time before (s) | Export time now (s) |
|:-----------:|:----------------------:|:----------------------:|
| ssd_300     |         1201.27        |          6.85          |
| yolo_v3     |          88.89         |          73.11         |
| fsaf        |          19.89         |          12.62         |
| retinanet   |          59.33         |          13.30         |
| faster_rcnn |          21.24         |          18.02         |

In addition, performance is improved in ONNX Runtime.
This can be seen in the example of the RetinaNet model.
Performance measured using standard ONNX Runtime tools on 10 random inputs.
The table shows the average execution time of the NMS operation and the total inference time of the model:
| RetinaNet (ONNX Runtime) | Befor (s) | Now (s) |
|-------------------|------------|----------|
| NMS time | 15.99  | 0.0002574   |
|Total time | 16.78 | 0.76766 |


mmcv/ops/nms.py
```diff
@@ -90,7 +104,7 @@ def symbolic(g, boxes, scores, iou_threshold, sigma, min_score, method,
 
 
 @deprecated_api_warning({'iou_thr': 'iou_threshold'})
-def nms(boxes, scores, iou_threshold, offset=0):
+def nms(boxes, scores, iou_threshold, score_threshold=0, max_num=-1, offset=0):
```
RunningLeon, line 107: Could update docstring for argument `score_threshold` and `max_num`.

RunningLeon, line 107: The call of `nms` in Line 306 would be a problem,  because it would reduce `max_num` boxes for each split.

SemyonBevzuk: Done.

SemyonBevzuk: I used `mmdetection/tools/test.py` to collect quality metrics and did not notice any difference in `batched_nms` between if branches when using `max_num`.

mmcv/ops/nms.py
```diff
@@ -15,13 +14,21 @@
 class NMSop(torch.autograd.Function):
 
     @staticmethod
-    def forward(ctx, bboxes, scores, iou_threshold, offset):
+    def forward(ctx, bboxes, scores, iou_threshold, score_threshold, max_num,
```
RunningLeon, line 17: `score_threshold` and `max_num` arguments should be added after `offset`  to keep the compatibility.

mmcv/ops/nms.py
```diff
@@ -15,13 +14,21 @@
...
+
         inds = ext_module.nms(
             bboxes, scores, iou_threshold=float(iou_threshold), offset=offset)
+
+        inds = valid_inds[inds[:max_num]]
         return inds
 
     @staticmethod
-    def symbolic(g, bboxes, scores, iou_threshold, offset):
+    def symbolic(g, bboxes, scores, iou_threshold, score_threshold, max_num,
```
RunningLeon, line 30: `score_threshold` and `max_num` arguments should be added after `offset`  to keep the compatibility.

mmcv/ops/nms.py
```diff
@@ -90,7 +105,7 @@ def symbolic(g, boxes, scores, iou_threshold, sigma, min_score, method,
 
 
 @deprecated_api_warning({'iou_thr': 'iou_threshold'})
-def nms(boxes, scores, iou_threshold, offset=0):
+def nms(boxes, scores, iou_threshold, score_threshold=0, max_num=-1, offset=0):
```
RunningLeon, line 108: `score_threshold` and `max_num` arguments should be added after `offset`  to keep the compatibility.

mmcv/ops/nms.py
RunningLeon: It may have a bad influence for training where there are a large amount of `bboxes`. You could post models and their mertics in the PR description as well.

SemyonBevzuk: This potential problem can be worked around by moving the changes to `if torch.onnx.is_in_onnx_export()`. Is it okay?

mmcv/ops/nms.py
```diff
@@ -14,21 +14,24 @@
...
-                offset):
-        valid_mask = scores > score_threshold
-        bboxes, scores = bboxes[valid_mask], scores[valid_mask]
-        valid_inds = torch.nonzero(valid_mask, as_tuple=False).squeeze(dim=1)
+    def forward(ctx, bboxes, scores, iou_threshold, offset, score_threshold,
+                max_num):
+        if torch.onnx.is_in_onnx_export():
+            valid_mask = scores > score_threshold
+            bboxes, scores = bboxes[valid_mask], scores[valid_mask]
+            valid_inds = torch.nonzero(
```
RunningLeon, line 22: NonZero op is dynamic and should be avoided because we have to support TensorRT as well, which do not have NonZero op.

SemyonBevzuk: If TensorRT works with ONNX graph and supports NMS operation, this shouldn't be a problem.

When the `symbolic` function is implemented, it replaces the `forward` function in the ONNX graph.
So there should be no problem with dynamic NonZero.

This is how the current NMS operation looks in the ONNX graph:
nms

RunningLeon: In that case, it's OK.

mmcv/ops/nms.py
RunningLeon: It could be a workaround.

mmcv/ops/nms.py
```diff
@@ -15,13 +14,24 @@
 class NMSop(torch.autograd.Function):
 
     @staticmethod
-    def forward(ctx, bboxes, scores, iou_threshold, offset):
+    def forward(ctx, bboxes, scores, iou_threshold, offset, score_threshold,
+                max_num):
+        if torch.onnx.is_in_onnx_export():
```
RunningLeon, line 19: There may be too many `if torch.onnx.is_in_onnx_export()` and better to simplify this.
For example, we only pass score_threshold > 0 and max_num > 0 if torch.onnx.is_in_onnx_export().
Here we could replace with `if score_threshold > 0`

mmcv/ops/nms.py
```diff
@@ -15,13 +14,24 @@
...
+            valid_mask = scores > score_threshold
+            bboxes, scores = bboxes[valid_mask], scores[valid_mask]
+            valid_inds = torch.nonzero(
+                valid_mask, as_tuple=False).squeeze(dim=1)
+
         inds = ext_module.nms(
             bboxes, scores, iou_threshold=float(iou_threshold), offset=offset)
+
+        if torch.onnx.is_in_onnx_export():
```
RunningLeon, line 28: To simplify, we could replace with `if max_num > 0`

mmcv/ops/nms.py
```diff
@@ -141,7 +161,10 @@ def nms(boxes, scores, iou_threshold, offset=0):
         }
         inds = ext_module.nms(*indata_list, **indata_dict)
     else:
-        inds = NMSop.apply(boxes, scores, iou_threshold, offset)
+        if max_num < 0:
```
RunningLeon, line 164: we could pass `max_num` to nms even though max_num <= 0.

mmcv/ops/nms.py
SemyonBevzuk: Removed `torch.onnx.is_in_onnx_export()` from `forward` and added a workaround with `max_num` in `batchd_nms`:
https://github.com/open-mmlab/mmcv/blob/5c083b88e3c312e34a707cface506126036a390a/mmcv/ops/nms.py#L310-L323

This should solve the problem with the loss of `bboxes` during training.

SemyonBevzuk: In addition, we can now remove `dets = dets [: max_num]` after calling `batch_nms`, as, for example, in `mmdetection::multiclass_nms`:
https://github.com/open-mmlab/mmdetection/blob/2313bd76940a6059b7e8bad4908c4558b3160f75/mmdet/core/post_processing/bbox_nms.py#L84-L88

mmcv/ops/nms.py
SemyonBevzuk: If we pass `max_num < 0` to the `NMSop.forward` function, then during export this value will be written to the ONNX graph.
This is not good, so we need to prepare the correct arguments before calling this function.

RunningLeon: if `max_num < 0` , then we could set `max_output_per_class` to `sys.maxsize` as original style:
```
            max_output_per_class = g.op(
                'Constant',
                value_t=torch.tensor([sys.maxsize], dtype=torch.long))
```

SemyonBevzuk: It seems to me that using `sys.maxsize` significantly overestimates the value of the `max_output_per_class` argument.

The number of indices after the NMS operation can be limited by the size of `bboxes`.
Then, if `max_num` is not specified or `max_num <= 0`, we can calculate its value in the ONNX graph using `bbox` shape.

RunningLeon: Agree, `sys.maxsize` is indeed too large.

RunningLeon: However, there's another issue. `max_output_per_class` should be constant for TensorRT backend. Using `bbox.shape` as `max_num` would create dynamic `max_output_per_class`.

mmcv/ops/nms.py
RunningLeon: It should be feasible for now.

mmcv/ops/nms.py
```diff
@@ -37,14 +50,26 @@ def symbolic(g, bboxes, scores, iou_threshold, offset):
             from torch.onnx.symbolic_opset9 import select, squeeze, unsqueeze
             boxes = unsqueeze(g, bboxes, 0)
             scores = unsqueeze(g, unsqueeze(g, scores, 0), 0)
-            max_output_per_class = g.op(
-                'Constant',
-                value_t=torch.tensor([sys.maxsize], dtype=torch.long))
+
+            from torch.onnx.symbolic_helper import _is_value, _size_helper
+            if not _is_value(max_num):
```
RunningLeon, line 55: Here `max_num` should be a constant int value. We could remove `if` block and only assert `max_num > 0` here.
```
assert max_num > 0, 'max_num should be greater than 0 while in exploring to ONNX.'
```

RunningLeon, line 55: Could you update the test unit of `nms` and `batched _nms` in `test_onnx.py` and `test_tensorrt.py` if possible.

mmcv/ops/nms.py
SemyonBevzuk: So, if this limitation only applies to TensorRT, then perhaps it should only be considered when exporting to TensorRT?
For example, we can replace `bbox.shape` with a constant value somewhere in this function:
https://github.com/open-mmlab/mmcv/blob/5d75c7a8ee4a0710208a86c75bc654551023b0f3/mmcv/tensorrt/tensorrt_utils.py#L7

mmcv/ops/nms.py
SemyonBevzuk: It seems that after the changes in `tensorrt_utils.preprocess_onnx` before exporting to TensorRT, we can save the ability to define `max_num` not only as a constant int value.

mmcv/ops/nms.py
RunningLeon: Normally,  `max_num` should be required to set when exporting to ONNX for deployment.

mmcv/ops/nms.py
RunningLeon: It's not necessary to set `max_num` to be dynamic for deployment. Memory allocation for tensorrt nms plugin highly depends on this parameter and it shouldn't be dynamic.

mmcv/tensorrt/tensorrt_utils.py
```diff
@@ -65,10 +84,11 @@ def parse_data(name, typ):
 
             if len(node_inputs) >= 3:
                 max_output_boxes_per_class = parse_data(
-                    node_inputs[2], np.int64)
+                    node_inputs[2], np.int64, max_output_boxes_per_class)
```
RunningLeon, line 87: Here `max_output_boxes_per_class` should become static. But if `max_num` is set by `bbox.shape`, then it may be too large for TensorRT to allocate memory for NMS plugin.

mmcv/ops/nms.py
SemyonBevzuk: If there is one NMS operation in the model, then the `max_num` value in the config will be specified for it.
But for example in Faster RCNN, there are two NMS operations.
For the last operation (from BBoxHead) the `max_num` value is specified as a constant.
For NMS operation from RPNHead, when executed in PyTorch, the `max_num` value is not specified, that is, all indexes are used in further processing.
It would be correct to keep this behavior in ONNX as well.

Dynamically defining `max_num` through the `bbox.shape` in the ONNX graph allows us to use the reshaping of the model.
It is possible that when the input resolution is changed, the stored constant value of `max_num` in NMS from RPNHead will be incorrect.

mmcv/tensorrt/tensorrt_utils.py
SemyonBevzuk: If `max_output_boxes_per_class` is not constant, then the default value (1000000) will be inserted instead, and unnecessary nodes will be removed from the ONNX graph before exporting to TensorRT. The default value of the constant can be decreased.

mmcv/ops/nms.py
RunningLeon: Then what benefits it would bring if you keep `max_num` as dynamic for backend other than TensorRT?

SemyonBevzuk: For example, this can be useful in the OpenVINO backend.
Support for the ability of `max_num` to be non-constant allows for model reshaping and improves performance.

RunningLeon:
@SemyonBevzuk Coud you fix lint error?

mmcv/ops/nms.py
RunningLeon: Then if you force to set `max_num`to a constant value and this value is much smaller than the `bbox.shape[0]`, won't it improve the performance much more? `max_num` is normally set to 100 in the config file.

mmcv/ops/nms.py
RunningLeon: `max_num` for RPNHead is set in the [config](https://github.com/open-mmlab/mmdetection/blob/2f2cb37cdaeac0862710045a066ee1669e286a42/configs/_base_/models/faster_rcnn_r50_fpn.py#L99).

mmcv/ops/nms.py
SemyonBevzuk: But by choosing too small a value for `max_num`, we can reduce the quality of the model.

Previously, calling `batched_nms` from RPNHead did not pass `max_num` and this caused some performance problems. To solve them, it was proposed to define `max_num` using `bbox.shape`.

mmcv/ops/nms.py
SemyonBevzuk: Yes, I didn't notice that filtering by `max_per_img` is applied after calling `batched_nms`. Then nms was called without limitation on `max_output_boxes_per_class` and the `sys.maxsize` value was stored in the ONNX graph.

I will add the `max_per_img` value to `cfg.nms`.

SemyonBevzuk:
> @SemyonBevzuk Could you fix the lint error?

Yes, I'll take a look.

tests/test_ops/test_onnx.py
```diff
@@ -93,9 +93,11 @@ def test_nms():
     np_scores = np.array([0.6, 0.9, 0.7, 0.2], dtype=np.float32)
     boxes = torch.from_numpy(np_boxes)
     scores = torch.from_numpy(np_scores)
-    pytorch_dets, _ = nms(boxes, scores, iou_threshold=0.3, offset=0)
+    pytorch_dets, _ = nms(
+        boxes, scores, iou_threshold=0.3, offset=0, score_threshold=0)
     pytorch_score = pytorch_dets[:, 4]
-    nms = partial(nms, iou_threshold=0.3, offset=0)
+    nms = partial(
+        nms, iou_threshold=0.3, offset=0, score_threshold=0, max_num=100)
```
RunningLeon, line 100: There are only 4 bboxes, and `max_num` set to 100 is unmeaningful. Could set to `0` as in line 97.

RunningLeon:
@SemyonBevzuk , Could you fix CI error?

RunningLeon:
@ZwwWayne See if you have any comments on this PR.

tests/test_tensorrt_preprocess.py
```diff
@@ -0,0 +1,74 @@
+import os
```
RunningLeon, line 1: Could we move this test file to the `test_ops` directory for better organization?

tests/test_tensorrt_preprocess.py
```diff
@@ -0,0 +1,74 @@
...
+from functools import wraps
+
+import onnx
+import torch
+
+from mmcv.ops import nms
+from mmcv.tensorrt.tensorrt_utils import preprocess_onnx
```
RunningLeon, line 8: Because github CI does not have TensorRT and this `import operation` would fail. We could move `preproces_onnx` to a new file `preprocess.py` and import the as `from mmcv.tensorrt.preprocess import preprocess_onnx` to avoid installing   TensorrRT for the running of test units of `preprocess_onnx`.

tests/test_tensorrt_preprocess.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def remove_tmp_file(func):
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        onnx_file = 'tmp.onnx'
+        kwargs['onnx_file'] = onnx_file
+        result = func(*args, **kwargs)
```
RunningLeon, line 17: If the running of `func` fails, then `tmp.onnx` may not be deleted successfully. Could wrap it `try, except, finally`.

tests/test_tensorrt_preprocess.py
```diff
@@ -0,0 +1,74 @@
...
+        output_names=['output'])
+
+    onnx_model = onnx.load(onnx_file)
+    onnx.checker.check_model(onnx_model)
+    return onnx_model
+
+
+def test_can_handle_nms_with_constant_maxnum():
+
+    class NMS_with_const_maxnum(torch.nn.Module):
```
RunningLeon, line 48: The name of this class may be too long.

tests/test_tensorrt_preprocess.py
```diff
@@ -0,0 +1,74 @@
...
+    onnx_model = export_nms_module_to_onnx(NMS_with_const_maxnum)
+    preprocess_onnx_model = preprocess_onnx(onnx_model)
+    for node in preprocess_onnx_model.graph.node:
+        if 'NonMaxSuppression' in node.name:
+            assert len(node.attribute) == 5, 'The NMS must have 5 attributes.'
+
+
+def test_can_handle_nms_with_undefined_maxnum():
+
+    class NMS_with_const_maxnum(torch.nn.Module):
```
RunningLeon, line 62: The name of this class may be too long and should follow CamelCase style.

SemyonBevzuk: Done.

tests/test_tensorrt_preprocess.py
SemyonBevzuk: Done.

tests/test_tensorrt_preprocess.py
SemyonBevzuk: Done.

tests/test_tensorrt_preprocess.py
SemyonBevzuk: Done.

tests/test_tensorrt_preprocess.py
SemyonBevzuk: Done.

RunningLeon:
@SemyonBevzuk hi，could you kindly fix the CI error?

tests/test_ops/test_tensorrt_preprocess.py
```diff
@@ -41,7 +41,6 @@ def export_nms_module_to_onnx(module, onnx_file):
         output_names=['output'])
 
     onnx_model = onnx.load(onnx_file)
-    onnx.checker.check_model(onnx_model)
```
RunningLeon: Maybe it is not necessary to remove this checker. Could skip torch==1.3.1 in the unittest.

SemyonBevzuk: Added version check before calling `onnx.checker.check_model`.

Or did you mean changes to `build.yml`?
Example:
```
build_cpu:
    ...
      - name: Run unittests and generate coverage report (torch == 1.3.1)
        if: ${{ matrix.torch == 1.3.1 }}
        run: |
          pip install -r requirements/test.txt
          coverage run --branch --source=mmcv -m pytest tests/ --ignore test_ops/test_tensorrt_preprocess.py
          coverage xml
          coverage report -m
      - name: Run unittests and generate coverage report (torch > 1.3.1)
        if: ${{ matrix.torch > 1.3.1 }}
        run: |
          pip install -r requirements/test.txt
          coverage run --branch --source=mmcv -m pytest tests/
          coverage xml
          coverage report -m
```

RunningLeon (approved):
LGTM

RunningLeon:
Kindly ping @ZwwWayne

ZwwWayne (approved):

PR #958, opened 2021-04-16
fix init cfg

jshilong:
## Motivation
I think the existence of attributes in an instance  should not depend on the value of the arguments. So I change the all `hasattr(self, 'init_cfg')` to `self.init_cfg` in BaseModule.
## BC-breaking
None


jshilong:
Please @hhaAndroid @ZwwWayne check whether it affects the existing changes in the downstream codebase.

MeowZheng (approved):
LGTM

PR #972, opened 2021-04-20
support for CPU training

wangruohui:
I am not quite sure whether this is a bug in historical codes. Because it seems currently no package support training with CPU so the modified line will never enter.

### DIFF
--- a/mmcv/parallel/data_parallel.py
+++ b/mmcv/parallel/data_parallel.py
@@ -71,7 +71,7 @@ class MMDataParallel(DataParallel):
             # We add the following line thus the module could gather and
             # convert data containers as those in GPU inference
             inputs, kwargs = self.scatter(inputs, kwargs, [-1])
-            return self.module.val_step(*inputs, **kwargs)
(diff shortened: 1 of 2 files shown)


yptheangel:
do you support CPU inference?

wangruohui:
> do you support CPU inference?

It has been supported in master. And this PR does not break this functionality.

wangruohui:
> do you support CPU inference?

It has been supported in master. And this PR does not break this functionality.

zhouzaida (approved):
LGTM

mmcv/parallel/data_parallel.py
```diff
@@ -49,7 +49,7 @@ def train_step(self, *inputs, **kwargs):
             # We add the following line thus the module could gather and
             # convert data containers as those in GPU inference
             inputs, kwargs = self.scatter(inputs, kwargs, [-1])
-            return self.module.train_step(*inputs, **kwargs)
+            return self.module.train_step(*inputs[0], **kwargs[0])
```
ZwwWayne, line 52: Shall we also modify that in val_step?

PR #983, opened 2021-04-24
[Fix] Read in rb mode then decode

wangruohui:
On Windows :

```python
with open("some_file", r) as f:
    f.read()
```

will decode the content of the file based on windows's system language setting. 

Using `rb` mode then decode within python will force using utf-8.

note: the ruby precommit hook does not work on windows and I will recover it when everything other is good before actual merge.

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -91,7 +91,7 @@ class Config:
     @staticmethod
     def _validate_py_syntax(filename):
         with open(filename, 'r', encoding='utf-8') as f:
-            # Setting encoding explicitly help resolve some coding issue on windows
+            # Setting encoding explicitly to resolve coding issue on windows
             content = f.read()
         try:
             ast.parse(content)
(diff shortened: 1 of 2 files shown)


mmcv/utils/config.py
```diff
@@ -90,8 +90,9 @@ class Config:
 
     @staticmethod
     def _validate_py_syntax(filename):
-        with open(filename, 'r') as f:
-            content = f.read()
+        with open(filename, 'rb') as f:
```
zhouzaida, line 93:
```python
with open(filename, 'r', encoding='utf-8') as f:
```
Setting encoding will be more explicit.

ZwwWayne:
Kindly ping @wangruohui

ZwwWayne:
@wangruohui ，lint failed. The comments are too long and need to be separated in two lines.

ZwwWayne (approved):

PR #987, opened 2021-04-24
[Fix] Use len(result) as batch size

### LINKED ISSUE #985: Progress bar does not forward correctly
Problem should lies at the following codes. 

https://github.com/open-mmlab/mmcv/blob/5a99f587ed09797ae6720d33888c753ad88f178a/mmcv/engine/test.py#L35-L38

When running with classification, the first values (image_meta) is a data container with nested list, like `DataContainer([[{meta1}, {meta2}, ... ]])`, thus `batch_size` = 1 and is not desired. 

I suggest using the length of the result as batch size, would it be good to all packages?

wangruohui:
fix #985

### DIFF
--- a/mmcv/engine/test.py
+++ b/mmcv/engine/test.py
@@ -32,6 +32,7 @@ def single_gpu_test(model, data_loader):
             result = model(return_loss=False, **data)
         results.extend(result)
 
+        # Assume result has the same length of batch_size, refer to https://github.com/open-mmlab/mmcv/issues/988
         batch_size = len(result)
         for _ in range(batch_size):
             prog_bar.update()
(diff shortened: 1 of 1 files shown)


mmcv/engine/test.py
```diff
@@ -33,7 +33,7 @@ def single_gpu_test(model, data_loader):
         results.extend(result)
 
         # use the first key as main key to calculate the batch size
```
dreamerlin, line 35: remove this comments

zhouzaida:
@dreamerlin

dreamerlin (approved):

mmcv/engine/test.py
```diff
@@ -32,8 +32,7 @@ def single_gpu_test(model, data_loader):
             result = model(return_loss=False, **data)
         results.extend(result)
 
-        # use the first key as main key to calculate the batch size
-        batch_size = len(next(iter(data.values())))
```
ZwwWayne, line 36: Add comment to indicate that: we assume the results has similar length as batch size, and refer to this PR and issue.

ZwwWayne (approved):

PR #995, opened 2021-04-26
[Feature] Support clipping min_lr in StepLrUpdaterHook to help MMDet3D

### LINKED ISSUE #994: [Feature] Support min_lr clipping in lr_scheduler (MMDet3D related problem)
**Hi, I am Ziyi, a developer of MMDet3D. Can we support learning rate clipping in lr_scheduler? For example, when I use StepLR which decays by 0.7 every 10 epochs with a `init_lr==1e-3`, I would like to keep my `lr==1e-5` after it decays to some values lower than `1e-5`. Such min_lr clipping is common in several 3D point cloud papers.**

**Motivation**
For example, in the PointNet++ official code, they decay lr by 0.7 every 10 epochs and clip a minium lr of 1e-5 as shown [here](https://github.com/charlesq34/pointnet2/blob/master/scannet/train.py#L85). This is widely adopted in other point cloud related papers like PointNet (see [here](https://github.com/charlesq34/pointnet/blob/master/train.py#L80)), DGCNN, etc.

**Related resources**
I think it's quite easy to implement. We can add a `min_lr` args to `LrUpdaterHook`, and use a `max(lr, self.min_lr)` to determine the return value in `get_lr`? However, I am not sure whether it will cause other questions because I am not very familiar with MMCV.

**Additional context**
Is there any existing way to achieve my feature request? If you not yet and you think my solution is okay, I can create a PR to support it. Thanks!

Wuziyi616:
Fix #994. Please refer to that issue for detailed motivation. This feature will be used in MMDet3D. Thanks!

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -165,7 +165,7 @@ class FixedLrUpdaterHook(LrUpdaterHook):
 @HOOKS.register_module()
 class StepLrUpdaterHook(LrUpdaterHook):
 
-    def __init__(self, step, gamma=0.1, **kwargs):
+    def __init__(self, step, gamma=0.1, min_lr=None, **kwargs):
         assert isinstance(step, (list, int))
         if isinstance(step, list):
             for s in step:
@@ -176,6 +176,7 @@ class StepLrUpdaterHook(LrUpdaterHook):
             raise TypeError('"step" must be a list or integer')
         self.step = step
         self.gamma = gamma
+        self.min_lr = min_lr
         super(StepLrUpdaterHook, self).__init__(**kwargs)
 
     def get_lr(self, runner, base_lr):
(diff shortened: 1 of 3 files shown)


zhouzaida (approved):
LGTM. If passing CI, it could be merged.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -165,7 +165,7 @@ def get_lr(self, runner, base_lr):
 @HOOKS.register_module()
 class StepLrUpdaterHook(LrUpdaterHook):
 
-    def __init__(self, step, gamma=0.1, **kwargs):
+    def __init__(self, step, gamma=0.1, min_lr=None, **kwargs):
```
ZwwWayne, line 168: Unit tests should be added to verify this feature.

ZwwWayne, line 168: Suggest adding docstring for this module accordingly.

Wuziyi616: Done.

Wuziyi616: Yes.

Wuziyi616:
Seems that I can't run the workflow? May need @ZwwWayne to approve it.

ZwwWayne:
> Seems that I can't run the workflow? May need @ZwwWayne to approve it.

lint failed.

Wuziyi616:
The linting check error said that "mmcv/engine/test.py:35:80: E501 line too long (113 > 79 characters)", which is absolutely wrong because I didn't even modify that file. I think this PR is ready for merge @ZwwWayne?

ZwwWayne (approved):

PR #997, opened 2021-04-27
Support corner_pool related custom operators for onnxruntime in mmcv

v-qjqs:
Hi, this PR supports to export corner_pool related custom operators (`TopPool`,`BottomPool`,`LeftPool`,`RightPool` ) to onnx, and run it with onnxruntime.
It's working in process during my free time.

### DIFF
--- a/docs/onnxruntime_custom_ops.md
+++ b/docs/onnxruntime_custom_ops.md
@@ -27,6 +27,12 @@
     - [Inputs](#inputs-3)
     - [Outputs](#outputs-3)
     - [Type Constraints](#type-constraints-3)
+  - [CornerPool](#corner_pool)
+    - [Description](#description-4)
+    - [Parameters](#parameters-4)
+    - [Inputs](#inputs-4)
+    - [Outputs](#outputs-4)
+    - [Type Constraints](#type-constraints-4)
 
 <!-- TOC -->
 
@@ -171,3 +177,34 @@ Perform sample from `input` with pixel locations from `grid`.
 ### Type Constraints
 
 - T:tensor(float32, Linear)
+
+## CornerPool
+
+### Description
+
+Perform CornerPool on `input` features. Read [CornerNet -- Detecting Objects as Paired Keypoints](https://arxiv.org/abs/1808.01244) for more details.
+
+### Parameters
+
+| Type    | Parameter       | Description                                                      |
+| ------- | --------------- | ---------------------------------------------------------------- |
+| `int`   | `mode`          | corner pool mode, (0: `top`, 1: `bottom`, 2: `left`, 3: `right`) |
+
+### Inputs
+
+<dl>
+<dt><tt>input</tt>: T</dt>
+<dd>Input features. 4-D tensor of shape (N, C, H, W). N is the batch size.</dd>
+</dl>
+
+### Outputs
+
+<dl>
+<dt><tt>output</tt>: tensor(int64)</dt>
+<dd>Output the pooled features. 4-D tensor of shape (N, C, H, W).</dd>
+</dl>
+
+### Type Constraints
+
+- T:tensor(float32)
+
--- a/docs/onnxruntime_custom_ops.md
+++ b/docs/onnxruntime_custom_ops.md
@@ -200,7 +200,7 @@ Perform CornerPool on `input` features. Read [CornerNet -- Detecting Objects as 
 ### Outputs
 
 <dl>
-<dt><tt>output</tt>: tensor(int64)</dt>
+<dt><tt>output</tt>: T</dt>
 <dd>Output the pooled features. 4-D tensor of shape (N, C, H, W).</dd>
 </dl>
 
--- a/docs/onnxruntime_custom_ops.md
+++ b/docs/onnxruntime_custom_ops.md
@@ -207,4 +207,3 @@ Perform CornerPool on `input` features. Read [CornerNet -- Detecting Objects as 
 ### Type Constraints
 
 - T:tensor(float32)
-
--- a/docs/onnxruntime_custom_ops.md
+++ b/docs/onnxruntime_custom_ops.md
@@ -27,7 +27,7 @@
     - [Inputs](#inputs-3)
     - [Outputs](#outputs-3)
     - [Type Constraints](#type-constraints-3)
-  - [CornerPool](#corner_pool)
+  - [CornerPool](#cornerpool)
     - [Description](#description-4)
     - [Parameters](#parameters-4)
     - [Inputs](#inputs-4)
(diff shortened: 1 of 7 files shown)


v-qjqs:
Hi, any reviewers?

grimoire:
Thanks for the contribution!
Would it be better to add a `cummax` custom op instead of `CornerPool` op? It is more general and can support torch>1.5.0.

v-qjqs:
> Thanks for the contribution!
> Would it be better to add a `cummax` custom op instead of `CornerPool` op? It is more general and can support torch>1.5.0.
Hi, I tested the implemented `mmcv::CornerPool` and it works with torch1.5.0. I wonder why [this line](https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/corner_pool.py#L116) is needed? Is is ok for just using `self.corner_pool.apply(x)`.

grimoire:
PyTorch built-in ops support more platforms and might get better support from different deploy backends in the future. 
As we do plan to add more backends support. It is a good idea to avoid custom ops unless we have to do.

mmcv/ops/csrc/onnxruntime/cpu/corner_pool.cpp
```diff
@@ -0,0 +1,157 @@
...
+                // do top_pool
+                for (int ind = 1; ind < height; ind <<= 1) {
+                    for (int h = 0; h < height - ind; h++) {
+                        output[index_n_c + h * width + w] = std::max(tmp_output[h], tmp_output[h+ind]);
+                    }
+                    // copy column from updated output to tmp_output
+                    for (int h = 0; h < height - ind; h++) {
+                        tmp_output[h] = output[index_n_c + h * width + w];
+                    }
+                } // for ind
```
grimoire, line 27: As the corner pool is basically a `Scan` operation on a given dimension. This code seems too complex.
Could you provide a brief description of the logic behind the code?

grimoire (commented):
Hi, could you please add document about your ops in [onnxruntime_custom_ops](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_custom_ops.md) and [onnxruntime_op](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md)?

mmcv/ops/csrc/onnxruntime/cpu/corner_pool.cpp
```diff
@@ -0,0 +1,157 @@
...
+    int batch_size = nthreads / channels / width / height;
+    for (int n = 0; n < batch_size; n++) {
+        int index_n = n * channels * width * height;
+        for (int c = 0; c < channels; c++) {
+            int index_n_c = index_n + c * width * height;
+            for (int w = 0; w < width; w++) {
+                // copy column from output to tmp_output
+                for (int h = 0; h < height; h++) {
+                    int index = index_n_c + h * width + w;
+                    tmp_output[h] = output[index];
```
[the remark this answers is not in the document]
v-qjqs: No, `output` here is not replaceable with `input`, since the `output` is updated while `input` is not changed.

mmcv/ops/csrc/onnxruntime/cpu/corner_pool.cpp
v-qjqs: Hi, actually the implementation is modified from [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/csrc/pytorch/corner_pool.cpp).

grimoire: This looks like some kind of parallel implementation of scan (like Kogge-Stone or Hillis-Steele?). I do not think it is the best choice to implement this ops on the CPU.

Code below can do the very same as `CornerPool('right')` without memcpy, additional temp workspace, and it is much simple.
```python
def cp_right(data): 
     batch,channel,height,width = data.shape 
     output = torch.empty_like(data) 
     for n in range(batch): 
         for c in range(channel): 
            for h in range(height): 
                 output[n,c,h,0] = data[n,c,h,0] 
                 for w in range(1, width): 
                     output[n,c,h,w] = max(output[n,c,h,w], data[n,c,h,w-1]) 
     return output 
```

v-qjqs:
> Thanks for the contribution!
> Would it be better to add a `cummax` custom op instead of `CornerPool` op? It is more general and can support torch>1.5.0.

The implementation of `mmcv::CornerPool` helps to those who use the forward function `self.corner_pool.apply(x)`.
Also I'm planing to support a custom `cummax` op  in my next PR, following the official pytorch implementation.

mmcv/ops/csrc/onnxruntime/cpu/corner_pool.cpp
v-qjqs: Thanks. Another version with no memcpy will be committed soon.

docs/onnxruntime_custom_ops.md
```diff
@@ -27,6 +27,12 @@
     - [Inputs](#inputs-3)
     - [Outputs](#outputs-3)
     - [Type Constraints](#type-constraints-3)
+  - [CornerPool](#corner_pool)
```
grimoire, line 30: Please fix this link.

docs/onnxruntime_op.md
```diff
@@ -21,6 +21,7 @@
 |     [RoIAlign](onnxruntime_custom_ops.md#roialign)     |   Y   |   N   |     1.2.5     |
 |          [NMS](onnxruntime_custom_ops.md#nms)          |   Y   |   N   |     1.2.7     |
 | [grid_sampler](onnxruntime_custom_ops.md#grid_sampler) |   Y   |   N   |    master     |
+| [CornerPool](onnxruntime_custom_ops.md#corner_pool)    |   Y   |   N   |    master     |
```
grimoire, line 24: Please fix this link.

mmcv/ops/csrc/onnxruntime/cpu/corner_pool.cpp
```diff
@@ -0,0 +1,115 @@
...
+    OrtTensorDimensions out_dimensions(ort_, input);
+    OrtValue *output = ort_.KernelContext_GetOutput(context, 0, out_dimensions.data(), out_dimensions.size());
+    T *output_data = ort_.GetTensorMutableData<T>(output);
+
+    // get output_size
+    int batch_size = out_dimensions.data()[0];
+    int input_channels = out_dimensions.data()[1];
+    int input_height = out_dimensions.data()[2];
+    int input_width = out_dimensions.data()[3];
+    int output_size = batch_size * input_channels * input_height * input_width;
```
grimoire, line 104: As long as output_size is used to compute batch_size inside forward function, could we just pass batch_size to the forward?

grimoire (approved):
LGTM!

grimoire (approved):

grimoire:
@ZwwWayne any comments?

ZwwWayne (approved):

PR #1007, opened 2021-04-30
Add test util for checking stand-alone python scripts

innerlee:
This function can be used to check stand-alone scripts, without tweaking `sys.path` manually. Moreover, it will increase test coverage for codes the script touches.

### DIFF
--- a/tests/data/scripts/hello.py
+++ b/tests/data/scripts/hello.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+
+import argparse
+import warnings
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Compute the perfect recall.')
+    parser.add_argument('name', help='Path to sdk json file.')
+
+    args = parser.parse_args()
+
+    return args
+
+
+def main():
+    args = parse_args()
+    print(f'hello {args.name}!')
+    if args.name == 'lizz':
+        warnings.warn('I have a secret!')
+
+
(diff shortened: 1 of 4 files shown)


tests/test_utils/test_testing.py
```diff
@@ -180,3 +180,14 @@ def test_assert_params_all_zeros():
...
+
+
+def test_check_python_script(capsys):
+    mmcv.utils.check_python_script('./tests/data/scripts/hello.py zz')
+    captured = capsys.readouterr().out
+    assert captured == 'hello zz!\n'
+    mmcv.utils.check_python_script('./tests/data/scripts/hello.py lizz')
+    captured = capsys.readouterr().out
+    assert captured == 'hello lizz!\n'
+    with pytest.raises(SystemExit):
```
ZwwWayne, line 192: Add comments to indicate what error is it testing.

tests/data/scripts/hello.py
```diff
@@ -0,0 +1,24 @@
...
+
+    args = parser.parse_args()
+
+    return args
+
+
+def main():
+    args = parse_args()
+    print(f'hello {args.name}!')
+    if args.name == 'lizz':
```
ZwwWayne, line 19: Is it better to use a more general name rather than using your ID? Though I am OK with that.

innerlee: haha

ZwwWayne (approved):

PR #1010, opened 2021-05-01
Support custom operators cummax and cummin for onnxruntime

v-qjqs:
Hi, this PR tries to implement custom operator `mmcv::cummax` and `mmcv::cummin`, which supports to export `torch.cummax` and `torch.cummin` to onnx format, and run it with onnxruntime.
`mmcv::cummax` is a more general operation and an extension to previous [`mmcv::CornerPool`](https://github.com/open-mmlab/mmcv/pull/997), since the former supports arbitrary tensor shapes while the latter only supports for 4-D input tensor.
Based on the customized `mmcv::cummax` and `mmcv::CornerPool` operation, CornerNet in MMDet might be exportable to onnx for various pytorch version.

### DIFF
--- a/mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
+++ b/mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
@@ -1,10 +1,9 @@
-#include "reduce_ops.h"
-
 #include <assert.h>
 
 #include <vector>
 
 #include "../ort_mmcv_utils.h"
+#include "reduce_ops.h"
 
 // modified from
 // https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ReduceOps.cpp
@@ -47,9 +46,12 @@ void cummax_cummin_helper(const T1 *input, T1 *output, T2 *indices,
 }
 
 // modified `tensor_dim_apply3` from
-// https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/TensorDimApply.h
-// currently `tensor_dim_apply3` is only used for `cummax` and `cummin`,
-// according to the official pytorch source codes.
+// https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/TensorDimApply.h.
+// the difference is that: (1) use `reversed_dim_cumprod` for fast computing of
+// tensor `size` and `stride`. (2) the same `stride` is used for input, output,
+// and indices, since it's unnecessary to use separate values. currently
+// `tensor_dim_apply3` is only used for `cummax` and `cummin`, according to the
+// official pytorch projects: https://github.com/pytorch/pytorch.
 template <typename T1, typename T2, typename Function>
 void tensor_dim_apply3(const T1 *input, T1 *output, T2 *indices,
                        const int64_t dim, const int64_t ndims,
--- a/mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
+++ b/mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
@@ -1,0 +1,142 @@
+#include "reduce_ops.h"
+
+#include <assert.h>
+
+#include <vector>
+
+#include "../ort_mmcv_utils.h"
+
+// modified from
+// https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ReduceOps.cpp
+
+static inline int64_t maybe_wrap_dim(int64_t dim, int64_t ndims) {
+  int64_t min = -ndims;
+  int64_t max = ndims - 1;
+  assert(dim >= min && dim <= max);
+  if (dim < 0) dim += ndims;
+  return dim;
+}
+
+static inline int64_t get_dim_stride(const int64_t dim, const int64_t ndims,
+                                     const int64_t *reversed_dim_cumprod) {
+  return dim == ndims - 1 ? 1 : reversed_dim_cumprod[dim + 1];
+}
+
+static inline int64_t get_dim_size(const int64_t dim, const int64_t ndims,
+                                   const int64_t *reversed_dim_cumprod) {
+  return dim == ndims - 1
+             ? reversed_dim_cumprod[dim]
+             : reversed_dim_cumprod[dim] / reversed_dim_cumprod[dim + 1];
+}
+
+template <typename T1, typename T2, typename Operation>
+void cummax_cummin_helper(const T1 *input, T1 *output, T2 *indices,
+                          const int64_t input_dim_size, const int64_t stride) {
+  Operation op;
+  T1 out = input[0];
+  int64_t idx = 0;
+  for (int64_t i = 0; i < input_dim_size; i++) {
+    T1 curr_elem = input[i * stride];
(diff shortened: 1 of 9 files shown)


mmcv/ops/nms.py
```diff
@@ -218,6 +218,15 @@ def soft_nms(boxes,
                                      float(iou_threshold), float(sigma),
                                      float(min_score), method_dict[method],
                                      int(offset))
+
+    if torch.onnx.is_in_onnx_export():
```
grimoire, line 222: It seems that size(0) would be treated as a tensor. There is no need to add this branch.
tmp onnx

tests/test_ops/test_onnx.py
```diff
@@ -494,3 +494,79 @@ def corner_pool_func(input):
...
+    # Note generally `cummax` or `cummin` is exportable to ONNX
+    # as long as the pytorch version >= 1.5.0, since `torch.cummax`
+    # is only supported with torch >= 1.5.0.
+    # But when `cummax` or `cummin` serves as an intermediate component
+    # whose outputs is used as inputs for another modules, it's expected
+    # that pytorch version must be >= 1.7.0. Otherwise error appears like:
+    # `RuntimeError: tuple  appears in op that does not forward tuples,
+    # unsupported 'kind: prim::PythonOp`.
+    if version.parse(torch.__version__) < version.parse('1.7.0'):
+        warnings.warn('test_cummax_cummin should be ran with pytorch >= 1.7.0')
```
grimoire, line 513: Please use skip instead of return.

tests/test_ops/test_onnx.py
```diff
@@ -494,3 +494,79 @@ def corner_pool_func(input):
...
+        # arbitrary shape, e.g. 1-D, 2-D, 3-D, ...
+        torch.rand((2, 3, 4, 1, 5)),
+        torch.rand((1)),
+        torch.rand((2, 0, 1)),  # tensor.numel() is 0
+        torch.FloatTensor(),  # empty tensor
+    ]
+
+    cummax_cummin_funcs = {'cummax': torch.cummax, 'cummin': torch.cummin}
+
+    for i, input in enumerate(input_list):
```
grimoire, line 535: Please remove enumerate if `i` is not used.

mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  // cummax is only applied to input which is non-zero dim and non-empty
+  if (numel) {
+    // compute the cumulative production on dimension size,
+    // which is then used for computing the stride or size of a specific `dim`.
+    reversed_dim_cumprod[ndims - 1] = out_dimensions.data()[ndims - 1];
+    for (int64_t dim_i = ndims - 2; dim_i >= 0; dim_i--) {
+      reversed_dim_cumprod[dim_i] =
+          reversed_dim_cumprod[dim_i + 1] * out_dimensions.data()[dim_i];
+    }
```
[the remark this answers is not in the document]
v-qjqs: Hi, here the tmp memory is used for saving time to compute the `stride`, which is described in `get_dim_stride`.
Directly using `out_dimensions.data()` to compute `stride` is ok while need extra time cost. If the `ndim` is small this might be ignored?

mmcv/ops/nms.py
v-qjqs:
```python
a = torch.rand((3,4))
b=a.size(0)
c=torch._shape_as_tensor(a)[0]
type(b)  # <class 'int'>
type(c) # <class 'torch.Tensor'> with ndim 0

```

mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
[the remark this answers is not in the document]
grimoire: I means put stride in `reversed_dim_cumprod` and remove `get_dim_stride` and `get_dim_size`.

mmcv/ops/nms.py
grimoire: `Tensor.size()` and `Tensor.shape` would be treated as [Shape](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Shape) op in ONNX even when their type is `int` in python.
You can visualize the generated ONNX at https://netron.app/.

mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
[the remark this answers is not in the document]
v-qjqs: Hi, did you mean that use `out_dimensions.data()` for size and `reversed_dim_cumprod` for stride?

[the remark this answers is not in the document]
grimoire: Yes, I think that would be better.
And, this is just a recommendation. You can still keep your code unchanged.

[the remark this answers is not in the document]
v-qjqs: Ok, I may still keep with this implementation.

mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
```diff
@@ -0,0 +1,184 @@
...
+    output[i * stride] = out;
+    indices[i * stride] = idx;
+  }
+}
+
+// modified `tensor_dim_apply3` from
+// https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/TensorDimApply.h
+// currently `tensor_dim_apply3` is only used for `cummax` and `cummin`,
+// according to the official pytorch source codes.
```
ZwwWayne, line 52: 1. Can we also explain what is modified and why we need to modify them?
2. `according to the official pytorch source codes.` is a little bit vague here, does it mean the source code in `https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/TensorDimApply.h` or that in the CornerNet source code?

tests/test_ops/test_onnx.py
```diff
@@ -494,3 +494,78 @@ def corner_pool_func(input):
...
+def test_cummax_cummin(key, opset=11):
+    if torch.__version__ == 'parrots':
+        pytest.skip('onnx is not supported in parrots directly')
+
+    # Note generally `cummax` or `cummin` is exportable to ONNX
+    # as long as the pytorch version >= 1.5.0, since `torch.cummax`
+    # is only supported with torch >= 1.5.0.
+    # But when `cummax` or `cummin` serves as an intermediate component
+    # whose outputs is used as inputs for another modules, it's expected
+    # that pytorch version must be >= 1.7.0. Otherwise error appears like:
```
ZwwWayne, line 509: If so, we need to add the information about supported pytorch version in the documentation, e.g., in the section of cummax op

mmcv/ops/csrc/onnxruntime/cpu/reduce_ops.cpp
v-qjqs: Hi, the new comments is committed.

ZwwWayne (approved):

Issue #1008, opened 2021-04-30
Pass `None` to `cfg` in `build_from_cfg`


mzr1996:
**Describe the feature**
When we use `build_from_cfg`, I think `cfg` should be for users to customize configs and `default_args` is for developers to set default config.
However, now if we pass `None` to `cfg`, `build_from_cfg` will report an error.
https://github.com/open-mmlab/mmcv/blob/c4aa4146f50e60638b3455870c5d0ff1c2af10d5/mmcv/utils/registry.py#L19-L20
I think we should allow passing `None` to `cfg` if `build_from_cfg` is not None.

zhouzaida:
hi, @mzr1996
```
I think we should allow passing None to cfg if build_from_cfg is not None.
```
Could you tell me why it needs to be revised like this?

mzr1996:
Closed

PR #1012, opened 2021-05-08
Allow pass None to cfg in build_from_cfg

mzr1996:
## Motivation
Refer to https://github.com/open-mmlab/mmcv/issues/1008.
For example, when develop a backbone, we need to provide a default configuration for a layer and allow user to override these configuration by config file as well.
```python
def build_dropout(cfg, default_args=None):
    """Builder for drop out layers."""
    return build_from_cfg(cfg, DROPOUT_LAYERS, default_args)
```
However, in many situations, we don't need users to provide detail configs of one layer unless they want to custom it.
Now, we have to pass a dict cfg as default value.
```python
class ExampleLayer(nn.Module):

    def __init__(self, dropout_cfg=None):
        super().__init__()
        default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
        if dropout_cfg is not None:
            default_dropout_cfg.update(dropout_cfg)
        self.drop = build_dropout(default_dropout_cfg)
```
or set default value of cfg as an empty dict.
```python
class ExampleLayer(nn.Module):

    def __init__(self, dropout_cfg=None):
        super().__init__()
        default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
        if dropout_cfg is None:
            dropout_cfg = {}
        self.drop = build_dropout(dropout_cfg, default_dropout_cfg)
```

## Modification
By allowing pass None to cfg, we can use default_args param to set default config.
```python
class ExampleLayer(nn.Module):

    def __init__(self, dropout_cfg=None):
        super().__init__()
        default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
        self.drop = build_dropout(dropout_cfg, default_dropout_cfg)
```

## Checklist

✅ Pre-commit or other linting tools are used to fix the potential lint issues.
✅ The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.

### DIFF
--- a/mmcv/utils/registry.py
+++ b/mmcv/utils/registry.py
@@ -16,6 +16,9 @@ def build_from_cfg(cfg, registry, default_args=None):
     Returns:
         object: The constructed object.
     """
+    if cfg is None:
+        cfg = dict()
+
     if not isinstance(cfg, dict):
         raise TypeError(f'cfg must be a dict, but got {type(cfg)}')
     if 'type' not in cfg:
--- a/tests/test_utils/test_registry.py
+++ b/tests/test_utils/test_registry.py
@@ -235,6 +235,12 @@ def test_build_from_cfg():
     assert isinstance(model, ResNet)
     assert model.depth == 50 and model.stages == 4
 
+    cfg = None
+    model = mmcv.build_from_cfg(
+        cfg, BACKBONES, default_args=dict(type=ResNet, depth=50))
+    assert isinstance(model, ResNet)
+    assert model.depth == 50 and model.stages == 4
+
     # not a registry
     with pytest.raises(TypeError):
         cfg = dict(type='VGG')


ZwwWayne:
From my point of view, we can simply use `def __init__(self, dropout_cfg=dict(type='DropPath', drop_prob=0.))`, which is a more common case. But the modification is acceptable to me, so please @xvjiarui and @ycxioooong  have a look.

nbei:
*Just Suggestions.* We may be careful here. Since `None` will indicate removing(being without) such modules in some cases, we should be much more careful about this behavior and its influence.

mzr1996:
> _Just Suggestions._ We may be careful here. Since `None` will indicate removing(being without) such modules in some cases, we should be much more careful about this behavior and its influence.

I got that. But now, `None` indicates removing in some cases, using default config in some cases, it depends on upper-level interface definition. If it indicates removing, The developer need to explicitly indicate what to do if it's `None` like below:
```python
class ExampleLayer(nn.Module):

    def __init__(self, norm_cfg=dict(type='LN')):
        super().__init__()
        if norm_cfg is not None:
            self.norm = build_norm_layer(norm_cfg)
```

Therefore, after modification, the only situation developers need to avoid is assuming `build_norm_layer(cfg=None, default_args=dict(type='LN'))` returns an `Identity` layer.

innerlee:
Related https://github.com/open-mmlab/mmaction2/discussions/634

We need to take care of the semantic meaning of None.

There is a difference between "I don't care" and "Please use the default"

xvjiarui:
HI @mzr1996 

Thanks for the PR. This may be potentially BC breaking. In most use cases, if we set a module to `None`, we are not building this module. If we convert `None` into empty `dict()` for arg `cfg`, the module could be built successfully, but it may be confusing in the config file. 

Will this implementation do the trick? 
```python
class ExampleLayer(nn.Module):

    def __init__(self, dropout_cfg=dict()):
        super().__init__()
        default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
        self.drop = build_dropout(dropout_cfg, default_dropout_cfg)
```

mzr1996:
> HI @mzr1996
> 
> Thanks for the PR. This may be potentially BC breaking. In most use cases, if we set a module to `None`, we are not building this module. If we convert `None` into empty `dict()` for arg `cfg`, the module could be built successfully, but it may be confusing in the config file.
> 
> Will this implementation do the trick?
> 
> ```python
> class ExampleLayer(nn.Module):
> 
>     def __init__(self, dropout_cfg=dict()):
>         super().__init__()
>         default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
>         self.drop = build_dropout(dropout_cfg, default_dropout_cfg)
> ```

Well, that's exactly what I do now.
As for BC-breaking, in most uses, if `None` means `Don't build the model`, `build_from_cfg` won't be called because it doesn't accept `None`.
However, I think your worry is reasonable, we need to discriminate among `don't build`, `use default config`, and `override config`.
Here may be the best practice?
```python
class ExampleLayer1(nn.Module):

    def __init__(self, dropout_cfg=None):  # Don't build dropout_layer by default
        super().__init__()
        if dropout_cfg is not None:
            self.drop = build_dropout(dropout_cfg)

class ExampleLayer2(nn.Module):

    def __init__(self, dropout_cfg=dict()):  # Use default config unless get `None`
        super().__init__()
        if dropout_cfg is not None:
            default_dropout_cfg = dict(type='DropPath', drop_prob=0.)
            self.drop = build_dropout(dropout_cfg, default_dropout_cfg)
```

PR #1013, opened 2021-05-08
Support resume for fp16 training

ycxioooong:

## Motivation
Sometimes the resumed fp16 training produces wired behaviors because the current fp16 training process does not save `Fp16OptimizerHook.loss_scaler` state to the checkpoint file. 
Hence in this PR, we enable resuming fp16 training by saving and loading fp16 related hook states.

## Modification
Basically, to enable resume in fp16 training, we add the following steps:
1. Save loss_scaler state_dict: at each iteration, the OptimizerHook will save its loss_scaler's state_dict to runner.meta;
2. At save time, CheckpointHook will write the recorded meta into checkpoint file.
3. When resuming, runner will recover the meta dict from the checkpoint file first.
4.  OptimizerHook.before_run will check if the runner has recorded loss_scaler state_dict in runner.meta, if so, the state_dict will be automatically loaded to re-initialize loss_scaler.
[Note that by doing so, there is no need to revise any downstream projects (e.g. You may directly use MMdetection for resuming an fp16 task without any extra code revision).

## BC-breaking (Optional)
Does the modification introduce changes that break the back-compatibility of the downstream repos?
No.

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -324,7 +324,6 @@ class BaseRunner(metaclass=ABCMeta):
     def resume(self,
                checkpoint,
                resume_optimizer=True,
-               resume_meta=True,
                map_location='default'):
         if map_location == 'default':
             if torch.cuda.is_available():
@@ -354,8 +353,8 @@ class BaseRunner(metaclass=ABCMeta):
                 self.logger.info('the iteration number is changed due to '
                                  'change of GPU number')
 
-        if resume_meta:
-            self.meta = checkpoint['meta']
+        # resume meta information meta
+        self.meta = checkpoint['meta']
 
         if 'optimizer' in checkpoint and resume_optimizer:
             if isinstance(self.optimizer, Optimizer):
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -324,6 +324,7 @@ class BaseRunner(metaclass=ABCMeta):
     def resume(self,
                checkpoint,
                resume_optimizer=True,
+               resume_meta=True,
                map_location='default'):
         if map_location == 'default':
             if torch.cuda.is_available():
@@ -352,6 +353,9 @@ class BaseRunner(metaclass=ABCMeta):
                                  self.world_size)
                 self.logger.info('the iteration number is changed due to '
                                  'change of GPU number')
+
+        if resume_meta:
+            self.meta = checkpoint['meta']
 
         if 'optimizer' in checkpoint and resume_optimizer:
             if isinstance(self.optimizer, Optimizer):
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -100,8 +100,8 @@ if TORCH_VERSION != 'parrots' and TORCH_VERSION >= '1.6.0':
             # wrap model mode to fp16
             wrap_fp16_model(runner.model)
             # resume from state dict
-            if 'fp16.loss_scaler' in runner.meta:
-                scaler_state_dict = runner.meta['fp16.loss_scaler']
+            if 'fp16' in runner.meta and 'loss_scaler' in runner.meta['fp16']:
+                scaler_state_dict = runner.meta['fp16']['loss_scaler']
                 self.loss_scaler.load_state_dict(scaler_state_dict)
 
         def copy_grads_to_fp32(self, fp16_net, fp32_weights):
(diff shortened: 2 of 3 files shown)


mmcv/runner/base_runner.py
```diff
@@ -324,6 +324,7 @@ def load_checkpoint(self,
     def resume(self,
                checkpoint,
                resume_optimizer=True,
+               resume_meta=True,
```
nbei, line 327: The docstring for this function?

mmcv/runner/hooks/optimizer.py
```diff
@@ -99,6 +99,10 @@ def before_run(self, runner):
             """Preparing steps before Mixed Precision Training."""
             # wrap model mode to fp16
             wrap_fp16_model(runner.model)
+            # resume from state dict
+            if 'fp16.loss_scaler' in runner.meta:
+                scaler_state_dict = runner.meta['fp16.loss_scaler']
```
nbei, line 104: I'm not sure whether `fp16.loss_scalar` is suitable here. I may suggest using `_` instead of `.`.

nbei (commented):
Just suggestions.

mmcv/runner/hooks/optimizer.py
ZwwWayne: I suggest we simply use `['fp16']` in cause there are other related keys of FP16 in the future.

mmcv/runner/base_runner.py
```diff
@@ -353,6 +354,9 @@ def resume(self,
                 self.logger.info('the iteration number is changed due to '
                                  'change of GPU number')
 
+        if resume_meta:
```
ZwwWayne, line 357: How about we make it a default behavior to load checkpoint['meta'] if the key 'meta' exists?

ycxioooong: I agree.

mmcv/runner/base_runner.py
```diff
@@ -354,8 +353,8 @@ def resume(self,
                 self.logger.info('the iteration number is changed due to '
                                  'change of GPU number')
 
-        if resume_meta:
-            self.meta = checkpoint['meta']
+        # resume meta information meta
+        self.meta = checkpoint['meta']
```
ZwwWayne, line 357: what if checkpoint does not have 'meta'? in case of BC-breaking, shall we use `checkpoint.get('meta', None)`?

ycxioooong: That's also my original concern. But after reading the codes, I found the original resume function already assumes checkpoint dict always has 'meta' inside (https://github.com/open-mmlab/mmcv/blob/0e01dfc7701821e3f5b8c92fac28a3da7aaf6d62/mmcv/runner/base_runner.py#L340). It makes sense because the checkpoint for resuming is generated by mmcv, hence it must contain 'meta' in the checkpoint file. (And there's no BC breaking here)

ZwwWayne: OK, so we do not need to do that..

mmcv/runner/hooks/optimizer.py
```diff
@@ -276,3 +290,7 @@ def after_train_iter(self, runner):
             if has_overflow:
                 runner.logger.warning('Check overflow, downscale loss scale '
                                       f'to {self.loss_scaler.cur_scale}')
+
+            # save state_dict of loss_scaler
+            runner.meta.setdefault(
```
ZwwWayne, line 295: Shall we reduce the time to update self.loss_scalar? For example, only update the meta when `has_overflow` and at the initialization and resuming.

ycxioooong: I agree that we shall reduce the frequency if possible.

For mmcv based fp16_optimizer, we can do this by checking the value of `has_overflow`. 
But for pytorch.amp based fp16_optimizer, we have to hack the internal status of the loss scaler for `has_overflow`. 
In order to uniform the codes and preserve better readability, the current solution looks better to me. 

Besides, compare with some `if condition`, updating a small and simple dict does not too much extra cost.

ZwwWayne (approved):

ZwwWayne:
Can be merged if the functionality is tested correctly.
1. should we add a unit test for that?
2. Need to involve @LXXXXR  and @AronLin to test with MMCls and MMDet

mmcv/runner/base_runner.py
SuTanTank: bug here. The new config in the resuming run (can be different from the one loaded from meta) is overwritten.

PR #1015, opened 2021-05-09
Fix a bug where logs are missing when two or more loggers were set

ritosonn:
## Motivation
When the users want to check multi-logs like both the CLI and the other (like tensorboard, wandb, mlflow), current implementation may be missing some values because of their `reset_flag` option. This PR fixed these bugs.

## Modification

this PR includes 2 commits.

1. `LoggerHook.before_run(runner)` set a flag whether `self` is the last logger or not. Then, other `LoggerHook`s have to call this method.
2. The `reset_flag`s should be initialized `False` in order to `before_run()` works well.

## BC-breaking (Optional)
nothing.

## Use cases (Optional)
For example, using `MlflowLoggerHook` and `TextLoggerHook` as follows:

```py
from mmcv.runner import IterBasedRunner
from mmcv.runner.hooks.logger.text import TextLoggerHook
from mmcv.runner.hooks.logger.mlflow import MlflowLoggerHook
from logging import Logger
class Model:
    def train_step(self, *args, **kwargs):
        pass
model = Model()
logger = Logger(__name__)
runner = IterBasedRunner(model, logger=logger, work_dir="./")
runner.register_hook(MlflowLoggerHook())
runner.register_hook(TextLoggerHook())
runner.call_hook("before_run")
for hook in runner.hooks:
    print(hook.__class__.__name__, hook.reset_flag)
```

original output: both `LoggerHook`s resets the values, then `TextLoggerHook` miss to print some logging values.

```py
>>> MlflowLoggerHook True
>>> TextLoggerHook True
```

assumed output & fixed output: only `TextLoggerHook` reset the values.

```py
>>> MlflowLoggerHook False
>>> TextLoggerHook True
```

## Checklist

- [ ] Pre-commit or other linting tools are used to fix the potential lint issues.
    - The diff is small, so I didn't install the `pre-commit` environment, and visually checked.
- [ ] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
    - It seems to need another unit test, but I don't know how to construct unit tests. Would you please tell me how to do.
- [ ] If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
    - I couldn't confirmed because I'm not familiar with downstream projects.
 - [ ] The documentation has been modified accordingly, like docstring or example tutorials.
    - It seems the documentation for this diff is not needed.

### DIFF
--- a/mmcv/runner/hooks/logger/mlflow.py
+++ b/mmcv/runner/hooks/logger/mlflow.py
@@ -60,6 +60,7 @@ class MlflowLoggerHook(LoggerHook):
 
     @master_only
     def before_run(self, runner):
+        super(MlflowLoggerHook, self).before_run(runner)
         if self.exp_name is not None:
             self.mlflow.set_experiment(self.exp_name)
         if self.tags is not None:
--- a/mmcv/runner/hooks/logger/mlflow.py
+++ b/mmcv/runner/hooks/logger/mlflow.py
@@ -13,7 +13,7 @@ class MlflowLoggerHook(LoggerHook):
                  log_model=True,
                  interval=10,
                  ignore_last=True,
-                 reset_flag=True,
+                 reset_flag=False,
                  by_epoch=True):
         """Class to log metrics and (optionally) a trained model to MLflow.
 
--- a/mmcv/runner/hooks/logger/pavi.py
+++ b/mmcv/runner/hooks/logger/pavi.py
@@ -22,7 +22,7 @@ class PaviLoggerHook(LoggerHook):
                  add_last_ckpt=False,
                  interval=10,
                  ignore_last=True,
-                 reset_flag=True,
+                 reset_flag=False,
                  by_epoch=True,
                  img_key='img_info'):
         super(PaviLoggerHook, self).__init__(interval, ignore_last, reset_flag,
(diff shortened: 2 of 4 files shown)


nbei:
This modification is very helpful and makes the logger system robust in some cases. Currently, most users write the config file like:

```python
log_config = dict(
    interval=100,
    hooks=[
        dict(type='TextLoggerHook'),
        dict(type='TensorboardLoggerHook')])
```
Thus, they will not meet the bug mentioned in this PR. However, once we change the order of the hooks like this:

```python
log_config = dict(
    interval=100,
    hooks=[
        dict(type='TensorboardLoggerHook'),
        dict(type='TextLoggerHook')])
```
The `reset_flag` in the `TensorboardLoggerHook` will cause the missing of some important values.

OpenMMLab-Assistant005:
Hi @ritosonn ！First of all, we want to express our gratitude for your significant PR in the MMCV project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA

If you have WeChat account，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

PR #1018, opened 2021-05-11
fix basemodule

jshilong:
## Motivation
In the original implementation, when init a module with the pre-train model,
https://github.com/open-mmlab/mmcv/blob/c77e95a65f0128c8faee15f0bc301744320d3609/mmcv/runner/base_module.py#L44
if there are children who have init_weights, the pretrain model would be overridden which is not reasonable.

## Modification
Retune when find it is loading a pre-train model.

## BC-breaking (Optional)
None

### DIFF
--- a/mmcv/runner/base_module.py
+++ b/mmcv/runner/base_module.py
@@ -44,6 +44,9 @@ class BaseModule(nn.Module, metaclass=ABCMeta):
             if self.init_cfg:
                 initialize(self, self.init_cfg)
                 if isinstance(self.init_cfg, (dict, ConfigDict)):
+                    # Avoid the parameters of the pre-training model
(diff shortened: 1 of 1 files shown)


mmcv/runner/base_module.py
```diff
@@ -41,6 +43,10 @@ def init_weights(self):
         if not self._is_init:
             if self.init_cfg:
                 initialize(self, self.init_cfg)
+                if isinstance(self.init_cfg, (dict, ConfigDict)):
+                    if self.init_cfg['type'] == 'Pretrained':
```
ZwwWayne, line 47: Need to add comments to indicate the reason and refer this PR in the code.

PR #1020, opened 2021-05-12
[Fix] Fix load_ckpt from pavi and s3

LXXXXR:
Hi,
This pull request fixes load checkpoint from pavi and s3. All ranks download ckpt now. Pls kindly take a look.Thank you.

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -293,7 +293,7 @@ def load_from_http(filename, map_location=None, model_dir=None):
 @CheckpointLoader.register_scheme(prefixes='pavi://')
 def load_from_pavi(filename, map_location=None):
     """load checkpoint through the file path prefixed with pavi. In distributed
-    setting, this function download ckpt at all ranks to to different temporary
+    setting, this function download ckpt at all ranks to different temporary
     directories.
 
     Args:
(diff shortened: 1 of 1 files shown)


mmcv/runner/checkpoint.py
```diff
@@ -293,7 +293,7 @@ def load_from_http(filename, map_location=None, model_dir=None):
 @CheckpointLoader.register_scheme(prefixes='pavi://')
 def load_from_pavi(filename, map_location=None):
     """load checkpoint through the file path prefixed with pavi. In distributed
-    setting, this function only download checkpoint at local rank 0.
+    setting, this function dowload ckpt at all ranks to seperate temp dir.
```
ZwwWayne, line 296: to different temporary directory

ZwwWayne (commented):
LGTM, see if related developers have any comments.
@xvjiarui @innerlee

mmcv/runner/checkpoint.py
innerlee, line 296: typo

innerlee:
This is a waste of bandwidth most of the time haha.

Will these downloaded chkpts be cached anywhere? or they are transient

ZwwWayne:
> This is a waste of bandwidth most of the time haha.
> 
> Will these downloaded chkpts be cached anywhere? or they are transient

A common situation is that we do not know the job is run on how many separated machines without shared storage. In such a case, the simplest way is to download in each rank.

ZwwWayne (approved):

innerlee:
@ZwwWayne How about making it a two-stage process.

1. Step 1: download ckpt at rank 0
2. Step 2: for all ranks that cannot find the file, download by itself

We can avoid unnecessary downloads altogether, and will have very little influence on user experience.

Issue #1028, opened 2021-05-17
Add fp16 unit tests to the CUDA ops when pytorch >= 1.6


AronLin:
**Describe the Issue**

I notice that there was a runtime error #1004 and it was fixed in mmcv 1.3.4.
However, the same problem still exists when using DCNv2.

So I suggest to check the support of CUDA ops for fp16 or add unit tests to CUDA ops to support fp16，when pytorch >= 1.6.

**Environment**
pytorch >= 1.6.0
mmcv >= 1.3.2

I will send a PR to fix the bug when using DCNv2 with the same way in PR #1014

zhouzaida:
thanks

PR #1014, opened 2021-05-08
[Fix] Fix DCN fp16

shinya7y:
## Motivation
Fix https://github.com/open-mmlab/mmcv/issues/1004

## Modification
Cast DCN weights by type_as, not by autocast.

## BC-breaking (Optional)
No.

Issue #1004, opened 2021-04-29
RuntimeError: expected scalar type Half but found Float (DCN, fp16, MMCV 1.3.2)


shinya7y:
**Describe the Issue**

MMCV 1.3.2 causes the RuntimeError for DCN with fp16 training.

**Reproduction**

Training with the following mmdet config reproduces the error.

faster_rcnn_r50_fpn_fp16_dconv_c3-c5_1x_coco.py
```python
_base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
model = dict(
    backbone=dict(
        dcn=dict(type='DCN', deform_groups=1, fallback_on_stride=False),
        stage_with_dcn=(False, True, True, True)))
fp16 = dict(loss_scale=512.)
```

**Environment**

* Error:
  * PyTorch: 1.7.1, MMCV: 1.3.2, MMDetection: 2.11.0+b536fb0
* No error:
  * PyTorch: 1.7.1, MMCV: 1.3.1, MMDetection: 2.11.0+b536fb0
  * PyTorch: 1.7.1, MMCV: 1.3.0, MMDetection: 2.11.0+b536fb0

**Error traceback**
If applicable, paste the error traceback here.

```none
  File "/home/shinya7y/miniconda3/envs/mmd2110/lib/python3.7/site-packages/mmcv/runner/fp16_utils.py", line 124, in new_func
    output = old_func(*new_args, **new_kwargs)
...
  File "/home/shinya7y/work/mmdetection/mmdet/models/backbones/resnet.py", line 272, in _inner_forward
    out = self.conv2(out)
  File "/home/shinya7y/miniconda3/envs/mmd2110/lib/python3.7/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/home/shinya7y/miniconda3/envs/mmd2110/lib/python3.7/site-packages/mmcv/ops/deform_conv.py", line 355, in forward
    self.dilation, self.groups, self.deform_groups)
  File "/home/shinya7y/miniconda3/envs/mmd2110/lib/python3.7/site-packages/mmcv/ops/deform_conv.py", line 100, in forward
    im2col_step=cur_im2col_step)
RuntimeError: expected scalar type Half but found Float
```

shinya7y:
If no progress, I will send a PR.

ZwwWayne:
Hi @shinya7y ,
[...]

PR #1029, opened 2021-05-17
[Fix] Support amp (pytorch >= 1.6.0) on DCN and DCNv2/ Add unit tests on DCN/DCNv2 amp

AronLin:
## Motivation
mmcv 1.3.2 supports automatic mixed precision with pytorch >= 1.6.0  but the CUDA ops don't support it now, resulting in some runtime errors like #1004 and #1028.

#1014 fixes the bug in #1004. However, when the type of input data for DCN is float32, the bug still exists. The same bug also exits in DCNv2.

Here is I'd like to do:

- Fixes the bugs whether the input type is float16 or float32, when using amp.
- Add amp unit tests for DCN/DCNv2 ops.

## Modification

- Cast the type of DCN/DCNv2 _weight_ and _input_ manually according to the type of _offset_. 
    The _offset_ in deform_conv is computed by a Conv2d which supports amp, so the type of _offset_ is the flag for whether we use fp16 or amp.
- Add amp unit tests for DCN/DCNv2 ops. 
    With amp (pytorch >= 1.6.0), the model type will not be set manually. The input type might be float32 and float16; the unit tests should include both cases.

## BC-breaking (Optional)
No.

### DIFF
--- a/mmcv/ops/deform_conv.py
+++ b/mmcv/ops/deform_conv.py
@@ -71,7 +71,7 @@ class DeformConv2dFunction(Function):
         ctx.im2col_step = im2col_step
 
         # The flag for whether to use fp16 (pytorch < 1.6.0) or
-        # map (pytorch >= 1.6.0) is the type of "offset", we
+        # amp (pytorch >= 1.6.0) is the type of "offset", we
         # cast weight and input to temporarily support fp16 and
         # amp whatever the pytorch version is.
         input = input.to(offset.dtype)
--- a/mmcv/ops/deform_conv.py
+++ b/mmcv/ops/deform_conv.py
@@ -70,8 +70,11 @@ class DeformConv2dFunction(Function):
         ctx.deform_groups = deform_groups
         ctx.im2col_step = im2col_step
 
-        # until the code is modified for torch.cuda.amp.autocast,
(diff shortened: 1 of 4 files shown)


xvjiarui (approved):

mmcv/ops/deform_conv.py
```diff
@@ -70,8 +70,14 @@ def forward(ctx,
...
-        # we need to cast weight to avoid type mismatch in fp16 training
+        # When pytorch version >= 1.6.0, amp is adopted for fp16 mode;
+        # amp won't cast the type of model (float32), but "offset" is cast
+        # to float16 by nn.Conv2d automatically, leading to the type
+        # mismatch with input (when it is float32) or weight.
+        # The flag for whether to use fp16 or amp is the type of "offset",
+        # we cast weight and input to temporarily support fp16 and amp
+        # whatever the pytorch version is.
+        input = input.to(offset.dtype)
         weight = weight.type_as(input)
```
ZwwWayne, line 81: We can use .to() or .type_as() for all cases.  Please unify the usages.

AronLin: @ZwwWayne Done.

ZwwWayne (approved):

ZwwWayne (approved):

PR #1041, opened 2021-05-21
Fix typos

innerlee:
Note: a class name is changed.


ZwwWayne:
The main issue in this PR is the `TRTWraper` is changed to `TRTWrapper`, which will cause BC-breaking in MMDetection [here](https://github.com/open-mmlab/mmdetection/blob/master/tools/deployment/onnx2tensorrt.py#L10). Can you also add a deprecated warning for `TRTWraper` as a transition？We may delete it in the next month.

ZwwWayne (approved):

ZwwWayne:
Can be merged after resolving conflicts.

PR #1042, opened 2021-05-21
add ms_deformable_attn in parrots

luopeichao:
## Modification
- Add ms_deformable_attn support in parrots.
- Modify parrots ext_module loading strategy. Now you will get a fake function if the function you try to load is not implemented in parrots, and the fake function can raise exception when you call it.


ZwwWayne (approved):

ZwwWayne:
PR can be merged after fixing lint issues.

PR #1043, opened 2021-05-21
Imporve windows support for list_from_file

innerlee:
* The default encoding in some windows machine is `'cp936'`, etc. We surely want utf-8.
* Remove \r in windows.


zhouzaida (approved):
LGTM

PR #1046, opened 2021-05-24
Use more cpus in building

innerlee:
The time saved is to be benchmarked.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install system dependencies
         run: sudo apt-get update && sudo apt-get install -y ffmpeg libturbojpeg
+      - name: Install utils
+        run: pip install psutil
       - name: Build and install
         run: rm -rf .eggs && pip install -e .
       - name: Validate the installation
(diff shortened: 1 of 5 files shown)


ZwwWayne (approved):

ZwwWayne:
ping @innerlee , CI failed.

innerlee:
The building time in CI does not decrease, it seems.

.github/workflows/build.yml
```diff
@@ -221,6 +229,8 @@ jobs:
           python-version: 3.7
       - name: Install system dependencies
         run: brew install ffmpeg jpeg-turbo
+      - name: Install utils
```
ZwwWayne, line 232: Does this mean mmcv need one more dependency after this PR?

innerlee: Yes

ZwwWayne: then should we update requirements?

innerlee: The problem is that, this pkg is used in the setup.py script, which will be run before requirements are parsed.

So the current code works like this:

* if psutil is installed prior compiling mmcv, then it will use (num_of_cpu - 1) threads
* otherwise, use the default value 4

innerlee:
Note that whether this setting acutally speed up things is not verified yet. Compare 

before https://github.com/open-mmlab/mmcv/actions/runs/970064971

image

and after https://github.com/open-mmlab/mmcv/actions/runs/942855516

image

Maybe CI does not have many cpus

ZwwWayne:
Then can we verify it on the internal server?

innerlee:
Maybe @zhouzaida would be interested in this

zhouzaida:
> Maybe @zhouzaida would be interested in this

yet, I test it in our internal machine. It actually speeds the compiling process because it automatically detects the number of CPU and uses more CPUs to compile mmcv.

ZwwWayne:
Need to resolve conflicts before merge.

PR #1057, opened 2021-05-28
[Features] Support name of the base classe matching in init_cfg

MeowZheng:
## Motivation

Previous ``INITIALIZERS`` only initialize the class in  ``layer`` of  ``init_cfg``. However, there is a requirement to initialize derived classes from a base class like ``_ConvNd``, ``_BatchNorm``. This pr is for matching the name of the base class, and we can specify the ``layer`` of ``init_cfg`` as the base names and initialize its derived classes.  

## Modification

1. Add ``_get_bases_name`` in mmcv/cnn/utils/weight_init.py.
2. Revise initializers to determine whether base name or class name is in ``layer`` of ``init_cfg``.
3. Add related unit tests.

### DIFF
--- a/mmcv/cnn/utils/weight_init.py
+++ b/mmcv/cnn/utils/weight_init.py
@@ -93,6 +93,10 @@ def bias_init_with_prob(prior_prob):
     return bias_init
 
 
+def _get_bases_name(m):
+    return [b.__name__ for b in m.__class__.__bases__]
+
+
 class BaseInit(object):
 
     def __init__(self, *, bias=0, bias_prob=None, layer=None):
@@ -146,7 +150,8 @@ class ConstantInit(BaseInit):
                 constant_init(m, self.val, self.bias)
             else:
                 layername = m.__class__.__name__
-                if layername in self.layer:
+                basesname = _get_bases_name(m)
+                if bool(set(self.layer) & set([layername] + basesname)):
                     constant_init(m, self.val, self.bias)
 
         module.apply(init)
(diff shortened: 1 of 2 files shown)


mmcv/cnn/utils/weight_init.py
```diff
@@ -146,7 +150,8 @@ def init(m):
                 constant_init(m, self.val, self.bias)
             else:
                 layername = m.__class__.__name__
-                if layername in self.layer:
+                basesname = _get_bases_name(m)
+                if bool(set(self.layer) & set([layername] + basesname)):
```
xvjiarui, line 154:
```suggestion
                if len(set(self.layer) & set([layername] + basesname)):
```
maybe be more intuitive.

mmcv/cnn/utils/weight_init.py
```diff
@@ -93,6 +93,10 @@ def bias_init_with_prob(prior_prob):
     return bias_init
 
 
+def _get_bases_name(m):
+    return [b.__name__ for b in m.__class__.__bases__]
```
xvjiarui, line 97: Just wondering, if there is multi level inheritance in PyTorch, e.g. C->B->A? Since the `__bases__` will only get the first level parent.

MeowZheng: yes, only get the first level parent. ``__bases__.__bases__`` is the second level

xvjiarui (approved):

Issue #783, opened 2021-01-11
decreasing the cycle amplitude after each period in cyclic lr


minouei-kl:
currently, the implementation of cyclic lr does not support decreasing [policies](https://github.com/bckenstler/CLR#policies) which could be really useful.
I wanted to have one cycle per epoch and decrease the cycle amplitude gradually so here is what I did:
set cyclic_times equal to my number of epochs that is 12
define a gamma parameter in CyclicLrUpdaterHook and add these lines in the get_lr method:
```
                if start_iter==0:
                    end_ratio=end_ratio * self.gamma**runner.epoch
                else:
                    start_ratio=start_ratio * self.gamma**runner.epoch
```
the logic here is since there are two phases that one start from zero to max and the other start from the max to the min, I had to decrease the end_ratio in phase one and start_ratio in phase two.
this is working now and after each epoch, the cycle amplitude drops depends on the gamma.
I'm sure that this is not the best way to implement this so I hope this feature gets implemented in the right way.

zhouzaida:
resolved by #1655

PR #1059, opened 2021-05-29
[Enhancement] Add decay to cyclic LR

hokmund:
Implements CLR policy mentioned #783 

Parameter `gamma` allows to add decay to cycles of CLR. If `gamma` is set to a value between 0 and 1, each next cycle will peak at the lower value.

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -366,15 +366,16 @@ class CyclicLrUpdaterHook(LrUpdaterHook):
     3D detection area.
 
     Args:
-        by_epoch (bool): Whether to update LR by epoch.
-        target_ratio (tuple[float]): Relative ratio of the highest LR and the
+        by_epoch (bool, optional): Whether to update LR by epoch.
+        target_ratio (tuple[float], optional): Relative ratio of the highest LR and the
             lowest LR to the initial LR.
-        cyclic_times (int): Number of cycles during training
-        step_ratio_up (float): The ratio of the increasing process of LR in
+        cyclic_times (int, optional): Number of cycles during training
+        step_ratio_up (float, optional): The ratio of the increasing process of LR in
             the total cycle.
-        anneal_strategy (str): {'cos', 'linear'}
+        anneal_strategy (str, optional): {'cos', 'linear'}
             Specifies the annealing strategy: 'cos' for cosine annealing,
             'linear' for linear annealing. Default: 'cos'.
+        gamma (float, optional): Cycle decay ratio. Default: 1.
     """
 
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/lr_updater.py
```diff
@@ -397,10 +399,14 @@ def __init__(self,
...
         assert 0 <= step_ratio_up < 1.0, \
             '"step_ratio_up" must be in range [0,1)'
+        assert 0 < gamma <= 1, \
+            '"gamma" must be in range (0, 1]'
 
         self.target_ratio = target_ratio
         self.cyclic_times = cyclic_times
         self.step_ratio_up = step_ratio_up
+        self.gamma = gamma
+        self.max_iter_per_phase = None
```
zhouzaida, line 409: It is duplicated with line 428.

zhouzaida:
We should add unittest to check whether gamma works as we expected.

mmcv/runner/hooks/lr_updater.py
hokmund: It's duplicated on purpose, since I followed the convention to always declare all fields in the `__init__` method. If this convention is not generally followed in mmcv, I can remove it.

zhouzaida: Agree with you

mmcv/runner/hooks/lr_updater.py
```diff
@@ -366,15 +366,16 @@ class CyclicLrUpdaterHook(LrUpdaterHook):
     3D detection area.
 
     Args:
-        by_epoch (bool): Whether to update LR by epoch.
-        target_ratio (tuple[float]): Relative ratio of the highest LR and the
+        by_epoch (bool, optional): Whether to update LR by epoch.
```
zhouzaida, line 369: please add `Default: False.`

mmcv/runner/hooks/lr_updater.py
```diff
@@ -366,15 +366,16 @@ class CyclicLrUpdaterHook(LrUpdaterHook):
...
+        target_ratio (tuple[float], optional): Relative ratio of the highest LR and the
             lowest LR to the initial LR.
-        cyclic_times (int): Number of cycles during training
-        step_ratio_up (float): The ratio of the increasing process of LR in
+        cyclic_times (int, optional): Number of cycles during training
```
zhouzaida, line 372: append `Default: 1.`

zhouzaida:
hi @hokmund , CI failed.

zhouzaida:
hi @hokmund, do you have time to fix the CI and we want to merge the PR in the next verison

Issue #1064, opened 2021-05-31
Can we save checkpoints of last several epochs?


mzr1996:
**Describe the feature**
If max epoch is 300, only save checkpoints of epoch 290~300.
Or save checkpoints each 10 epochs before epoch 290, and save every checkpoint after epoch 290.

**Motivation**
Because of storage limit, sometimes we cannot save every checkpoint. But we need to choose the best checkpoint in the final period, or merge several last checkpoints.

zhouzaida:
Are you willing to create a PR to resolve it?

mzr1996:
> Are you willing to create a PR to resolve it?

Have created, see https://github.com/open-mmlab/mmcv/pull/1068

zhouzaida:
got it

mzr1996:
Closed by #1068

PR #1068, opened 2021-06-01
Save last several checkpoints by checkpoint hook

mzr1996:
## Motivation
Because of the storage limit, sometimes we cannot save every checkpoint. But we need to choose the best checkpoint in the final period or merge several last checkpoints. See #1064 

## Modification
1. Fix checkpoint hook unit test bug, should use `_build_demo_runner_without_hook`.
    Before, it uses `_build_demo_runner`, and the function will register another checkpoint hook to the runner.

2. Checkpoint hook accepts int `save_last` parameter to decide to save how many checkpoints approaching the end.
    `save_last` parameter also accepts bool as before and will be converted to int, therefore there is no BC-breaking.

## Checklist
- [ x ] Pre-commit or other linting tools are used to fix the potential lint issues.
- [ x ] The modification is covered by complete unit tests. If not, please add more unit tests to ensure correctness.
- [ x ] If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
- [ x ] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmcv/runner/hooks/checkpoint.py
+++ b/mmcv/runner/hooks/checkpoint.py
@@ -24,8 +24,10 @@ class CheckpointHook(Hook):
             In some cases we want only the latest few checkpoints and would
             like to delete old ones to save the disk space.
             Default: -1, which means unlimited.
-        save_last (bool, int): Whether to force the last several checkpoints to
+        save_last (bool): Whether to force the last several checkpoints to
             be saved regardless of interval. Default: True.
+        num_last_ckpts (int): How many last checkpoints to be saved if
+            ``save_last=True``. Default: 1.
         sync_buffer (bool): Whether to synchronize buffers in different
             gpus. Default: False.
     """
@@ -37,6 +39,7 @@ class CheckpointHook(Hook):
                  out_dir=None,
                  max_keep_ckpts=-1,
                  save_last=True,
+                 num_last_ckpts=1,
                  sync_buffer=False,
                  **kwargs):
         self.interval = interval
@@ -44,7 +47,8 @@ class CheckpointHook(Hook):
         self.save_optimizer = save_optimizer
         self.out_dir = out_dir
         self.max_keep_ckpts = max_keep_ckpts
-        self.save_last = int(save_last)
+        self.save_last = save_last
+        self.num_last_ckpts = num_last_ckpts
         self.args = kwargs
         self.sync_buffer = sync_buffer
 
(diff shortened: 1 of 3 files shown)


zhouzaida:
should we support save checkpoints at specified epochs like epcoh=10 epoch=30 epoch=90

ZwwWayne:
This feature actually has some conflict with `max_keep_ckpts`.
Use `save_last=10` is actually a little bit undesirable because in the first few epochs the checkpoints will not be saved and cannot be resumed if the training is corrupted. To only keep the last few checkpoints, you can simply use max_keep_ckpts. For example, `max_keep_ckpts=10` could achieve similar goal as `save_last=10`.

mmcv/runner/hooks/checkpoint.py
```diff
@@ -58,10 +58,10 @@ def after_train_epoch(self, runner):
 
         # save checkpoint for following cases:
         # 1. every ``self.interval`` epochs
-        # 2. reach the last epoch of training
-        if self.every_n_epochs(
-                runner, self.interval) or (self.save_last
-                                           and self.is_last_epoch(runner)):
+        # 2. reach the last `self.save_last` epochs of training
+        if self.every_n_epochs(runner, self.interval) or (
```
ZwwWayne, line 62: This would be better.
```python
if self.every_n_epochs(runner, self.interval) \
    or (self.save_last and self.near_last_epoch(runner, self.save_last)):
```

mmcv/runner/hooks/checkpoint.py
```diff
@@ -109,12 +112,17 @@ def after_train_iter(self, runner):
...
             runner.logger.info(
                 f'Saving checkpoint at {runner.iter + 1} iterations')
             if self.sync_buffer:
                 allreduce_params(runner.model.buffers())
             self._save_checkpoint(runner)
+
+    def near_last_epoch(self, runner):
+        return runner.epoch + self.num_last_ckpts >= runner._max_epochs
+
+    def near_last_iter(self, runner):
```
ycxioooong, line 127: In iter-based runner, usually, the users may want to save checkpoint more frequently. However, save checkpoint per iteration is too frequent. A parameter interval for this case could be considered.

mmcv/runner/hooks/checkpoint.py
```diff
@@ -24,8 +24,10 @@ class CheckpointHook(Hook):
             In some cases we want only the latest few checkpoints and would
             like to delete old ones to save the disk space.
             Default: -1, which means unlimited.
-        save_last (bool): Whether to force the last checkpoint to be saved
-            regardless of interval.
+        save_last (bool): Whether to force the last several checkpoints to
+            be saved regardless of interval. Default: True.
+        num_last_ckpts (int): How many last checkpoints to be saved if
```
ZwwWayne, line 29: The number of the last several checkpoints to be saved

mmcv/runner/hooks/checkpoint.py
mzr1996: So maybe we need a piece-wise interval parameter?

ZwwWayne: We can do that. Suggest having a design first.

mzr1996:
This need has been solved by `save_best` and `max_keep_ckpts`.

PR #1075, opened 2021-06-02
Add DvcliveLoggerHook

daavoo:
## Motivation

I would like to use https://dvc.org/doc/dvclive as a Logger.

## Modification

I added a new `LoggerHook` child to the `HOOKS` registry.

## Use cases (Optional)

Logging training parameters with `dvclive` (https://github.com/iterative/dvclive)


mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -0,0 +1,53 @@
...
+@HOOKS.register_module()
+class DvcliveLoggerHook(LoggerHook):
+
+    def __init__(self,
+                 path,
+                 interval=10,
+                 ignore_last=True,
+                 reset_flag=True,
+                 by_epoch=True):
+        """Class to log metrics with dvclive.
```
zhouzaida, line 16: doctsring should be moved to line 9

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -0,0 +1,53 @@
...
+
+        It requires `dvclive`_ to be installed.
+
+        Args:
+            path (str): directory where dvclive will write TSV log files.
```
zhouzaida, line 21: directory -> Directory

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -0,0 +1,53 @@
...
+            interval (int): Logging interval (every k iterations).
```
zhouzaida, line 22: Append `Default: 10.`

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -0,0 +1,53 @@
...
+            ignore_last (bool): Ignore the log of last iterations in each epoch
```
zhouzaida, line 23: Append `Default: True.`

tests/test_runner/test_hooks.py
```diff
@@ -931,6 +932,23 @@ def test_neptune_hook():
...
+    sys.modules['dvclive'] = MagicMock()
+    runner = _build_demo_runner()
+
+    (tmp_path / 'dvclive').mkdir()
+    hook = DvcliveLoggerHook(str(tmp_path / 'dvclive'))
+    loader = DataLoader(torch.ones((5, 2)))
+
+    runner.register_hook(hook)
+    runner.run([loader, loader], [('train', 1), ('val', 1)])
+    shutil.rmtree(runner.work_dir)
```
zhouzaida, line 945: Does we need remove `runner.work_dir`

daavoo: I need to check if it is really necessary

daavoo: It is actually needed because we are using `mkdtemp`:

https://github.com/open-mmlab/mmcv/blob/master/tests/test_runner/test_hooks.py#L967

And as stated in the official docs (https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp):

> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

We should probably update that function to use pytest tmp_path fixtures instead:

https://docs.pytest.org/en/6.2.x/tmpdir.html

zhouzaida: got it, thanks!

zhouzaida (approved):
LGTM

PR #739, opened 2020-12-21
Eval hook

dreamerlin:

PR #1076, opened 2021-06-03
EvalHook uses case-insensitive key indicator matching and configurabl…

ly015:
## Motivation
Make the EvalHook more flexible to ease the inheritance and usage of it in the downstream repos. 

## Modification
1. Use case-insensitive key indicator matching to infer the comparing rule.
2. Use a configurable test function that can be passed as an initialization parameter. If not set, the default test function "mmcv.engine.single_gpu_test" (or "mmcv.engine.multi_gpu_test" for DistEvalHook) will be used.

## BC-breaking (Optional)
May need to check if there are any case-sensitive key indicators used in downstream repos.


dreamerlin:
Usage and Migration are needed in the PR description for downstream repos, like https://github.com/open-mmlab/mmcv/pull/739

ly015:
@dreamerlin 
Usage and Migration parts are added to the PR message. Please have a look.

dreamerlin:
LGTM

dreamerlin (approved):

mmcv/runner/hooks/evaluation.py
```diff
@@ -367,8 +391,13 @@ def _do_evaluate(self, runner):
         if tmpdir is None:
             tmpdir = osp.join(runner.work_dir, '.eval_hook')
 
-        from mmcv.engine import multi_gpu_test
-        results = multi_gpu_test(
+        if self.test_fn is None:
```
ZwwWayne, line 394: How about we do this once during initialization rather than import the function every epoch?

mmcv/runner/hooks/evaluation.py
```diff
@@ -99,11 +104,14 @@ def __init__(self,
             self.best_ckpt_path = None
```
ZwwWayne, line 104: Is it possible to make greater_keys and less_keys configurable through the initialization arguments. Thus downstream tasks do not need to inherit a new module and modify greater_keys/less_keys anymore.

ly015: Yes, it is possible. Even so, I think it's better to inherit a new EvalHook in downstream repos, to override the default settings like test_fn and greater/less keys. Or they have to be written in the configs, which seems not user-friendly.

mmcv/runner/hooks/evaluation.py
```diff
@@ -124,13 +157,21 @@ def _init_rule(self, rule, key_indicator):
 
         if rule is None:
             if key_indicator != 'auto':
-                if key_indicator in self.greater_keys:
+                # use case-insensitive key indicator to infer the rule
+                key_indicator_lc = key_indicator.lower()
```
ZwwWayne, line 161: indicate that `lc` is the abbrev. for `lower case` in the comments.

mmcv/runner/hooks/evaluation.py
```diff
@@ -124,13 +157,21 @@ def _init_rule(self, rule, key_indicator):
...
+                greater_keys = [key.lower() for key in self.greater_keys]
+                less_keys = [key.lower() for key in self.less_keys]
+
+                print(key_indicator)
```
ZwwWayne, line 165: These print seems to be unnecessary.

mmcv/runner/hooks/evaluation.py
ly015: done.

mmcv/runner/hooks/evaluation.py
ly015: fixed.

mmcv/runner/hooks/evaluation.py
```diff
@@ -95,6 +109,24 @@ def __init__(self,
...
+        if test_fn is None:
+            from mmcv.engine import single_gpu_test
+            self.test_fn = single_gpu_test
+        else:
+            self.test_fn = test_fn
+
+        if greater_keys is None:
+            self.greater_keys = self._default_greater_keys
+        else:
+            assert is_list_of(greater_keys, str)
```
ZwwWayne, line 121: Do we need to also support greater_keys to be a string like 'mIoU'?

xvjiarui: We may check if `isinstance(greater_keys, (list, tuple))`, otherwise, make it into tuple.

ly015:
Kindly ping. @xvjiarui @ZwwWayne

ZwwWayne (approved):

PR #1066, opened 2021-05-31
[Feature] Add flat cosine lr updater

wangg12:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation
In some cases we want to anneal lr after a 'flat lr' period. 

## Modification
Add a flat + cosine annealing lr scheduler similar to https://github.com/fastai/fastai/blob/master/fastai/callback/schedule.py#L128.

## BC-breaking (Optional)
Does the modification introduce changes that break the back-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)
If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1079, opened 2021-06-05
Fix warmup lr

twmht:

## Motivation
Fix bug when warmup by epoch.

## Modification

When using the following `lr_config`, this is how [ranger](https://github.com/lessw2020/Ranger-Deep-Learning-Optimizer])  works.

```
lr_config={
type='CosineAnnealing',
warmup='constant',
warmup_iters=75,
warmup_by_epoch=True
min_lr=0,
by_epoch=True,
warmup_ration=1
}

total_epochs=100
```

in the current implementations, it does not work as expected. it would decay the learning rate by cosine in second epoch.

this PR would fix the bug.

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -109,6 +109,7 @@ class LrUpdaterHook(Hook):
         # it will be set according to the optimizer params
         if isinstance(runner.optimizer, dict):
             self.base_lr = {}
+            self.group_lr = {}
             for k, optim in runner.optimizer.items():
                 for group in optim.param_groups:
                     group.setdefault('initial_lr', group['lr'])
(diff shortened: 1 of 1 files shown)


wangg12:
The lr schedule for ranger could be achieved by https://github.com/open-mmlab/mmcv/pull/1066

wangg12:
I think the lr functions in mmcv usually treat the beginning iteration as 1 (the first iteration) regardless of warmup. If you change it to the end of warmup iter, it might break the behavior of many other models.

twmht:
@wangg12 

What I think is that this is a bug-fix-PR,  not a feature enhancement PR.

wangg12:
Not sure if it is a "bug fix", because it breaks the old behavior.

twmht:
@wangg12 

Can you explain an example which is broken out by this PR?

PR #1083, opened 2021-06-07
Support variables in base files for configs

innerlee:
You can reference variables in base files now. See the test file for examples.


mmcv/utils/config.py
```diff
@@ -120,6 +122,51 @@ def _substitute_predefined_vars(filename, temp_config_name):
...
+        would work."""
+        with open(filename, 'r', encoding='utf-8') as f:
+            # Setting encoding explicitly to resolve coding issue on windows
+            config_file = f.read()
+        base_var_dict = {}
+        regexp = r'\{\{\s*base\.([\w\.]+)\s*\}\}'
+        base_vars = set(re.findall(regexp, config_file))
+        for base_var in base_vars:
+            randstr = f'_{base_var}_{uuid.uuid4().hex.lower()[:6]}'
+            base_var_dict[randstr] = base_var
```
xvjiarui, line 137: Why we need randstr, for json/yaml config? We may add comments to explain these.

innerlee: They are temporary names used for varaible replacement. Adding some noise to reduce the probability of replacing strings that happens to be the same string

innerlee: Will add a test case for this

innerlee:
Open for discussion: should we use `{{ base.xxx }}` or `{{ _base_.xxx }}`

Benefit of `base`: clean, short.
Benefit of `_base`: can pass flake8.

ZwwWayne:
> Open for discussion: should we use `{{ base.xxx }}` or `{{ _base_.xxx }}`
> 
> Benefit of `base`: clean, short.
> Benefit of `_base`: can pass flake8.

suggest using _base_ to keep consistency with that in the file header.

innerlee:
@ZwwWayne pls check the comment to see if the _ was used as emph in markdown

innerlee:
Another "pro" for base is that we can extend the gramma to "env"

ZwwWayne:
> @ZwwWayne pls check the comment to see if the _ was used as emph in markdown

should be `_base_`

ZwwWayne:
Need to resolve confilcts.

ZwwWayne (approved):

innerlee:
ping @ZwwWayne ~~

PR #1091, opened 2021-06-11
Support image reading while ignoring EXIF orientation info

gaotongxiao:
## Motivation
Sometimes the orientation info in EXIF could lead to unexpected rotation of images, messing up the annotations in dataset. Current imread() would always take the info into account unless flagged with "unchanged", which sacrifices flexibility in data preprocessing.

## Modification
Added support for ignoring the orientation info in both color and grayscale mode. It works for cv2 and pillow backend. Also added unit test for them.


ZwwWayne:
LGTM, see if @innerlee has any comments.

innerlee:
How does turbo-jpeg handles orientation?

gaotongxiao:
> How does turbo-jpeg handles orientation?

Turbo-jpeg only supports "color" and "grayscale" mode now, but I found it always ignore EXIF information when loading images. So explicitly appending the "_ignore_orientation" suffix to both modes seems better?

innerlee:
In this case, turbo-jpeg naturally supports `_ignore_orientation`.
Requiring explict `_ignore_orientation` to turbo-jepg will break user's code. I would not suggest it.

innerlee:
There is also a tiff backend. You should take care of it.

gaotongxiao:
> In this case, turbo-jpeg naturally supports `_ignore_orientation`.
> Requiring explict `_ignore_orientation` to turbo-jepg will break user's code. I would not suggest it.

To be consistent, we can still add supports of `_ignore_orientation` modes to turbo-jpeg which do exactly the same things as the normal modes, but raise `depreciated_api_warning` for normal modes for the current version.

gaotongxiao:
> There is also a tiff backend. You should take care of it.

By default, tifffile also ignores EXIF info.

I've done few tests, and it seems tifffile backend hasn't been thoroughly tested beforehand. I'd recommend just leaving it as it is for now until bugs have been fixed.

innerlee:
>  raise depreciated_api_warning for normal modes for the current version.

This is not needed.

If you treat the current behavior of turbo-jpeg on color flag is a bug, then you can fix this bug by supporting the exif rotations for turbo-jpeg in later pr.

gaotongxiao:
> > raise depreciated_api_warning for normal modes for the current version.
> 
> This is not needed.
> 
> If you treat the current behavior of turbo-jpeg on color flag as a bug, then you can fix this bug by supporting the exif rotations for turbo-jpeg in later pr.

Letting the same flag have different behaviors on different backends could be dangerous, but changing it can break users' code. I would say there are two possible ways to go:
1. We don't change the default behaviors supported by backends, and add clear documentation (and maybe user warning) informing users of their difference 
2.  Or, just treat it as a bug and fix

The first choice seems safest to me

innerlee:
You can do both. First do 1, and then do 2 and undo 1.

Either 1 or 2 does not affect this pr

gaotongxiao:
Leave 2 as a future pr

tests/test_image/test_io.py
```diff
@@ -190,6 +190,22 @@ def test_imread(self):
         img_pil_exif_unchanged = mmcv.imread(
             self.exif_img_path, backend='pillow', flag='unchanged')
         assert img_cv2_exif_unchanged.shape == img_pil_exif_unchanged.shape
+        img_cv2_color_ignore_exif = mmcv.imread(
+            self.exif_img_path, flag='color_ignore_orientation')
+        img_pil_color_ignore_exif = mmcv.imread(
+            self.exif_img_path,
+            backend='pillow',
+            flag='color_ignore_orientation')
+        assert img_cv2_color_ignore_exif.shape == (300, 400, 3)
```
innerlee, line 199: What's the shape of img_cv2_exif  at L185? You may add an assertion on that as well

innerlee (approved):

PR #1100, opened 2021-06-15
Avoid bc-breaking of importing `MultiScaleDeformableAttention`

jshilong:
## Motivation
This PR can solve the BC-breaking of importing the MultiScaleDeformableAttention from ``mmcv.cnn.bricks.transformer``.

## Modification
Add a dummy module to raise the ImportError when exactly using the module instead of when importing it.


hellock:
Raise a warning when users import this module from the old path.

ZwwWayne (approved):

PR #1105, opened 2021-06-17
[Fix] Fix meta information loading in runner.resume

yeliudev:
The original `runner.resume` method loads all the meta information in the checkpoint, including epoch, iter, mmcv_version, mmdet_version (optional) into `runner.meta`. And these information will wrongly override the meta in subsequent checkpoints in `runner.save_checkpoint`. See [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/epoch_based_runner.py#L159).


yeliudev:
Duplicate #1108.

ZwwWayne:
Directly remove that might be inappropriate. We plan to merge #1108 to fix this issue, so this PR is closed.

PR #1048, opened 2021-05-24
[Fix] Fix iter base runner val step no optimizer param bug

shenmishajing:
## Motivation
If we use Iter base runner and set workflow to something like [('train', 1000), ('val', 1000)] in mmdetection project. We will get an error looks like `TypeError: val_step() missing 1 required positional argument: 'optimizer'`, which is caused by the code in mmcv/runner/iter_based_runner.py line 77. When we call self.model.val_step func we do not get the optimizer param to it. 
```python
    @torch.no_grad()
    def val(self, data_loader, **kwargs):
        self.model.eval()
        self.mode = 'val'
        self.data_loader = data_loader
        data_batch = next(data_loader)
        self.call_hook('before_val_iter')
        outputs = self.model.val_step(data_batch, **kwargs)
        if not isinstance(outputs, dict):
            raise TypeError('model.val_step() must return a dict')
        if 'log_vars' in outputs:
            self.log_buffer.update(outputs['log_vars'], outputs['num_samples'])
        self.outputs = outputs
        self.call_hook('after_val_iter')
        self._inner_iter += 1
```
Compare to the same code in train mode, which do not cause bug, because we add self.optimizer as the second param. 
```python
    def train(self, data_loader, **kwargs):
        self.model.train()
        self.mode = 'train'
        self.data_loader = data_loader
        self._epoch = data_loader.epoch
        data_batch = next(data_loader)
        self.call_hook('before_train_iter')
        outputs = self.model.train_step(data_batch, self.optimizer, **kwargs)
        if not isinstance(outputs, dict):
            raise TypeError('model.train_step() must return a dict')
        if 'log_vars' in outputs:
            self.log_buffer.update(outputs['log_vars'], outputs['num_samples'])
        self.outputs = outputs
        self.call_hook('after_train_iter')
        self._inner_iter += 1
        self._iter += 1
```
and the same code of val step in EpochBasedRunner.
```python
        elif train_mode:
            outputs = self.model.train_step(data_batch, self.optimizer,
                                            **kwargs)
        else:
            outputs = self.model.val_step(data_batch, self.optimizer, **kwargs)
```
Therefore, this pr fix this bug by adding self.optimizer as the seconde param when we call self.model.val_step func.
## Modification
Add self.optimizer as the seconde param when we call self.model.val_step func in mmcv/runner/iter_based_runner.py line 77.
change code from
```python
        outputs = self.model.val_step(data_batch, **kwargs)
```
to
```python
[...]

PR #1106, opened 2021-06-17
Remove optimizer arg in val_step

fangwudi:
## Motivation
According to the discussion in PR([Fix] Fix iter base runner val step no optimizer param bug #1048 ),  the optimizer param should be removed from val_step(). And in my opinion, epoch_based_runner should be consistent with iter_based_runner (For now, epoch_based_runner's val_step has optimizer while iter_based_runner not )

## Modification
Remove optimizer arg in val_step

## BC-breaking (Optional)
Does the modification introduce changes that break the back-compatibility of the downstream repos?

## Use cases (Optional)

## Checklist

### DIFF
--- a/docs/runner.md
+++ b/docs/runner.md
@@ -74,7 +74,7 @@ def val(self, data_loader, **kwargs):
     data_batch = next(data_loader)
     # it will execute all before_val_iter function in the hooks registered. You may want to watch out for the order.
     self.call_hook('before_val_iter')
-    outputs = self.model.val_step(data_batch, self.optimizer, **kwargs)
+    outputs = self.model.val_step(data_batch, **kwargs)
     self.outputs = outputs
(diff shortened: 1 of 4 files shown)


zhouzaida:
however, it will break the back-compatibility of the downstream repos

fangwudi:
> however, it will break the back-compatibility of the downstream repos

Maybe downstream repos can be modified accordingly?

zhouzaida:
Optimizer of `val_step` will be removed in a future version. In order to support `IterBasedRunner`, we can temporarily set the `optimizer=None` in the `val_step`.

fangwudi:
> Optimizer of `val_step` will be removed in a future version. In order to support `IterBasedRunner`, we can temporarily set the `optimizer=None` in the `val_step`.

thanks.

PR #1117, opened 2021-06-21
Fix the dimension

jaemin93:
## Motivation
A bug was found while testing modulated deformconv batch inference.

## Modification
- Must use inputs[3].d[0] instead of inputs.[2].d[0]


grimoire (approved):
LGTM!

hellock (approved):

PR #1121, opened 2021-06-21
[Enhancement] Support resize or rescale to multiple

Junjun2016:
## Motivation
To meet the requirement of some networks that the size of the input image should be multiple to the divisor

## Modification
Add `divisor` argument to `imresize` and `imrescale` in `mmcv.image`.
This argument rounds up the resized or rescaled image size to the nearest value that can be divisible by the divisor.

### DIFF
--- a/mmcv/image/geometric.py
+++ b/mmcv/image/geometric.py
@@ -3,6 +3,7 @@ import numbers
 
 import cv2
 import numpy as np
+from torch.nn.modules.utils import _pair as to_2tuple
 
 from .io import imread_backend
 
@@ -50,7 +51,8 @@ def imresize(img,
              return_scale=False,
              interpolation='bilinear',
              out=None,
-             backend=None):
+             backend=None,
+             divisor=None):
     """Resize image to a given size.
 
     Args:
@@ -64,12 +66,18 @@ def imresize(img,
         backend (str | None): The image resize backend type. Options are `cv2`,
             `pillow`, `None`. If backend is None, the global imread_backend
             specified by ``mmcv.use_backend()`` will be used. Default: None.
+        divisor (None | tuple | int): Resized image size will be multiple to
+            divisor. If divisor is tuple, divisor is (w_divisor, h_divisor).
+            Default: None.
 
     Returns:
         tuple | ndarray: (`resized_img`, `w_scale`, `h_scale`) or
             `resized_img`.
     """
     h, w = img.shape[:2]
+    if divisor is not None:
+        divisor = to_2tuple(divisor)
+        size = tuple([int(np.ceil(s / d)) * d for s, d in zip(size, divisor)])
     if backend is None:
         backend = imread_backend
     if backend not in ['cv2', 'pillow']:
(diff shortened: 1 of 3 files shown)


mmcv/image/geometric.py
```diff
@@ -3,6 +3,7 @@
 
 import cv2
 import numpy as np
+from torch.nn.modules.utils import _pair as to_2tuple
```
ZwwWayne, line 6: try not to add it here in case parrots do not support this internal API.

xvjiarui: How about we implement it in MMCV?

ZwwWayne: > How about we implement it in MMCV?

agree

mmcv/image/geometric.py
```diff
@@ -92,6 +94,71 @@ def imresize(img,
...
+                         size=None,
+                         scale_factor=None,
+                         keep_ratio=False,
+                         return_scale=False,
+                         interpolation='bilinear',
+                         out=None,
+                         backend=None):
+    """Resize image according to a given size or scale factor and then rounds
+    up the the resized or rescaled image size to the nearest value that can be
+    divisible by the divisor.
```
ZwwWayne, line 108: can be divided

mmcv/image/geometric.py
```diff
@@ -92,6 +94,71 @@ def imresize(img,
...
+
+    Args:
+        img (ndarray): The input image.
+        divisor (int | tuple): Resized image size will be multiple to divisor.
```
ZwwWayne, line 112: be a multiple of ``divisor``

mmcv/image/geometric.py
```diff
@@ -92,6 +94,71 @@ def imresize(img,
...
+            If divisor is tuple, divisor is (w_divisor, h_divisor).
```
ZwwWayne, line 113: is a tuple, divisor should be

mmcv/image/geometric.py
```diff
@@ -92,6 +94,71 @@ def imresize(img,
...
+        size (None | int | tuple[int]): Target size (w, h). Default: None.
+        scale_factor (None | float | tuple[float]): Multiplier for spatial
+            size. Has to match input size if it is a tuple and the 2D style is
```
ZwwWayne, line 116: Should match input size

mmcv/image/geometric.py
Junjun2016: Refer to this [PR](https://github.com/open-mmlab/mmcv/pull/1125)

ZwwWayne (approved):

ZwwWayne:
LGTM, can be merged if @xvjiarui has no further comments.

mmcv/image/geometric.py
```diff
@@ -92,6 +94,72 @@ def imresize(img,
...
+    h, w = img.shape[:2]
+    if size is not None and scale_factor is not None:
+        raise ValueError('only one of size or scale_factor should be defined')
+    elif size is None and scale_factor is None:
+        raise ValueError('one of size or scale_factor should be defined')
+    elif size is not None:
+        if isinstance(size, int):
+            size = (size, size)
+        if keep_ratio:
+            size = rescale_size((w, h), size, return_scale=False)
```
xvjiarui, line 143: If user input a float number, no error will be raised. But `rescale_size` has different effect for float and tuple[int]. 
We may check `isinstance(size, (int, float))`.

PR #1158, opened 2021-06-29
[Fix] use LooseVersion for version checking

kennymckormick:
## Motivation

Origin version checking compares 2 strings, which leads to bugs like '0.8.0' > '0.10.0'

## Modification

Use LooseVersion for Version Checking.


kennymckormick:
@zhouzaida

zhouzaida (approved):
LGTM

PR #1164, opened 2021-07-01
Resolves the duplicate logs issue

### LINKED ISSUE #1000: log repeat twice
The log print in screen like
**`2021-04-28 17:55:36,274 - mmseg - INFO - Distributed training: True `
`INFO:mmseg:Distributed training: True`**
All log messages ware printed twice in screen, but in log files, the message was only writed once. 
I tried to locate the bug, and found that after run `c = c.parent`(anaconda3/envs/open-mmlab/lib/python3.7/logging/__init__.py, line 1590),  c was changed to `[StreamHandler <stderr> (NOTSET)]`. And then, the message was print again.

------------------------------------------------------------
sys.platform: linux
Python: 3.7.10 (default, Feb 26 2021, 18:47:35) [GCC 7.3.0]
CUDA available: True
GPU 0,1,2,3,4,5,6,7: GeForce RTX 3090
CUDA_HOME: /data/lfxuan/softwares/cuda-11.1
NVCC: Build cuda_11.1.TC455_06.29069683_0
GCC: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
PyTorch: 1.8.1+cu111
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v1.7.0 (Git Hash 7aed236906b1f7a05c0917e5257a1af05e9ff683)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.1
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86
  - CuDNN 8.0.5
  - Magma 2.5.2
  - Build settings: BLAS_INFO=mkl, BUILD_TYPE=Release, CUDA_VERSION=11.1, CUDNN_VERSION=8.0.5, CXX_COMPILER=/opt/rh/devtoolset-7/root/usr/bin/c++, CXX_FLAGS= -Wno-deprecated -fvisibility-inlines-hidden -DUSE_PTHREADPOOL -fopenmp -DNDEBUG -DUSE_KINETO -DUSE_FBGEMM -DUSE_QNNPACK -DUSE_PYTORCH_QNNPACK -DUSE_XNNPACK -O2 -fPIC -Wno-narrowing -Wall -Wextra -Werror=return-type -Wno-missing-field-initializers -Wno-type-limits -Wno-array-bounds -Wno-unknown-pragmas -Wno-sign-compare -Wno-unused-parameter -Wno-unused-variable -Wno-unused-function -Wno-unused-result -Wno-unused-local-typedefs -Wno-strict-overflow -Wno-strict-aliasing -Wno-error=deprecated-declarations -Wno-stringop-overflow -Wno-psabi -Wno-error=pedantic -Wno-error=redundant-decls -Wno-error=old-style-cast -fdiagnostics-color=always -faligned-new -Wno-unused-but-set-variable -Wno-maybe-uninitialized -fno-math-errno -fno-trapping-math -Werror=format -Wno-stringop-overflow, LAPACK_INFO=mkl, PERF_WITH_AVX=1, PERF_WITH_AVX2=1, PERF_WITH_AVX512=1, TORCH_VERSION=1.8.1, USE_CUDA=ON, USE_CUDNN=ON, USE_EXCEPTION_PTR=1, USE_GFLAGS=OFF, USE_GLOG=OFF, USE_MKL=ON, USE_MKLDNN=ON, USE_MPI=OFF, USE_NCCL=ON, USE_NNPACK=ON, USE_OPENMP=ON, 

TorchVision: 0.9.1+cu111
OpenCV: 4.5.1
MMCV: 1.3.2
MMCV Compiler: GCC 9.3
MMCV CUDA Compiler: 11.1
MMSegmentation: 0.9.0+fed4b96
------------------------------------------------------------

qizhuli:
## Motivation
The console handler present in the root logger leads to duplicate messages being printed on the console, creating unnecessary clutter. This has also been discussed [here](https://discuss.pytorch.org/t/distributed-1-8-0-logging-twice-in-a-single-process-same-code-works-properly-in-1-7-0/114103/6). This PR fixes this issue, and resolves #1000. 

## Modification 

In `mmcv.utils.get_logger()`, all `StreamHandler` present in the root logger are set to log at the `ERROR` level, instead of `NOTSET` which is the default.

### DIFF
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -28,12 +28,6 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
         logging.Logger: The expected logger.
     """
     logger = logging.getLogger(name)
-
-    # set the root logger's StreamHandler, if any, to log at the ERROR level
-    for handler in logger.root.handlers:
-        if type(handler) is logging.StreamHandler:
-            handler.setLevel(logging.ERROR)
-
     if name in logger_initialized:
         return logger
     # handle hierarchical names
@@ -42,6 +36,17 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
     for logger_name in logger_initialized:
         if name.startswith(logger_name):
             return logger
+
+    # handle duplicate logs to the console
+    # Starting in 1.8.0, PyTorch DDP attaches a StreamHandler <stderr> (NOTSET)
+    # to the root logger. As logger.propagate is True by default, this root
+    # level handler causes logging messages from rank>0 processes to
+    # unexpectedly show up on the console, creating much unwanted clutter.
+    # To fix this issue, we set the root logger's StreamHandler, if any, to log
+    # at the ERROR level.
+    for handler in logger.root.handlers:
+        if type(handler) is logging.StreamHandler:
+            handler.setLevel(logging.ERROR)
 
     stream_handler = logging.StreamHandler()
     handlers = [stream_handler]
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -28,7 +28,12 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
         logging.Logger: The expected logger.
     """
     logger = logging.getLogger(name)
-    logger.root.handlers.clear()
+
+    # set the root logger's StreamHandler, if any, to log at the ERROR level
+    for handler in logger.root.handlers:
+        if type(handler) is logging.StreamHandler:
+            handler.setLevel(logging.ERROR)
+
     if name in logger_initialized:
         return logger
     # handle hierarchical names
@@ -93,6 +98,7 @@ def print_log(msg, logger=None, level=logging.INFO):
     elif isinstance(logger, str):
         _logger = get_logger(logger)
         _logger.log(level, msg)
+        return _logger
     else:
         raise TypeError(
             'logger should be either a logging.Logger object, str, '
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -28,6 +28,7 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
         logging.Logger: The expected logger.
     """
     logger = logging.getLogger(name)
+    logger.root.handlers.clear()
     if name in logger_initialized:
         return logger
     # handle hierarchical names
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -98,7 +98,6 @@ def print_log(msg, logger=None, level=logging.INFO):
     elif isinstance(logger, str):
         _logger = get_logger(logger)
         _logger.log(level, msg)
-        return _logger
     else:
         raise TypeError(
             'logger should be either a logging.Logger object, str, '


zhouzaida:
hi @qizhuli ,please fix the CI

qizhuli:
@zhouzaida Hi there. I have revised my commit to satisfy the tests. Now it sets any StreamHandler associated with the root logger to log at the ERROR level (instead of NOTSET by default).

Over the past few days, I did a little more searching on this matter, and it seems to me that the cleanest solution would be to do `logger.propagate = False` right after calling `logger = logging.getLogger(name)`, which would stop messages from being sent to the root logger completely. Coincidentally, that's exactly the [approach](https://github.com/facebookresearch/detectron2/blob/a17b768895a1207c540e909225fa5a9389564927/detectron2/utils/logger.py#L58-L60) adopted by Detectron2.  Unfortunately, this approach is problematic with the `caplog` fixture in mmcv's tests, as killing child-to-parent propagation would cut off the LogCaptureHandler too, on which `caplog` depends.

mmcv/utils/logging.py
```diff
@@ -28,6 +28,12 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
         logging.Logger: The expected logger.
     """
     logger = logging.getLogger(name)
+
+    # set the root logger's StreamHandler, if any, to log at the ERROR level
+    for handler in logger.root.handlers:
```
zhouzaida, line 33: should those lines be executed after Line 37?

```python
if name in logger_initialized:
    return logger
 
for handler in logger.root.handlers:
   ...
```

zhouzaida: More comments should be added here.

qizhuli: Great suggestions! I've modified the code and added more comments accordingly.

mmcv/utils/logging.py
```diff
@@ -37,6 +37,17 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
...
 
+    # handle duplicate logs to the console
+    # Starting in 1.8.0, PyTorch DDP attaches a StreamHandler <stderr> (NOTSET)
+    # to the root logger. As logger.propagate is True by default, this root
+    # level handler causes logging messages from rank>0 processes to
+    # unexpectedly show up on the console, creating much unwanted clutter.
+    # To fix this issue, we set the root logger's StreamHandler, if any, to log
+    # at the ERROR level.
+    for handler in logger.root.handlers:
+        if type(handler) is logging.StreamHandler:
```
zhouzaida, line 48: Should we change `if type(handler) is logging.StreamHandler ` to `if isinstance(hanlder, logging.StreamHandler)`? 
please refer to https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

qizhuli: Actually I tried that first before switching to the current version. Problem with `isinstance` is that it also returns True for subclasses of `logging.StreamHandler`, which unfortunately includes `pytest.logging.LogCaptureHandler` (see here https://docs.pytest.org/en/6.2.x/_modules/_pytest/logging.html for details), and causes unit tests that rely on `caplog` to fail.

zhouzaida: got it

ZwwWayne (approved):

ZwwWayne:
Can be merged after verification in downstream repos like MMSeg/MMDet

PR #1181, opened 2021-07-08
[Docs] Add header for files

xvjiarui:


innerlee:
May you put the script in `tools/`

mmcv/cnn/bricks/__init__.py
```diff
@@ -1,3 +1,4 @@
+# Copyright (c) Open-MMLab. All rights reserved.
```
hellock, line 1: We need to change all `Open-MMLab` to `OpenMMLab`.

hellock:
Apply this to cpp/cu/cuh files too.

xvjiarui:
@hellock done

hellock:
In some files, e.g., cu, cpp, Open-MMLab is not replaced with OpenMMLab.

xvjiarui:
@hellock All files are replaced.

hellock (approved):

PR #1183, opened 2021-07-08
[Feature] Support reduce_lr_update

gengenkai:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmcv/runner/hooks/lr_updater.py
+++ b/mmcv/runner/hooks/lr_updater.py
@@ -1,6 +1,7 @@
 # Copyright (c) Open-MMLab. All rights reserved.
 import numbers
 from math import cos, pi
+from typing import Optional
 
 import mmcv
 from .hook import HOOKS, Hook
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/lr_updater.py
```diff
@@ -614,3 +615,223 @@ def format_param(name, optim, param):
...
+                 factor: float = 0.1,
+                 patience: int = 10,
+                 threshold: float = 1e-4,
+                 threshold_mode: str = 'rel',
+                 cooldown: int = 0,
+                 min_lr: float = 0.,
+                 eps: float = 1e-8,
+                 **kwargs):
+        assert isinstance(periods, list), '"periods" must be a list'
+        assert mmcv.is_list_of(periods, int) and all([s >= 0 for s in periods])
```
MeowZheng, line 674: Please check the behavior, when used in ``iterbasedrunner``, or if this lr scheduler does not suit  ``iterbasedrunner``, I think it would be better to assert  ``self.by_epoch`` in ``__inti__``.

mmcv/runner/hooks/lr_updater.py
```diff
@@ -614,3 +615,223 @@ def format_param(name, optim, param):
...
+            if self.is_better(current, self.best):
+                self.best = current
+                self.num_bad_epochs = 0
+            else:
+                self.num_bad_epochs += 1
+
+            if self.in_cooldown:
+                self.cooldown_counter -= 1
+                self.num_bad_epochs = 0
+        print('epoch--', cur_epoch, ' lr:', self.regular_lr)
```
teamwong111, line 782: Shall we use `print` sentence here?

teamwong111 (commented):
LGTM for code. However, it may be difficult to use for users because the `periods` parameters need to be set carefully and it cannot work well in train mode and test mode at the same time. So I am not sure if there is a better design.

PR #1185, opened 2021-07-09
[Improvement] Improve digit_version & use it for version_checking

kennymckormick:


tests/test_utils/test_version_utils.py
```diff
@@ -4,19 +4,19 @@
...
-    assert digit_version('0.2.16') == (0, 2, 16)
-    assert digit_version('1.2.3') == (1, 2, 3)
-    assert digit_version('1.2.3rc0') == (1, 2, 2, 0)
-    assert digit_version('1.2.3rc1') == (1, 2, 2, 1)
-    assert digit_version('1.0rc0') == (1, -1, 0)
+    assert digit_version('0.2.16') == (0, 2, 16, 0, 0, 0)
+    assert digit_version('1.2.3') == (1, 2, 3, 0, 0, 0)
+    assert digit_version('1.2.3rc0') == (1, 2, 3, 0, -1, 0)
+    assert digit_version('1.2.3rc1') == (1, 2, 3, 0, -1, 1)
+    assert digit_version('1.0rc0') == (1, 0, 0, 0, -1, 0)
```
innerlee, line 12: Adding more test case would be great, like `1.5.0+cuda90_cudnn7.6.3_lms` you mentioned earlier

innerlee: Also test the comparison. Like rc0 < rc1, alpha < beta, etc.

ZwwWayne:
need to resolve comments and fix CI

requirements/runtime.txt
```diff
@@ -3,4 +3,5 @@ numpy
 Pillow
 pyyaml
 regex;sys_platform=='win32'
+setuptools>=52
```
innerlee, line 6: Which part of code relies on setuptools' version?

kennymckormick: In earlier versions of setuptools, the version object has no attribute named release.

kennymckormick:
@ZwwWayne CI passed.

setup.py
```diff
@@ -32,6 +32,8 @@ def choose_requirement(primary, secondary):
 
 
 def get_version():
+    import setuptools
+    print(setuptools.__version__, flush=True)
```
zhouzaida, line 36: this line can be removed

tests/test_utils/test_version_utils.py
```diff
@@ -4,19 +4,28 @@
...
+    assert digit_version('1.0') == digit_version('1.0.0')
+    assert digit_version('1.5.0+cuda90_cudnn7.6.3_lms') == digit_version('1.5')
+    assert digit_version('1.0.0dev') < digit_version('1.0.0a')
+    assert digit_version('1.0.0a') < digit_version('1.0.0a1')
+    assert digit_version('1.0.0a') < digit_version('1.0.0b')
+    assert digit_version('1.0.0b') < digit_version('1.0.0rc')
+    assert digit_version('1.0.0rc1') < digit_version('1.0.0')
+    assert digit_version('1.0.0') < digit_version('1.0.0post')
+    assert digit_version('1.0.0post') < digit_version('1.0.0post1')
```
zhouzaida, line 20: add some cases which will fail

zhouzaida:
https://github.com/open-mmlab/mmcv/blob/366c628a85daf71fa26f652b402277c90d9f7d6e/mmcv/utils/trace.py#L8 
This line should also be replaced with `digit_version`

kennymckormick:
> https://github.com/open-mmlab/mmcv/blob/366c628a85daf71fa26f652b402277c90d9f7d6e/mmcv/utils/trace.py#L8
> 
> 
> This line should also be replaced with `digit_version`

resolved

zhouzaida (approved):
LGTM

kennymckormick:
@zhouzaida

kennymckormick:
@ZwwWayne

tests/test_utils/test_version_utils.py
```diff
@@ -1,22 +1,40 @@
...
+    with pytest.raises(AssertionError):
```
ZwwWayne, line 24: Add comments indicate what error you are testing

mmcv/runner/fp16_utils.py
```diff
@@ -122,8 +121,8 @@ def new_func(*args, **kwargs):
                     else:
                         new_kwargs[arg_name] = arg_value
             # apply converted arguments to the decorated method
-            if (TORCH_VERSION != 'parrots'
-                    and LooseVersion(TORCH_VERSION) >= LooseVersion('1.6.0')):
+            if (TORCH_VERSION != 'parrots' and
+                    digit_version(TORCH_VERSION) >= digit_version('1.6.0')):
```
innerlee, line 125: should digit_version deal with "parrots"?

mmcv/runner/hooks/logger/tensorboard.py
```diff
@@ -24,7 +23,7 @@ def __init__(self,
     @master_only
     def before_run(self, runner):
         super(TensorboardLoggerHook, self).before_run(runner)
-        if (LooseVersion(TORCH_VERSION) < LooseVersion('1.1')
+        if (digit_version(TORCH_VERSION) < digit_version('1.1')
```
innerlee, line 26: should test parrots related comparisons, `digit_version("parrots") < digit_version('1.1')` etc.

innerlee:
My sole interest is the test cases. They should be as complete as possible. Here are some idea sources:

https://github.com/JuliaInterop/VersionParsing.jl/blob/master/test/runtests.jl
https://github.com/eieste/VersionParser/blob/master/version_parser/test.py
https://github.com/pypa/packaging/blob/main/tests/test_version.py

mmcv/runner/fp16_utils.py
kennymckormick: We do not deal with "parrots" in version comparison, "parrots" is dealt like a special case.

kennymckormick:
> My sole interest is the test cases. They should be as complete as possible. Here are some idea sources:
> 
> https://github.com/JuliaInterop/VersionParsing.jl/blob/master/test/runtests.jl
> https://github.com/eieste/VersionParser/blob/master/version_parser/test.py
> https://github.com/pypa/packaging/blob/main/tests/test_version.py

Since we use the parse_version API from pkg_resources, the capability of `digit_version` is not unlimited. For example, it can not parse sth. like 'vM1m2b3' and will treat '0.99.1+1' like '0.99.1'. However, I think can parse versions of most python packages correctly. The test cases added define the boundary of its capability.

mmcv/runner/fp16_utils.py
innerlee: Notice this pattern suggests that the digit_version should not accept "parrots" as input. If this is the case, then we can add an assertion to enforce

image

kennymckormick: OK, I will add an assertion in digit_version to make sure that.

innerlee (approved):

zhiqi-li:
image

Can mmcv and mmegine improve the ability of digit_version function to handle custom complied torch version?

kennymckormick:
> image
> 
> Can mmcv and mmegine improve the ability of digit_version function to handle custom compiled torch version?

That might be difficult because you need the entire commit history to make it possible. Besides, there is no guarantee that all commits are in linear (actually they can be parallel, it's not possible to determine the ranking in that case).

PR #1186, opened 2021-07-09
Reimplement cc_attention using einops.

Leojc:
## Motivation
Reimplement a faster and more readable cc_attention using einops. 

## Modification
1. Rewrite `cc_attention.py`
2. Add einops to the `requirement.txt`

### DIFF
--- a/mmcv/ops/cc_attention.py
+++ b/mmcv/ops/cc_attention.py
@@ -15,14 +15,19 @@ def NEG_INF_DIAG(n, device):
 
 class CrissCrossAttention(nn.Module):
 
-    def __init__(self, in_dim):
+    def __init__(self, in_channels):
         super().__init__()
+        self.in_channels = in_channels
         self.query_conv = nn.Conv2d(
-            in_channels=in_dim, out_channels=in_dim // 8, kernel_size=1)
+            in_channels=in_channels,
+            out_channels=in_channels // 8,
+            kernel_size=1)
         self.key_conv = nn.Conv2d(
-            in_channels=in_dim, out_channels=in_dim // 8, kernel_size=1)
+            in_channels=in_channels,
+            out_channels=in_channels // 8,
+            kernel_size=1)
         self.value_conv = nn.Conv2d(
-            in_channels=in_dim, out_channels=in_dim, kernel_size=1)
+            in_channels=in_channels, out_channels=in_channels, kernel_size=1)
         self.gamma = nn.Parameter(torch.zeros(1))
 
     def forward(self, x):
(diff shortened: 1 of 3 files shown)


ZwwWayne:
Hi @Leojc ，
1. Do you have some statistic numbers that verifies the current implementation?
2. I understand that the CUDA ops might be replaced, but the interfaces of this module should not be changed, e.g., in_channels should not be changed to in_dims.

Leojc:
Hi @ZwwWayne ,
1. 
    (1) I calculate the time for one forward (input size is [2,512,97,97]). The CUDA version is 0.0299619 and my implementation is 0.00544943. So about 5x faster.
    (2) I also test it in https://github.com/open-mmlab/mmsegmentation. I replace the cc_attention of ccnet in mmsegmentation. Then I run with a config of `ccnet_r50-d8_769x769_40k_cityscapes.py`. The CUDA version gets 78.46 mIoU, and this implementation gets 78.99. 
2. Sorry about that! I just fixed it.

ZwwWayne:
Thank you @Leojc , the statistical results are important. You may also highlight the speed comparison in the docstring of the module. I will invite @xvjiarui to have look.

mmcv/ops/cc_attention.py
```diff
@@ -1,91 +1,79 @@
...
+    The diagonal are all "-inf". This is for avoiding calculating the
+    overlapped element in the Criss-Cross twice.
+    """
+    return torch.diag(torch.tensor(float('-inf')).to(device).repeat(n), 0)
 
 
 class CrissCrossAttention(nn.Module):
-    """Criss-Cross Attention Module."""
+    """Criss-Cross Attention Module.
+
```
ZwwWayne, line 18: Suggest to also update arguments in the docstring.

Leojc: OK.

hellock:
Can it be implemented with `torch.einsum` so that we can omit a third-party dependency?

Leojc:
> 
> 
> Can it be implemented with `torch.einsum` so that we can omit a third-party dependency?

Yes, but that will decrease the readability greatly. Is that OK?

Leojc:
> 
> 
> > Can it be implemented with `torch.einsum` so that we can omit a third-party dependency?
> 
> Yes, but that will decrease the readability greatly. Is that OK?

Wait... Maybe that won't decrease too much. I'll try implementing it.

Leojc:
Think I'll open a new PR.
Reimplement cc_attention using pure pytorch #1201

hellock:
> Think I should open a new PR.
> Reimplement cc_attention using pure PyTorch #1201

Looks great!

PR #1187, opened 2021-07-09
[Feature] Add `is_tracing` to wrap `torch.jit.is_tracing` in different versions.

mzr1996:
## Motivation

Head needs different behaviors in test and export, so we need to use `torch.jit.is_tracing` to judge different situations.

But `torch.jit.is_tracing` has a bug in PyTorch 1.6.0, so we need a wrapper to support trace judgement.

## Modification

Add a helper function in `mmcv/utils`, and add unit tests.

## BC-breaking (Optional)

No

## Use cases (Optional)

```python
import torch
from mmcv.utils import is_tracing

def foo(x):
    if is_tracing():
        return x
    else:
        return x.tolist()

>>> x = torch.ones(3)
>>> foo(x)
[1.0, 1.0, 1.0]
>>> traced_foo = torch.jit.trace(foo, (torch.rand(1), ))
>>> traced_foo(x)
tensor([1., 1., 1.])
```

### DIFF
--- a/mmcv/utils/trace.py
+++ b/mmcv/utils/trace.py
@@ -4,7 +4,7 @@ from distutils.version import LooseVersion
 import torch
 
 
-def is_tracing() -> bool:
+def is_jit_tracing() -> bool:
     if LooseVersion(torch.__version__) >= LooseVersion('1.6.0'):
         on_trace = torch.jit.is_tracing()
         # In PyTorch 1.6, torch.jit.is_tracing has a bug.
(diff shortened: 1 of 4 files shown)


mmcv/utils/trace.py
```diff
@@ -0,0 +1,21 @@
+import warnings
+from distutils.version import LooseVersion
+
+import torch
+
+
+def is_tracing() -> bool:
```
ZwwWayne, line 7: is_jit_tracing might be more clear

ZwwWayne (approved):

PR #1205, opened 2021-07-18
[Feature]: Support empty tensor in MMSyncBN

ZwwWayne:

## Motivation

In detection models trained by strong augmentation, the RoIHead may be trained with empty proposals if the objects are all contained in the cropped images, this will cause failure because the current MMSyncBN does not support training with empty tensor. 

Related discussion/implementations:
1. https://github.com/pytorch/pytorch/issues/36530
2. NaiveSyncBN in [Detectron2](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/batch_norm.py#L149)

## Modification

1. Support to skip the CUDA calculation and leave the mean/var/norm/std/ as zero if the input tensor is empty in both the forward and backward processes. It also synchronizes a `batch_flag` to indicate the global batch size because when the global batch is zero, we need to make the momentum to be zero to skip the update of `running_mean/running_var`.
2. Support a new mode, named `stats_mode` to produce more accurate batch statistics when empty tensor might be given.
3. Use `flatten` rather than `view(-1, ...)` when reshaping the tensor which better handles empty tensor.
4. The creation method of empty tensors is also changed from `empty/empty_like` to `zeros/zeros_like`, because the `torch.empty()` returns uninitialized values of tensor, which might contain unexpected values other than zeros. See related discussion in [stackoverflow](https://stackoverflow.com/questions/51140927/what-is-uninitialized-data-in-pytorch-empty-function).

## BC-breaking (Optional)

The modification does not introduce changes that break the backward compatibility of the downstream repos?
It support the new strong baseline.

## Use cases (Optional)

Inference with empty tensor is supported, see the added unit tests.

## Checklist

2. The modification is covered by complete unit tests.
3. If the modification has potential influence on downstream projects, this PR will be tested with MMDetection in https://github.com/open-mmlab/mmdetection/pull/5537


mmcv/ops/sync_bn.py
```diff
@@ -44,27 +44,48 @@ def forward(self, input, running_mean, running_var, weight, bias, momentum,
...
+        # sync mean and the batch flag
+        vec = torch.cat([mean, batch_flag])
         if self.group_size > 1:
-            dist.all_reduce(mean, group=self.group)
-            mean /= self.group_size
-        ext_module.sync_bn_forward_var(input3d, mean, var)
+            dist.all_reduce(vec, group=self.group)
+            mean = vec[:num_channels] / self.group_size
+        # need total batch to decide whether to update the momentum
+        total_batch = vec[-1].detach().clamp(max=1)
```
nbei, line 78: `total_batch_flag` may be better? In fact, this tensor works as a bool variable. Is it right?

ZwwWayne: Actually yes, the momentum is either zero or the passed value for now.
Are you suggesting to make it a bool value or just rename it?

nbei: Just renaming it is ok.

ZwwWayne: OK

nbei (approved):

ZwwWayne:
Please @nbei to check again as a new argument is just added and invite @grimoire to have a look.

mmcv/ops/sync_bn.py
```diff
@@ -86,36 +123,76 @@ def forward(self, input, running_mean, running_var, weight, bias, momentum,
...
 
     def __init__(self,
                  num_features,
                  eps=1e-5,
                  momentum=0.1,
                  affine=True,
                  track_running_stats=True,
-                 group=None):
+                 group=None,
+                 stats_mode=''):
```
grimoire, line 195: Do we have better name instead of empty string?

mmcv/ops/sync_bn.py
```diff
@@ -30,41 +30,78 @@ def symbolic(g, input, running_mean, running_var, weight, bias, momentum,
...
+
+        if self.stats_mode == 'N':
+            var *= batch_size
         if self.group_size > 1:
             dist.all_reduce(var, group=self.group)
+
+        if self.stats_mode == '':
             var /= self.group_size
+        else:
+            var /= total_batch.clamp(min=1)
```
grimoire, line 101: Use `elif self.stats_mode == 'N'` would be better, since we might add other stats_mode in the future?
 A `NotImplementError` could be placed in `else` block.

mmcv/ops/sync_bn.py
```diff
@@ -30,41 +30,78 @@ def symbolic(g, input, running_mean, running_var, weight, bias, momentum,
...
+        norm = torch.zeros_like(
             input3d, dtype=torch.float, device=input3d.device)
-        std = torch.empty(
-            input3d.size(1), dtype=torch.float, device=input3d.device)
+        std = torch.zeros(
+            num_channels, dtype=torch.float, device=input3d.device)
+
+        batch_size = input3d.size(0)
+        if batch_size > 0:
+            ext_module.sync_bn_forward_mean(input3d, mean)
```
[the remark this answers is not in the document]
ZwwWayne: torch.var_mean is a little bit slower then ext_module.sync_bn_forward_mean() and may have different calculation. This PR does not intends to change the calculation of SyncBN but to make it robust to empty tensor.

ZwwWayne:
The performance in MMDetection has been verified through https://github.com/open-mmlab/mmdetection/pull/5537

mmcv/ops/sync_bn.py
ZwwWayne: change to use 'default' for the default mode.

grimoire (approved):

ZwwWayne:
Merged as the accuracy of the final commit should be guaranteed.

PR #1208, opened 2021-07-20
`DvcliveLoggerHook` updates to work with `DVC`

daavoo:
## Motivation

Update to use latest version of `dvclive`: https://github.com/iterative/dvclive/releases


zhouzaida:
hi @daavoo , what is the default directory for saving log, and what will happen if the directory exists

daavoo:
> hi @daavoo , what is the default directory for saving log, and what will happen if the directory exists

The default directory would be `dvclive`. And it it exists, it's contents will be removed (we are discussing about updating this behavior):

https://dvc.org/doc/dvclive/api-reference/init#parameters

When using this callback alongside `DVC`, `dvclive` will get those values from the ones specified in the `stage`:

https://dvc.org/doc/command-reference/stage/add#options

I removed the call to `dvclive.init` and the `path` argument as they are not strictly necessary and to prevent unexpected behaviors with default values (i.e. by default, `dvclive` will be used as output directory)

daavoo:
> The default directory would be `dvclive`. And it it exists, it's contents will be removed (we are discussing about updating this behavior)
There is an open P.R. https://github.com/iterative/dvclive/pull/122

zhouzaida:
got it

daavoo:
@zhouzaida Maybe it would be better to make the default directory `{runner.work_dir}/dvclive` instead of the current `dvclive`?

daavoo:
Given the possibility of adding `custom_imports` to MMCV's config files, we would like to maintain this HOOK as part of the `dvclive` repository instead of here because `dvclive` is under development and we don't want to sature `mmcv` with P.R. for updating the hook.

Would it make sense to have the hook linked inside `mmcv` with something as follows?:

```python
# mmcv/runner/hooks/logger/dvclive.py
from dvclive.mmcv import DVCLiveLoggerHook
```

zhouzaida:
> Given the possibility of adding `custom_imports` to MMCV's config files, we would like to maintain this HOOK as part of the `dvclive` repository instead of here because `dvclive` is under development and we don't want to sature `mmcv` with P.R. for updating the hook.
> 
> Would it make sense to have the hook linked inside `mmcv` with something as follows?:
> 
> ```python
> # mmcv/runner/hooks/logger/dvclive.py
> from dvclive.mmcv import DVCLiveLoggerHook
> ```
> 
> Or should we just remove `mmcv/runner/hooks/logger/dvclive.py`?

In fact, I prefer to keep the DVCLiveLoggerHook in the mmcv

daavoo:
> > Given the possibility of adding `custom_imports` to MMCV's config files, we would like to maintain this HOOK as part of the `dvclive` repository instead of here because `dvclive` is under development and we don't want to sature `mmcv` with P.R. for updating the hook.
> > Would it make sense to have the hook linked inside `mmcv` with something as follows?:
> > ```python
> > # mmcv/runner/hooks/logger/dvclive.py
> > from dvclive.mmcv import DVCLiveLoggerHook
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or should we just remove `mmcv/runner/hooks/logger/dvclive.py`?
> 
> In fact, I prefer to keep the DVCLiveLoggerHook in the mmcv

I will update the P.R. with the latest changes then

zhouzaida:
hi @daavoo , please fix the CI

daavoo:
> hi @daavoo , please fix the CI

It is passing now :)

tests/test_runner/test_hooks.py
```diff
@@ -1108,17 +1108,14 @@ def test_dvclive_hook(tmp_path):
     sys.modules['dvclive'] = MagicMock()
     runner = _build_demo_runner()
 
-    (tmp_path / 'dvclive').mkdir()
-    hook = DvcliveLoggerHook(str(tmp_path / 'dvclive'))
+    hook = DvcliveLoggerHook()
```
zhouzaida, line 1111: suggest to add an unittest to check whether `model_file` param works well

zhouzaida:
hi @daavoo , CI failed

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -11,48 +13,62 @@ class DvcliveLoggerHook(LoggerHook):
...
         ignore_last (bool): Ignore the log of last iterations in each epoch
             if less than `interval`.
             Default: True.
         reset_flag (bool): Whether to clear the output buffer after logging.
-            Default: True.
+            Default: False.
         by_epoch (bool): Whether EpochBasedRunner is used.
             Default: True.
+        kwargs:
+            Arguments for instantiating `Live_`
```
zhouzaida, line 30:
```suggestion
            Arguments for instantiating `Live`_
```

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -11,48 +13,62 @@ class DvcliveLoggerHook(LoggerHook):
     It requires `dvclive`_ to be installed.
 
     Args:
-        path (str): Directory where dvclive will write TSV log files.
+        model_file (str):
```
zhouzaida, line 16: The purpose of this parameter is to save the latest checkpoint file, so isn't it possible to have a more intuitive name for it.

daavoo: It's a little tricky. 

The purpose of the parameter is to save the model at the end of each epoch. This is intended to be combined with a [DVC-specific feature](https://dvc.org/doc/user-guide/experiment-management/checkpoints)  called `checkpoints` (unfortunately a very reused term 😅). DVC expects MMCV (or other frameworks) to overwrite `model_file` at the end of each epoch.

It is necessary to use a specific string (it might not be the same name used for enabling MMCV checkpoints).

For users not using `DVC checkpoints`, they can just omit the parameter and use this integration just for logging (and the `MMCV Checkpoint` callback if they want to save the model)

zhouzaida: Got it.

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -11,48 +13,62 @@ class DvcliveLoggerHook(LoggerHook):
...
+
+    .. _Live:
+        https://dvc.org/doc/dvclive/api-reference/live#parameters
     """
 
     def __init__(self,
-                 path,
-                 interval=10,
+                 model_file=None,
+                 interval=1,
```
zhouzaida, line 41: Why do we need to set the interval as 1 default? BTW, the value in docstring is still 10.

daavoo: Nice catch. there is no need to set as 1

zhouzaida: got it.

zhouzaida (approved):

mmcv/runner/hooks/logger/dvclive.py
```diff
@@ -11,48 +13,62 @@ class DvcliveLoggerHook(LoggerHook):
...
 
     .. _dvclive:
         https://dvc.org/doc/dvclive
+
+    .. _Live:
+        https://dvc.org/doc/dvclive/api-reference/live#parameters
     """
 
     def __init__(self,
-                 path,
```
ZwwWayne, line 30: Does this mean that we have deprecated the argument `path`? If so, should we raise a warning and handle that?

PR #1213, opened 2021-07-22
[Fix] Fix the conflict between train info and evaluation info when evaluation iter

### LINKED ISSUE #1058: Evaluation results logging is inconsistent
**Describe the Issue**

Metrics calculated by evaluation hook are sometimes logged as `train/{metric_name}` and sometimes as `val/{metric_name}`.

More precisely, imagine that you have evaluation interval equal to 250 iterations and logging interval equal to 20 iterations. On the 250th iteration you will have your evaluation results logged as `val/{metric_name}`. After that, on the 500th iteration both train loss logging and evaluation will occur and you will have your evaluation results logged as `train/{metric_name}`.

It is extremely frustrating, especially if you use tensorboard logger that builds you two different charts for `train/{metric_name}` and `val/{metric_name}`.

**Bug fix**

This issue is caused by this line of code: https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/logger/base.py#L61

When training iteration is logged, `time` is in the buffer. Otherwise it's not.
This LOC comes from [this PR](https://github.com/open-mmlab/mmcv/pull/605/files) that applied `get_mode` method everywhere despite the fact that previously it was only used in the text logger which logs the mode for the whole iteration rather than for each data point separately. Some of the issues caused by that PR were already [fixed](https://github.com/open-mmlab/mmcv/commit/987cb58341878538a86c584757a49539bb1caab6#diff-6019bb677825d4e57b91906b6fc2d8e9306586ed68e45f4b176d77867c06ab30).

Before that pull request was merged, evaluation metrics were always logged with `train` tag. If such behavior is acceptable, I am willing to create a PR.

However, if you think that we need to always log evaluation results with `val` tag, it will require a lot of redesigning, because we will either need to create separate hook methods for evaluation or make EvalHook explicitly set `val` mode and flush the logger in the same manner as `val(...)` method of runner currently does.

sennnnn:
This PR try to resolve #1058.

## Problem Decription

1. When evaluation iter, the mode of log info is still `train`.

This is caused by mmcv/runner/hooks/logger/base.py Line 61:

```python
...
if 'time' in runner.log_buffer.output:
...
```

If the runner.log_buffer isn't clear out before EvalHook/DistEvalHook.evaluate(), the evaluation log info will be divided to train mode. Using `time` attribute to judge the mode of log info may not be strict.

2. When evaluation iter, the train info will be discarded.


zhouzaida:
closed by #1252

PR #1217, opened 2021-07-26
Fix potential bugs of basemodule when record the initilization information

jshilong:
## Motivation
1. There is a rare case,  `init_weights` may be called twice in a `top_level_module`, which would cause the error in 
https://github.com/open-mmlab/mmcv/blob/88d8c9ef1dd2ad763d4d2ed47077062f27d36223/mmcv/runner/base_module.py#L87
because the  `_params_init_info` has been deleted in the first time calling the `init_weights` of a `top_level_module`
2. The `from mmcv.runner.base_module import update_init_info` in  `mmcv/cnn/utils/weight_init.py` may cause recursive import.
3. Some users may adopt this way to do the initialization.
```
    def init_weights(self, prior_prob=1e-2):
        for conv in self.cls_preds:
            b = conv.bias.view(self.n_anchors, -1)
            b.data.fill_(-math.log((1 - prior_prob) / prior_prob))
            conv.bias = torch.nn.Parameter(b.view(-1), requires_grad=True)
```
They replace the original parameter with an initialized parameter which would cause  `KeyError:'temp_mean_value'`   when calling the `update_init_info`.
4. When dump the initialized information into the file, traversing the`_params_init_info` is not reasonable, because the user may delete some `nn.Parameter` to the module. 
## Modification
1. Add `_params_init_info` when executing the `init_weights` instead of executing `__init__`, and check the if the module is `top_level_module` with `hasattr(self, '_params_init_info')` instead of `len(self._params_init_info) == 0`
2. Move the `update_init_info` from `base_module.py` to `weight_init.py`, and import it in the runtime of `init_weights` in BaseModule
3. Add the check for whether the param is in the `_params_init_info`. 
4. When dump to the initialized information to the file, change to traverse the `named_parameters` of the model instead of `_params_init_info` 
5. Add the shape of the parameter to the logger.

The logger information would look like
```
neck.fpn_convs.1.conv.weight - torch.Size([256, 256, 3, 3]): 
XavierInit: gain=1, distribution=uniform, bias=0 

neck.fpn_convs.1.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

neck.fpn_convs.2.conv.weight - torch.Size([256, 256, 3, 3]): 
XavierInit: gain=1, distribution=uniform, bias=0 

neck.fpn_convs.2.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

neck.fpn_convs.3.conv.weight - torch.Size([256, 2048, 3, 3]): 
XavierInit: gain=1, distribution=uniform, bias=0 

neck.fpn_convs.3.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

neck.fpn_convs.4.conv.weight - torch.Size([256, 256, 3, 3]): 
XavierInit: gain=1, distribution=uniform, bias=0 

neck.fpn_convs.4.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.cls_convs.0.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.cls_convs.0.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.cls_convs.1.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.cls_convs.1.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.cls_convs.2.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.cls_convs.2.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.cls_convs.3.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.cls_convs.3.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.reg_convs.0.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.reg_convs.0.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.reg_convs.1.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.reg_convs.1.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.reg_convs.2.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.reg_convs.2.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.reg_convs.3.conv.weight - torch.Size([256, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.reg_convs.3.conv.bias - torch.Size([256]): 
The value is the same before and after calling `init_weights` of RetinaNet  

bbox_head.retina_cls.weight - torch.Size([720, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=-4.59511985013459 

bbox_head.retina_cls.bias - torch.Size([720]): 
NormalInit: mean=0, std=0.01, bias=-4.59511985013459 

bbox_head.retina_reg.weight - torch.Size([36, 256, 3, 3]): 
NormalInit: mean=0, std=0.01, bias=0 

bbox_head.retina_reg.bias - torch.Size([36]): 
NormalInit: mean=0, std=0.01, bias=0 
```


mmcv/runner/base_module.py
```diff
@@ -83,8 +72,22 @@ def is_init(self):
...
+            # - param_name (str): The name of parameter.
+            # - init_info (str): The string that describes the initialization.
+            # - tmp_mean_value (FloatTensor): The mean of the parameter,
+            #       which indicates whether the parameter has been modified.
+            # this attribute would be deleted after all parameters
+            # is initialized.
+            self._params_init_info = defaultdict(dict)
+            is_top_level_module = True
+
         if is_top_level_module:
```
MeowZheng, line 91: I think these two `if` might be merged into one.

MeowZheng (approved):
LGTM

nbei:
> ## Motivation
> 1. There is a rare case,  `init_weights` may be called twice in a `top_level_module`, which would cause the error in
>    https://github.com/open-mmlab/mmcv/blob/88d8c9ef1dd2ad763d4d2ed47077062f27d36223/mmcv/runner/base_module.py#L87
>    
>    because the  `_params_init_info` has been deleted in the first time calling the `init_weights` of a `top_level_module`
> 2. The `from mmcv.runner.base_module import update_init_info` in  `mmcv/cnn/utils/weight_init.py` may cause recursive import.
> 
> ## Modification
> 1. Add `_params_init_info` when executing the `init_weights` instead of executing `__init__`, and check the if the module is `top_level_module` with `hasattr(self, '_params_init_info')` instead of `len(self._params_init_info) == 0`
> 2. Bind the `update_init_info` as a method of `nn.Module` when calling the `init_weights`

About the modification 2, is it `nn.Module` or `mmcv.BaseModule`?

mmcv/cnn/utils/weight_init.py
```diff
@@ -200,8 +199,8 @@ def init(m):
                     xavier_init(m, self.gain, self.bias, self.distribution)
 
         module.apply(init)
-        if hasattr(module, '_params_init_info'):
-            update_init_info(module, init_info=self._get_init_info())
+        if hasattr(module, '_update_init_info'):
```
nbei, line 202: Currently, the logic here is much different from the early version. In this version, you check whether the `module` contains `_update_init_info` function. Will this behave differently from the last version?

jshilong: In this implementation, the `_update_init_info` and `_params_init_info` always appear or be deleted at the same time, and I check the existence of `_params_init_info` in `_update_init_info`. I believe it can keep the same behavior

nbei (approved):

jshilong:
> > ## Motivation
> > 
> > 1. There is a rare case,  `init_weights` may be called twice in a `top_level_module`, which would cause the error in
> >    https://github.com/open-mmlab/mmcv/blob/88d8c9ef1dd2ad763d4d2ed47077062f27d36223/mmcv/runner/base_module.py#L87
> >    
> >    because the  `_params_init_info` has been deleted in the first time calling the `init_weights` of a `top_level_module`
> > 2. The `from mmcv.runner.base_module import update_init_info` in  `mmcv/cnn/utils/weight_init.py` may cause recursive import.
> > 
> > ## Modification
> > 
> > 1. Add `_params_init_info` when executing the `init_weights` instead of executing `__init__`, and check the if the module is `top_level_module` with `hasattr(self, '_params_init_info')` instead of `len(self._params_init_info) == 0`
> > 2. Bind the `update_init_info` as a method of `nn.Module` when calling the `init_weights`
> 
> About the modification 2, is it `nn.Module` or `mmcv.BaseModule`?

sure, it is `nn.Module` instead of `BaseModule`

jshilong:
According to the comments from @nbei, binding the method `update_init_info`  to `nn.Module` is not reasonable, I change to import `update_init_info` in `init_weights` to avoid recursively import

MeowZheng (approved):
LGTM

tests/test_runner/test_basemodule.py
```diff
@@ -143,6 +143,80 @@ def __init__(self, init_cfg=None):
...
+    model.module2.init_weights()
+
+    assert not hasattr(model, '_params_init_info')
+    model.init_weights()
+    # assert `_params_init_info` would be deleted after `init_weights`
+    assert not hasattr(model, '_params_init_info')
+    # assert initialization information has been dumped
+    assert os.path.exists(log_file)
+
+    with open(log_file) as f:
```
ZwwWayne, line 210: mmcv.list_from_file()

mmcv/cnn/utils/weight_init.py
```diff
@@ -8,12 +8,38 @@
...
+            attribute `_params_init_info` which records the initialization
+            information.
+        init_info (str): The string that describes the initialization.
+    """
+    assert hasattr(
+        module,
+        '_params_init_info'), f'Can not find `_params_init_info` in {module}'
+    for name, param in module.named_parameters():
+        mean_value = param.data.mean()
+
```
ZwwWayne, line 31: Suggest showing both mean and std as std sometimes are more important and set

mmcv/cnn/utils/weight_init.py
```diff
@@ -8,12 +8,38 @@
...
+        if param not in module._params_init_info:
+            new_param_info = f'Initialized by user-defined ' \
+                             f'`init_weights` in {module.__class__.__name__} '
+            module._params_init_info[param]['tmp_mean_value'] = mean_value
+            module._params_init_info[param]['init_info'] = new_param_info
+
+        if module._params_init_info[param]['tmp_mean_value'] != mean_value:
```
ZwwWayne, line 38: Why we will meet this case? Should we do something more to deal with this case?

ZwwWayne: Might leave a comment here to explain that.

nbei (approved):

Issue #1219, opened 2021-07-27
pip install latest version is 1.3.9


canghaiyunfan:
pip install mmcv-full, installed version is 1.3.9 not 1.3.10

zhouzaida:
hi @canghaiyunfan , thanks for your feedback. I will fix it

zhouzaida:
mmcv-full v1.3.10 has been published

PR #1220, opened 2021-07-27
[Fix] Fix CI of publishing package

zhouzaida:

## Motivation

Related Issue: #1219
Related Action: https://github.com/open-mmlab/mmcv/runs/3150536246?check_suite_focus=true#step:4:17

Fix CI of publishing package

## Modification

- .github/workflows/build.yml
- .github/workflows/publish-to-pypi.yml

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install system dependencies
         run: sudo apt-get update && sudo apt-get install -y ffmpeg libturbojpeg
       - name: Upgrade Setuptools
-        run: pip install setuptools==52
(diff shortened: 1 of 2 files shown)


ZwwWayne:
Do we also need to update our requirements?

zhouzaida:
> Do we also need to update our requirements?

In fact, the requirement has been updated before but it didn't work because the error is raised before running `python setup.py develop`. Maybe this line can be removed.

https://github.com/open-mmlab/mmcv/blob/88d8c9ef1dd2ad763d4d2ed47077062f27d36223/requirements/runtime.txt#L6

PR #1227, opened 2021-07-30
[Fix] Loose the version requirement of setuptools

kennymckormick:


zhouzaida:
Maybe we should remove the dependency of setuptools.
In fact, the dependency wrote in the requirement will not work well because we import it in the setup.py.

Related issue: #1230

ZwwWayne:
Make sure the modification has been tested.

PR #1235, opened 2021-08-02
[Fix] Do not depend on setuptools>=52

kennymckormick:


setup.py
```diff
@@ -1,9 +1,11 @@
 import glob
 import os
 import re
-from pkg_resources import DistributionNotFound, get_distribution, parse_version
+from pkg_resources import DistributionNotFound, get_distribution
 from setuptools import find_packages, setup
 
+from packaging.version import parse
```
zhouzaida, line 7:
```suggestion
from packaging.version import parse as parse_version
```

innerlee:
See concerns here https://github.com/open-mmlab/mmcv/pull/1185#discussion_r669379489

kennymckormick:
@zhouzaida  CI passed

mmcv/version.py
```diff
@@ -32,6 +30,6 @@ def parse_version_info(version_str: str, length: int = 4) -> tuple:
     return tuple(release)
 
 
-version_info = parse_version_info(__version__)
+version_info = tuple([int(x) for x in __version__.split('.')])
```
zhouzaida, line 33:
```suggestion
version_info = tuple(int(x) for x in __version__.split('.'))
```
In addition, if `__version__` contains non-numeric string like `1.3.10.dev`, `int(x)` will throw an error

ZwwWayne (approved):

mmcv/version.py
```diff
@@ -32,6 +30,6 @@ def parse_version_info(version_str: str, length: int = 4) -> tuple:
     return tuple(release)
 
 
-version_info = parse_version_info(__version__)
+version_info = tuple(int(x) for x in __version__.split('.')[:3])
```
ZwwWayne, line 33: Why do not use parse_version_info here though you wrote it?

kennymckormick: When you execute this line during setup, "packaging" may not exist.

Issue #1248, opened 2021-08-07
Log for val results does not save correctly when `IterBasedRunner` is used.


haotian-liu:
When I am using `IterBasedRunner` is used, and the evaluation runs during the training (triggered by evaluation hook `evaluation=dict(interval=5000)`), the eval results are not logged with val mode, and are logged as train mode. This is different from `EpochBasedRunner`.

Specifically, for `EpochBasedRunner`, training losses will be logged in Tensorboard with prefix `train/` (e.g., train/loss, train/loss_cls, etc.), and evaluation results will be logged with prefix `val/` (e.g., val/table, val/acc, val/miou, etc.). However, in `IterBasedRunner`, they are all logged with prefix `train/` (e.g., train/loss, train/table, train/acc, train/miou), which is weird and causes troubles for visualization.

It seems that when evaluation hook is triggered in `after_iter` stage, the mode will not be changed to `eval` and remains as `train` instead, which might be the cause of the issue.

Please kindly advise if there is anything that I am not doing correctly, thank you so much!

zhouzaida:
hi @haotian-liu , thanks for you feedback, we are fixing the bug

haotian-liu:
Thank you so much!

zhouzaida:
closed by #1252

PR #1252, opened 2021-08-08
[Fix] Fix the bug that the training log and evaluating log are mixed

zhouzaida:

## Motivation

Fix the bug that the training log and evaluating log are mixed.

Because the priority of EvalHook is higher than LoggerHook, the training log and the evaluating log are mixed. Therefore, we need to dump the training log and clear it before evaluating log is generated. In addition, this problem will only appear in `IterBasedRunner`, because `EpochBasedRunner` calls `_do_evaluate` in `after_train_epoch` stage, and at this stage the log has been printed, so it will not cause any problem.

more details
- #1248 
- https://github.com/open-mmlab/mmsegmentation/issues/694

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmcv/runner/hooks/evaluation.py
+++ b/mmcv/runner/hooks/evaluation.py
@@ -455,6 +455,11 @@ class DistEvalHook(EvalHook):
             # the log has been printed, so it will not cause any problem.
             # more details at
             # https://github.com/open-mmlab/mmsegmentation/issues/694
+            if not self.by_epoch:
+                for hook in runner._hooks:
+                    if isinstance(hook, LoggerHook):
+                        hook.after_train_iter(runner)
+
             print('\n')
             runner.log_buffer.output['eval_iter_num'] = len(self.dataloader)
             key_score = self.evaluate(runner, results)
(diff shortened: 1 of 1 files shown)


mmcv/runner/hooks/evaluation.py
```diff
@@ -211,19 +212,31 @@ def before_train_epoch(self, runner):
...
+            # `IterBasedRunner` whose `self.by_epoch` is False, because
+            # `EpochBasedRunner` whose `self.by_epoch` is True calls
+            # `_do_evaluate` in `after_train_epoch` stage, and at this stage
+            # the training log has been printed, so it will not cause any
+            # problem. more details at
+            # https://github.com/open-mmlab/mmsegmentation/issues/694
+            for hook in runner._hooks:
+                if isinstance(hook, LoggerHook):
+                    hook.after_train_iter(runner)
+                    runner.log_buffer.clear()
```
Junjun2016, line 229:
```suggestion
                runner.log_buffer.clear()
```

Junjun2016: Out for loop.

ZwwWayne (approved):

dreamerlin (approved):

PR #1253, opened 2021-08-10
[Feature] Add revert_sync_batchnorm

gaotongxiao:
## Motivation

A model that uses SyncBN cannot support CPU inference. SyncBN can also cause some other issues during inference. We introduce `revert_sync_batchnorm` from @kapily's work (https://github.com/pytorch/pytorch/issues/41081#issuecomment-783961547), which can convert SyncBN in any model to BN. 

## Modification

Added `revert_sync_batchnorm` to `mmcv/cnn/utils/sync_bn.py` and its unittest.

## BC-breaking (Optional)

No, but there could be a potential minor risk -

PyTorch provides an implementation of convert_sync_batchnorm which converts BatchNorm1D, BatchNorm2D and BatchNorm3D to SyncBatchNorm. However, it doesn't provide an inverse function for that. The reason is SyncBatchNorm neither has a strict input dimension checking nor stores the expected input dimension, whereas BatchNormxD strictly validates the input dimension (and this is the only difference between BatchNorm1D, 2D, and 3D). **Therefore, if one converts BNxD to SyncBN using PyTorch's implementation and then converts it back to BN using this implementation, the input dimension check is no longer retained.**


ZwwWayne:
Please also consider/check/test MMSyncBN

gaotongxiao:
@ZwwWayne It seems `MMSyncBN` even cannot be initialized in a non-distributed environment.

ZwwWayne:
> @ZwwWayne It seems `MMSyncBN` even cannot be initialized in a non-distributed environment.

So it is also necessary to convert it if it is used in the config for training?

gaotongxiao:
@ZwwWayne A model with `MMSyncBN` layer cannot be built in a non-distributed environment. I think the only solution is to re-generate a `Config` replacing `MMSyncBN` with `BN` and use it to build the model, but the idea can be very different from the current implementation. I'm also not sure if the checkpoint is still compatible after such a conversion. BTW, what was the goal of `MMSyncBN`?

ZwwWayne:
> @ZwwWayne A model with `MMSyncBN` layer cannot be built in a non-distributed environment. I think the only solution is to re-generate a `Config` replacing `MMSyncBN` with `BN` and use it to build the model, but the idea can be very different from the current implementation. I'm also not sure if the checkpoint is still compatible after such a conversion. BTW, what was the goal of `MMSyncBN`?

It was written a long time ago when PyTorch does not provide SyncBN. It also supports some corner case that currently is not supported in PyTorch.

gaotongxiao:
@ZwwWayne Do we have any examples on using MMSyncBN?

ZwwWayne:
> @ZwwWayne Do we have any examples using MMSyncBN?

Simply change the SyncBN in the norm_cfg to MMSyncBN should work

tests/test_cnn/test_sync_bn.py
```diff
@@ -0,0 +1,15 @@
...
+import torch
+
+from mmcv.cnn.bricks import ConvModule
+from mmcv.cnn.utils import revert_sync_batchnorm
+
+
+def test_revert_sync_batchnorm():
+    conv = ConvModule(3, 8, 2, norm_cfg=dict(type='SyncBN'))
+    x = torch.randn(1, 3, 10, 10)
+    with pytest.raises(ValueError):
```
ZwwWayne, line 11: Add comment to indicate what you are testing above this line

ZwwWayne:
I suggest implementing a function to simply modify the config to change SyncBN/MMSyncBN back to BN. It works for both SyncBN and MMSyncBN and meets the demands that this PR wants to achieve. It is a little bit late to convert the model back after build SyncBN layer.

gaotongxiao:
@ZwwWayne Done

mmcv/cnn/utils/sync_bn.py
```diff
@@ -0,0 +1,56 @@
...
+                                     module.track_running_stats)
+        if module.affine:
+            with torch.no_grad():
+                module_output.weight = module.weight
+                module_output.bias = module.bias
+        module_output.running_mean = module.running_mean
+        module_output.running_var = module.running_var
+        module_output.num_batches_tracked = module.num_batches_tracked
+        module_output.training = module.training
+        if hasattr(module, 'qconfig'):
```
ZwwWayne, line 51: Need a comment here to indicate what is `qconfig`.

ZwwWayne (approved):

mmcv/cnn/utils/sync_bn.py
```diff
@@ -0,0 +1,57 @@
...
+        module (nn.Module): The module containing `SyncBatchNorm` layers.
+
+    Returns:
+        module_output: The converted module with `BatchNormXd` layers.
+    """
+    module_output = module
+    module_checklist = [torch.nn.modules.batchnorm.SyncBatchNorm]
+    if hasattr(mmcv, 'ops'):
+        module_checklist.append(mmcv.ops.SyncBatchNorm)
+    if isinstance(module, tuple(module_checklist)):
```
ZwwWayne, line 39: `tuple(module_checklist)` might be unnecessary here and we may directly use `module_checklist`

mmcv/cnn/utils/sync_bn.py
```diff
@@ -0,0 +1,57 @@
...
+        module_output = _BatchNormXd(module.num_features, module.eps,
+                                     module.momentum, module.affine,
+                                     module.track_running_stats)
+        if module.affine:
+            with torch.no_grad():
```
ZwwWayne, line 44: Is torch.no_grad() here necessary?

tests/test_cnn/test_revert_syncbn.py
```diff
@@ -0,0 +1,59 @@
...
+    with pytest.raises(ValueError):
+        y = conv(x)
+    conv = revert_sync_batchnorm(conv)
+    y = conv(x)
+    assert y.shape == (1, 8, 9, 9)
+
+
+def test_revert_mmsyncbn():
+    if 'SLURM_NTASKS' not in os.environ or int(
+            os.environ['SLURM_NTASKS']) != 4:
```
ZwwWayne, line 31: Are 4 tasks necessary here? Can one or two task make it work?

mmcv/cnn/utils/sync_bn.py
gaotongxiao: `isinstance` only accepts types or a tuple of types

mmcv/cnn/utils/sync_bn.py
gaotongxiao: The wrapped assignments do not involve any PyTorch ops and thus this line seems unnecessary (at least that's what my local experiment shows). But let's be more conservative and consistent with its official counterpart here. (https://github.com/pytorch/pytorch/blob/26b7ff5aeab49cb63faed2cd51fa6ba70f665610/torch/nn/modules/batchnorm.py#L807-L809)

ZwwWayne: I see, preferably refer to the official counterpart under/above this line as a reminder.

tests/test_cnn/test_revert_syncbn.py
ZwwWayne: We can make it as 2 to reduce the test cost. Originally set it as 4 for MMSyncBN is to test its behavior with different numbers of groups like 1, 2, 4. But they are unnecessary here.

ZwwWayne (approved):

PR #1254, opened 2021-08-11
[Enhance] Add onnx namespace for custom ops

grimoire:

## Motivation

Custom ops can not be exported to onnx unless the node is in a non-default namespace. This PR adds namespace `mmcv` to the custom ops so they can be exported.
Note that the converted node might not be used in some backend until the implementation of the backend is added.

## Modification

Add `mmcv` namespace to custom ops, fix data type of the attributes.


hellock (approved):

ZJLi2013:
wonder what about other ops, e.g. [MultiScaleDeformaAttension ](https://github.com/open-mmlab/mmcv/blob/main/mmcv/ops/multi_scale_deform_attn.py) ?

PR #1256, opened 2021-08-11
[Feature] Add Normalize transform

zhouzaida:

## Motivation

Add Normalize transform.

Reference
- https://github.com/open-mmlab/mmclassification/blob/e4188c828c265daf863e4a6cfaffdecbccc4c7c3/mmcls/datasets/pipelines/transforms.py#L802
- https://github.com/open-mmlab/mmdetection/blob/b4a84cc9cf38ee4c0fa6f6264e9916c8ba7cd3b3/mmdet/datasets/pipelines/transforms.py#L656

## Modification

Add Normalize class and test_normalize function

## BC-breaking (Optional)

No

## Use cases (Optional)

refer to docstring

## Checklist

### DIFF
--- a/mmcv/datasets/pipelines/transforms.py
+++ b/mmcv/datasets/pipelines/transforms.py
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from numbers import Number
 from typing import Union
 
 import numpy as np
@@ -12,8 +13,8 @@ class Normalize:
     """Normalize the image.
 
     Args:
-        mean (sequence): Mean values of channels.
-        std (sequence): Std values of channels.
+        mean (numeric | sequence): Mean values of channels.
+        std (numeric |sequence): Std values of channels.
         to_rgb (bool): Whether to convert the image from BGR to RGB,
             default is true.
 
@@ -34,8 +35,8 @@ class Normalize:
     """
 
     def __init__(self,
-                 mean: Union[int, Sequence],
-                 std: Union[int, Sequence],
+                 mean: Union[Number, Sequence],
+                 std: Union[Number, Sequence],
                  to_rgb: bool = True):
         self.mean = np.array(mean, dtype=np.float32)
         self.std = np.array(std, dtype=np.float32)
--- a/mmcv/datasets/pipelines/transforms.py
+++ b/mmcv/datasets/pipelines/transforms.py
@@ -10,13 +10,31 @@ from ..builder import PIPELINES
 class Normalize:
     """Normalize the image.
 
-    Added key is 'img_norm_cfg'.
+    `results` is the input of `__call__()`. The required key of `results` is
+    `img_fields`, which is a list of string. Every item of it is also the key
+    of `results`, whose value is an image. After invoking the `__call__()`, the
+    return results will add an another key `img_norm_cfg`.
 
     Args:
         mean (sequence): Mean values of 3 channels.
         std (sequence): Std values of 3 channels.
         to_rgb (bool): Whether to convert the image from BGR to RGB,
             default is true.
+
+    Example:
+        >>> from mmcv.image.io import imread
+        >>> from mmcv.datasets.pipelines import Normalize
+        >>> cfg = {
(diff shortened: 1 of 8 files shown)


mmcv/datasets/pipelines/transforms.py
```diff
@@ -0,0 +1,51 @@
...
+
+import mmcv
+from ..builder import PIPELINES
+
+
+@PIPELINES.register_module()
+class Normalize:
+    """Normalize the image.
+
+    Added key is 'img_norm_cfg'.
```
mzr1996, line 13: Should we add a docstring about "Required key" here?

zhouzaida: In fact, adding the docstring here will make the user confused. So I add it in __call__.

zhouzaida: Adding an example will be better

mmcv/datasets/pipelines/transforms.py
```diff
@@ -45,9 +40,12 @@ def __init__(self, mean: Sequence, std: Sequence, to_rgb: bool = True):
     def __call__(self, results: dict) -> dict:
         """Call function to normalize images.
 
+        The required key of `results` is `img_fields`, which is a list of
```
mzr1996, line 43: Single backquote show as italic, if you want to show as code, need to use double backquotes. Refers to [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup)

zhouzaida: got it

mmcv/datasets/pipelines/transforms.py
```diff
@@ -11,8 +12,8 @@ class Normalize:
     """Normalize the image.
 
     Args:
-        mean (sequence): Mean values of 3 channels.
-        std (sequence): Std values of 3 channels.
+        mean (sequence): Mean values of channels.
+        std (sequence): Std values of channels.
```
mzr1996, line 16: Since we accept `Union[int, Sequence]`, please change the docstring as well.

mmcv/datasets/pipelines/transforms.py
```diff
@@ -32,7 +33,10 @@ class Normalize:
         >>> results = normalize(results)
     """
 
-    def __init__(self, mean: Sequence, std: Sequence, to_rgb: bool = True):
+    def __init__(self,
+                 mean: Union[int, Sequence],
+                 std: Union[int, Sequence],
```
mzr1996, line 38: Not only `int` but also `float`.
```suggestion
from numbers import Number
...
    def __init__(self,
                 mean: Union[Number, Sequence],
                 std: Union[Number, Sequence],
```

mzr1996 (changes requested):

tests/test_datasets/test_pipelines/test_transform/test_transform.py
```diff
@@ -0,0 +1,49 @@
...
+
+    results['img_fields'] = ['img', 'img2']
+    results = transform(results)
+    assert np.equal(results['img'], results['img2']).all()
+
+    mean = np.array(img_norm_cfg['mean'])
+    std = np.array(img_norm_cfg['std'])
+    if to_rgb:
+        original_img = original_img[..., ::-1]
+    converted_img = (original_img - mean) / std
```
mzr1996, line 48: Here is an implicit type conversion, it works properly, but maybe making it explicit is better.
```suggestion
    converted_img = (original_img.astype(np.float32) - mean) / std
```
Otherwise, if `mean` is an integer, it will go wrong.

mzr1996 (changes requested):

tests/test_datasets/test_pipelines/test_transform/test_transform.py
zhouzaida: done

mzr1996 (approved):
LGTM

PR #1267, opened 2021-08-13
[Fix] Restrict the warning message

yyz561:

## Motivation

When I implemented a CenterNet2 model in mmdet, there was a situation when I had to use GroupNorm after Conv2d, and the Conv2d layer was supposed to have a bias. Once I finished the modeling part, I got a warning message 'ConvModule has norm and bias at the same time'. At first, I was not so sure about this warning, it made me uneased. Then I did a little research and realized that for BatchNorm (not so sure about SyncBN), the bias does not affect the calculation. But for other kinds of Norm, e.g., GN, I am pretty sure the bias does give some influence. I think maybe we should restrict the popping of this warning message, since it can be quite misleading for new starters.

## Modification

Restrict a warning message of 'ConvModule has norm and bias at the same time' to BN, BN1, BN2 and BN3 normalization.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmcv/cnn/bricks/conv_module.py
```diff
@@ -105,7 +105,8 @@ def __init__(self,
         self.with_bias = bias
 
         if self.with_norm and self.with_bias:
-            warnings.warn('ConvModule has norm and bias at the same time')
+            if self.conv_cfg.get('type') in ['BN', 'BN1d', 'BN2d', 'BN3d']:
```
zhouzaida, line 108: Can we change the `self.conv_cfg.get('type')` to `isinstance`

mmcv/cnn/bricks/conv_module.py
```diff
@@ -105,7 +105,8 @@ def __init__(self,
...
+                warnings.warn('ConvModule has norm and bias at the same time')
```
zhouzaida, line 109:
```suggestion
                warnings.warn('ConvModule has batch norm and bias at the same time')
```

zhouzaida:
> [CLA assistant check](https://cla-assistant.io/open-mmlab/mmcv?pullRequest=1267)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/open-mmlab/mmcv?pullRequest=1267) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/open-mmlab/mmcv?pullRequest=1267) it.

hi @yyz561 , please sight the CLA

mmcv/cnn/bricks/conv_module.py
yyz561: Could you please give the exact code you want to use?  I think we can only access the type string at this point, not an instance of norm ?

mmcv/cnn/bricks/conv_module.py
yyz561: Sure, good suggestion.

mmcv/cnn/bricks/conv_module.py
zhouzaida: yet, I miss the type of self.conv_cfg

yyz561: However, I do have a more elegant way, haha, push finished.

zhouzaida:
hi, please fix the CI

yyz561:
> [CLA assistant check](https://cla-assistant.io/open-mmlab/mmcv?pullRequest=1267)
> All committers have signed the CLA.

@zhouzaida do you mean this?

yyz561:
> hi, please fix the CI



> hi, please fix the CI

sorry, I am not very familiar with all these processes. Do ou mean Check List?

yyz561:
> hi, please fix the CI

Soryy, didn't noticed that code too long, will fix it immediately.

yyz561:
> hi, please fix the CI

It seems that the password is not valid for git push since yesterday. It took me some time to solve the problem, sorry for the delay. It should work now.

zhouzaida:
> > hi, please fix the CI
> 
> It seems that the password is not valid for git push since yesterday. It took me some time to solve the problem, sorry for the delay. It should work now.

yet, thanks for your contribution

mmcv/cnn/bricks/conv_module.py
```diff
@@ -105,7 +105,9 @@ def __init__(self,
         self.with_bias = bias
 
         if self.with_norm and self.with_bias:
-            warnings.warn('ConvModule has norm and bias at the same time')
+            if self.conv_cfg.get('type').startswith('BN'):
```
zhouzaida, line 108: Should we replace the `self.conv_cfg` with `self.norm_cfg` ?
In addition, if the user extends from BN, the condition is not enough robust.
```python
NORM_LAYERS.register_module()
class DummyBN(bb.BatchNorm2d):
     pass
```

yyz561: Sorry, will fix the spelling mistake asap. You are right  about the robustness problem. If possible, the isinstance method would be a better choice, but I could  not find the right place.

zhouzaida: Maybe we can use `issublclass()`
```python
obj_type = self.norm_cfg.get('type')
obj_cls = registry.get(obj_type)
if any(issubclass(obj_cls, norm_cls) for norm_cls in [nn.BatchNorm1d, ...]):
    warnings.warn('ConvModule ...')
```

yyz561: Got it. This is actually beyond my knowledge, thanks.

zhouzaida:
we also need to add a unittest

yyz561:
> we also need to add a unittest

will this work? I didn't 

> we also need to add a unittest

Sorry mate, my first time to write a unit test, will this work?

zhouzaida:
> > we also need to add a unittest
> 
> will this work? I didn't
> 
> > we also need to add a unittest
> 
> Sorry mate, my first time to write a unit test, will this work?

I think it will work

yyz561:
> > > we also need to add a unittest
> > 
> > 
> > will this work? I didn't
> > > we also need to add a unittest
> > 
> > 
> > Sorry mate, my first time to write a unit test, will this work?
> 
> I think it will work

I am exhausted by the format issue... both here and in another PR, could you please name a few external tools I can use? Thanks a lot.

zhouzaida:
hi, you can refer to https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md.

mmcv/cnn/bricks/conv_module.py
```diff
@@ -105,7 +105,11 @@ def __init__(self,
         self.with_bias = bias
 
         if self.with_norm and self.with_bias:
-            warnings.warn('ConvModule has norm and bias at the same time')
+            check_content = NORM_LAYERS.get(self.norm_cfg.get('type'))
```
zhouzaida, line 108: maybe we can check here to simplify the code
https://github.com/open-mmlab/mmcv/blob/ac0d8398bfd5d513848f37f20bb75e5cd25bc5c3/mmcv/cnn/bricks/conv_module.py#L148

For example
```python
        # build normalization layers
        if self.with_norm:
            # norm layer is after conv layer
            if order.index('norm') > order.index('conv'):
                norm_channels = out_channels
            else:
                norm_channels = in_channels
            self.norm_name, norm = build_norm_layer(norm_cfg, norm_channels)
            if self.with_bias and issubclass(norm, check_range):
                ...
            self.add_module(self.norm_name, norm)
```

yyz561: btw, do I need to remove the ruby source for windows os from CONTRIBUTING.md ?

zhouzaida: yet, you can create another PR?

zhouzaida: Can `check_range` be renamed to `batch_norm_list`?

yyz561: > yet, you can create another PR?

I don't think it worth another PR, hahaha. Thanks  you very much, the pre-commit tool saves me a lot of time.

zhouzaida (approved):
LGTM

mmcv/cnn/bricks/conv_module.py
zhouzaida: haha, thanks for your contribution.

mmcv/cnn/bricks/conv_module.py
```diff
@@ -147,6 +144,12 @@ def __init__(self,
                 norm_channels = in_channels
             self.norm_name, norm = build_norm_layer(norm_cfg, norm_channels)
             self.add_module(self.norm_name, norm)
+            if self.with_bias:
+                batch_norm_list = (nn.BatchNorm1d, nn.BatchNorm2d,
+                                   nn.BatchNorm3d)
+                if issubclass(norm, batch_norm_list):
```
zhouzaida, line 150:
```python
from mmcv.utils import TORCH_VERSION, _BatchNorm
if issubclass(norm, _BatchNorm):
```

yyz561: I noticed that SyncBatchNorm is also inherited from _BatchNorm, I guess the bias is applied to minibatch, while the norm applies to the bigger batch, so ?

zhouzaida: I guess the bias has no effect on batchnorm and at the same time the synchronization of batchnorm has not relation with bias.

yyz561: Here is my point: we know that after conv, you have options to use bias inside the conv or not, which is the last operation of conv module. 
Next, if you use BatchNorm, the previous bias in conv would be cancelled out during norm operation. So we want to give warnings that these kinds of BN and conv bias combination is unnecessary.
As for GN or SyncBatchNorm or others, you can't cancel out the conv bias. For GN, the conv bias is applied to a batch, but norms are applied separately to feature groups. Similarly, for SyncBatchNorm, conv bias is applied to local minibatch, while norm is based on the sum of all mini batches. I guess the the bias inside conv would affect the outputs after SyncBatchNorm. 
Back to my point, I think the warning message is intent to remind us, "Hey, mate, you are doing some meaningless operations" instead of "Woops, you are using bias in conv, which followed by batch norm"

zhouzaida: We can see the implementation of SyncBN here(https://github.com/chrisway613/Synchronized-BatchNormalization/blob/master/sync_bn.py#L55).  Master rank will collect the sum and square-sum from other ranks . Master rank first calculates the mean and variance and then broadcast to others. Anything is the same as BN, so the bias has no effect on SycnBN. BTW, If I have a misunderstanding of SycnBN, please let me know

yyz561: Yeah, I am not very familiar with SycnBN, I will check the resource you mentioned, and perhaps do an experiment. Thank you.

yyz561: You are right. I was wrong to think every copy of the model on different GPUS are different from each other, and averaged at the end.

zhouzaida:
hi @yyz561 , after reading the GN formula, I found the bias of conv also has no effect for the output of GN

yyz561:
> hi @yyz561 , after reading the GN formula, I found the bias of conv also has no effect for the output of GN

Hey, I didn't read the actual coding, but it is easy to do a GN experience. Try the following code:

_import torch 
import torch.nn as nn
torch.manual_seed(0)
input = torch.randn(1, 4, 4, 4)
conv = nn.Conv2d(4, 4, 3, padding=1, bias=True)
norm = nn.GroupNorm(2, 4)
output = norm(conv(input))
print(output[0, 0, :, :].view(-1))_

for comparasion:

_import torch 
import torch.nn as nn
torch.manual_seed(0)
input = torch.randn(1, 4, 4, 4)
conv = nn.Conv2d(4, 4, 3, padding=1, bias=False)
norm = nn.GroupNorm(2, 4)
output = norm(conv(input))
print(output[0, 0, :, :].view(-1))_

yyz561:
Hey @zhouzaida, I wrote an blog to talk about the problem. At the same time, I realized that  bias of conv also has no effect for the output of Instance Norm too. Could you please have a look at the blog? https://zhuanlan.zhihu.com/p/403444336

zhouzaida:
> Hey @zhouzaida, I wrote an blog to talk about the problem. At the same time, I realized that bias of conv also has no effect for the output of Instance Norm too. Could you please have a look at the blog? https://zhuanlan.zhihu.com/p/403444336

yet, you are right. Thanks for your patience

mmcv/cnn/bricks/conv_module.py
```diff
@@ -147,6 +145,11 @@ def __init__(self,
                 norm_channels = in_channels
             self.norm_name, norm = build_norm_layer(norm_cfg, norm_channels)
             self.add_module(self.norm_name, norm)
+            if self.with_bias:
```
ZwwWayne, line 148: May add a comment here to indicate why we only report warning for BN/IN here.

yyz561: How about changing the warning message into "'Unnecessary conv bias before batch/instance norm'"

ZwwWayne (approved):

PR #1275, opened 2021-08-16
Support deprecation info in Config

ly015:
## Motivation

Support deprecation information in Config. Users can add a field `_deprecation_` into a config file, which will trigger a deprecation warning during calling `Config.fromfile`. Optionally, the warning information can include a path to the expected config file to replace the deprecated one.

## Modification

- Add deprecation information support in `config.py`
- Add unittest

## Use cases (Optional)

For example, if a config file `old_config.py` has been renamed to `new_config.py`. To avoid bc-breaking, we can create an `old_config.py` file with deprecation information as following:

```python
_base_ = './new_config.py'
_deprecation_ = dict(expected='configs/new_config.py')

```

### DIFF
--- a/docs/understand_mmcv/config.md
+++ b/docs/understand_mmcv/config.md
@@ -177,3 +177,24 @@ item1 = 'a'
 item2 = dict(item3='b')
 item = dict(a='a', b='b')
 ```
+
+### Add deprecation information in configs
+
+Deprecation information can be added into a config file, which will trigger a `UserWarning` when this config file is loaded.
+
+`deprecated_cfg.py`
+
+```python
+_base_ = 'expected_cfg.py'
+
+_deprecation_ = dict(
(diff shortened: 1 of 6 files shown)


ZwwWayne:
The modification LGTM. It would be better to also update the documentation thus the users could know it right after this PR.

innerlee:
It would be great to provide a link to the related pr

innerlee:
I'm not sure about the usecase. If a user uses the wrong config file in his/her own config, then how should he/she get the warning?

ly015:
> It would be great to provide a link to the related pr

You mean adding a key to the  `_deprecatoin_` dict in the config file, like `reference` or `url`?

innerlee:
Yes, the pr will be very helpful since its diff provides many live examples

innerlee (approved):

ZwwWayne (approved):

PR #1278, opened 2021-08-17
Add DCN and Modulated DCN CPU implementation

eugene123tw:
## Motivation

Add DCN and Modulated DCN CPU implementation to mmcv/ops

## Modification

Adapting modulated DCN and DCN implementation from CUDA to Cpp


innerlee:
Thanks!

zhouzaida:
hi @eugene123tw , please fix the CI. And you can refer to https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md to fix the lint.

eugene123tw:
@zhouzaida Thanks! I've submitted the change

mmcv/ops/csrc/pytorch/deform_conv_cpu.cpp
```diff
@@ -0,0 +1,724 @@
...
+                                   output.options());
+
+  output_buffer = output_buffer.view(
+      {output_buffer.size(0), group, output_buffer.size(1) / group,
+       output_buffer.size(2), output_buffer.size(3)});
+
+  for (int elt = 0; elt < batchSize / im2col_step; elt++)
+  {
+
+    deformable_im2col_cpu(input[elt], offset[elt], nInputPlane, inputHeight,
```
grimoire, line 457: If the only difference between CPU and GPU launcher is this kernel function, can we share other code and invoke different kernel depend on device type?

grimoire (commented):
I do not prefer put `deform_conv_utils.hpp` inside common cause it is pytorch only. Any advice @zhouzaida ?

mmcv/ops/csrc/pytorch/deform_conv_cpu.cpp
eugene123tw: @grimoire As far as I know, CPU/GPU functions can't be shared. Cuda kernel to specify __global__ and functions need to add __device__ specifier (CPU functions can skip this). ROI align is another example for both GPU/CPU support and is implemented in the same fashion for this reason I think.

grimoire: Errr, I means:
```c++
// modulated_deform_conv_forward.cpp

// declare cuda kernel wrap function here
void modulated_deformable_im2col_cuda();

// declare cpp implement here
void modulated_deformable_im2col_cpu();

void modulated_deform_conv_forward(){
....

// choice different function
  if(use_cuda){
    modulated_deformable_im2col_cuda();
  }else{
    modulated_deformable_im2col_cpu();
  }

....
}
```

You can keep `modulated_deformable_im2col_cuda` inside `modulated_deform_conv_cuda.cu` and invoke cuda kernel there.

eugene123tw: Calling 2 kernel implementations are quite different. Cuda specific syntaxes (e.g. GET_BLOCKS, at::cuda::getCurrentCUDAStream()) in the wrap function (e.g. `modulated_deformable_im2col_cuda`) should be defined in a `.cu` file. I don't think we can define both wrap functions in a cpp file. This is to ensure that each compiler (gcc, nvcc) takes care of files it knows best to compile.

grimoire: You can put the `declaration` in `cpp` and `definition` in `.cu`. Here is an example:
[link](https://drive.google.com/file/d/1JuDYHWihxrrH3JJjzpve7k1ZbtZdygUV/view?usp=sharing)

eugene123tw: Thanks, @grimoire. I've refactored DCN for reference. Let me know how it looks and I can refactor modulated DCN as well.

grimoire: Looks great!

eugene123tw: Modulated DCN is refactored accordingly as well

grimoire (approved):
LGTM if lint is fixed.

ZwwWayne (approved):

PR #1281, opened 2021-08-18
add modulated_deform_conv in onnxruntime support

PeiqiWang:
## Motivation
Run some model in mmocr (e.g. dbnet) using onnxruntime.

## Modification

Following the the mmcv guide, add `modulated_deform_conv.cpp`, `modulated_deform_conv.h`  and `test_onnx.py`.
The modulated_deform_conv op definition is following [mmcv-tensorrt-custom-op](https://github.com/open-mmlab/mmcv/blob/master/docs/deployment/tensorrt_custom_ops.md#mmcvmodulateddeformconv2d)

Because the input `bias` is optional and there are parameters in the type of `list of ints`, like `stride`, the onnxruntime needs to upgrade to 1.8.0+.  For myself, I use onnxruntime-1.8.1. Comparing to the onnxruntime-1.5.1, which is given in mmcv official guild, some function definitions are changed. Thus I also make some modifications in other custom op definations.

### DIFF
--- a/mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
+++ b/mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
@@ -187,18 +187,13 @@ void deformable_conv2d_ref_fp32(
           }
         }
       } else {
-        for (int64_t oc = 0; oc < oc_per_gp; ++oc) {
-          for (int64_t hw = 0; hw < dst_h * dst_w; ++hw) {
-            dst_ptr[oc * dst_h * dst_w + hw] = 0.0f;
-          }
-        }
+        memset(dst_ptr, 0.0f, sizeof(float) * oc_per_gp * dst_h * dst_w);
       }
       gemm_ref_fp32(filter + g * oc_per_gp * ic_per_gp * kernel_h * kernel_w,
                     columns, nullptr, dst_ptr, 0, 0, oc_per_gp, dst_h * dst_w,
                     ic_per_gp * kernel_h * kernel_w, 1.0f, 1.0f, dst_ptr);
     }
   }
-  delete[] columns;
 }
 
 MMCVModulatedDeformConvKernel::MMCVModulatedDeformConvKernel(
--- a/mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
+++ b/mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
@@ -1,246 +1,223 @@
 // Copyright (c) OpenMMLab. All rights reserved
-# include <vector>
-# include <cmath>
+#include "modulated_deform_conv.h"
 
-# include "../ort_mmcv_utils.h"
-# include "modulated_deform_conv.h"
+#include <cmath>
+#include <vector>
 
-float bilinear_interpolate_2d(
-    const float* src,
-    const int64_t src_h,
-    const int64_t src_w,
-    const float h,
-    const float w)
-{
-    if (h <= -1 || src_h <= h || w <= -1 || src_w <= w) {
-        return 0;
-    }
+#include "../ort_mmcv_utils.h"
 
-    int64_t h_low = floor(h);
-    int64_t w_low = floor(w);
-    int64_t h_high = h_low + 1;
-    int64_t w_high = w_low + 1;
+float bilinear_interpolate_2d(const float *src, const int64_t src_h,
+                              const int64_t src_w, const float h,
+                              const float w) {
+  if (h <= -1 || src_h <= h || w <= -1 || src_w <= w) {
+    return 0;
+  }
 
-    float lh = h - h_low;
-    float lw = w - w_low;
-    float hh = 1 - lh;
-    float hw = 1 - lw;
+  int64_t h_low = floor(h);
+  int64_t w_low = floor(w);
+  int64_t h_high = h_low + 1;
+  int64_t w_high = w_low + 1;
 
-    float v1 = 0;
-    if (h_low >= 0 && w_low >= 0)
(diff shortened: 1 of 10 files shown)


RunningLeon:
@PeiqiWang Hi, you could sign the CLA at first.

RunningLeon:
@PeiqiWang Hi, could you fix CI error at first?

RunningLeon:
@AllentDan Could you take a look at this PR?

AllentDan:
Hi, @PeiqiWang. you could run `pip install pre-commit` in your dev environment and then run `pre-commit run --all-files` to check lint before each commit automatically.

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
```diff
@@ -0,0 +1,311 @@
...
+    const int64_t stride_w,
+    const int64_t pad_h,
+    const int64_t pad_w,
+    const int64_t dilation_h,
+    const int64_t dilation_w,
+    float *dst)
+{
+    const int64_t ic_per_gp = channels / group;
+    const int64_t oc_per_gp = num_output / group;
+    float *columns = new float[ic_per_gp * kernel_h * kernel_w * dst_h * dst_w];
```
grimoire, line 184: Can we use smart pointer for the columns?

tests/test_ops/test_onnx.py
```diff
@@ -638,3 +638,93 @@ def test_roll(shifts_dims_pair):
         pytorch_output = wrapped_model(input.clone())
 
     torch.testing.assert_allclose(ort_output, pytorch_output)
+
+
+def test_modulated_deform_conv2d():
+    if torch.__version__ == 'parrots':
+        pytest.skip('onnx is not supported in parrots directly')
+    if not torch.cuda.is_available():
+        pytest.skip('modulated_deform_conv2d only supports in GPU')
```
grimoire, line 647: Could be replaced with:
```python
@pytest.mark.skipif(condition, reason='xxxx')
def test_modulated_deform_conv2d():
```

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
```diff
@@ -0,0 +1,311 @@
...
+                        dst_ptr[oc * dst_h * dst_w + hw] = bias_ptr[oc];
+                    }
+                }
+            } else {
+                for (int64_t oc = 0; oc < oc_per_gp; ++oc) {
+                    for (int64_t hw = 0; hw < dst_h * dst_w; ++hw) {
+                        dst_ptr[oc * dst_h * dst_w + hw] = 0.0f;
+                    }
+                }
+            }
```
grimoire, line 215: Can we initialize `dst` before batch loop?
We can just init once and do `memcpy` for rest of dst.

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
```diff
@@ -0,0 +1,311 @@
...
+    const int32_t N,
+    const int32_t K,
+    const float alpha,
+    const float beta,
+    float *Y)
+{
+    for (int64_t m = 0; m < M; ++m) {
+        for (int64_t n = 0; n < N; ++n) {
+            float y = 0.0f;
+            if (!trans_A && !trans_B) { // MK, KN; NN
```
AllentDan, line 127: Can we move the condition line outside the loop funciton of gemm? AFAK, a inside `if else` may slow a loop a lot compared with outside.

AllentDan (commented):
It worked all fine for me except a little slow for a large input size of mdcn. But we can speed it up through some tricks like define `macros` outside the gemm, compute a sub block of a matrix and so on...

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
PeiqiWang: The `bias_ptr` depends on `g`. So we cannot initialize it just once and `memcpy` for rest of `dst_ptr`

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
```diff
@@ -0,0 +1,297 @@
...
+            dst_ptr[oc * dst_h * dst_w + hw] = 0.0f;
+          }
+        }
+      }
+      gemm_ref_fp32(filter + g * oc_per_gp * ic_per_gp * kernel_h * kernel_w,
+                    columns, nullptr, dst_ptr, 0, 0, oc_per_gp, dst_h * dst_w,
+                    ic_per_gp * kernel_h * kernel_w, 1.0f, 1.0f, dst_ptr);
+    }
+  }
+  delete[] columns;
```
AllentDan, line 201: If we use `Alloc` outside `deformable_conv2d_ref_fp32`, maybe `Free` columns outside is better.

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
grimoire: Ok, what about memset(0) when bias is nullptr?

mmcv/ops/csrc/onnxruntime/cpu/modulated_deform_conv.cpp
PeiqiWang: Sorry forgot to delete the Line201. The ort allocator_ will be free automatically when the session closed.

AllentDan (approved):
LGTM

grimoire (approved):

ZwwWayne (approved):

PR #1282, opened 2021-08-18
Remove extra dim in cpu mode at scatter

innerlee:
## Motivation

Currently the cpu version of scatter will add an extra dim for tensors, which causes problems such as in https://github.com/open-mmlab/mmocr/pull/327

## Modification

Remove the extra dim

## BC-breaking (Optional)

YES

Downstream repos might be affected. Please check mmdet, mmaction2 and mmpose before merge. Please check their tests and also demos in cpu mode.

## Checklist

### DIFF
--- a/mmcv/parallel/data_parallel.py
+++ b/mmcv/parallel/data_parallel.py
@@ -18,13 +18,25 @@ class MMDataParallel(DataParallel):
     Args:
         module (:class:`nn.Module`): Module to be encapsulated.
         device_ids (list[int]): Device IDS of modules to be scattered to.
-            Defaults to None when GPU is not available.
+            Use None when GPU is not available. Note this is different
+            from pytorch's DataParallel where None means use all gpu devices.
         output_device (str | int): Device ID for output. Defaults to None.
         dim (int): Dimension used to scatter the data. Defaults to 0.
     """
 
     def __init__(self, *args, dim=0, **kwargs):
-        super(MMDataParallel, self).__init__(*args, dim=dim, **kwargs)
+        device_ids = kwargs.get('device_ids', None)
+        if device_ids is None:
+            device_ids = []
+        if device_ids == []:
+            # Handle cpu case. Bypass DataParallel because it cannot use cpu
+            # when cuda is available.
+            super(DataParallel, self).__init__()
+            self.module = args[0]
+            self.device_ids = []
+        else:
+            super().__init__(*args, dim=dim, **kwargs)
+
         self.dim = dim
 
     def forward(self, *inputs, **kwargs):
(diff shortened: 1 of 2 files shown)


mmcv/parallel/data_parallel.py
```diff
@@ -18,13 +18,25 @@ class MMDataParallel(DataParallel):
...
-            Defaults to None when GPU is not available.
+            Use None when GPU is not available. Note this is different
+            from pytorch's DataParallel where None means use all gpu devices.
         output_device (str | int): Device ID for output. Defaults to None.
         dim (int): Dimension used to scatter the data. Defaults to 0.
     """
 
     def __init__(self, *args, dim=0, **kwargs):
-        super(MMDataParallel, self).__init__(*args, dim=dim, **kwargs)
+        device_ids = kwargs.get('device_ids', None)
```
zhouzaida, line 28:
```python
        device_ids = kwargs.get('device_ids', None)
        if device_ids is None:
            device_ids = []
        if device_ids == []:
```
-> 
```python
        device_ids = kwargs.get('device_ids',  [])
        if len(device_ids) == 0:
```

innerlee: device_ids could be None (although not here)

hhaAndroid:
@innerlee I tested mmdet and it will report an error for the following reasons：

```shell
class Scatter:

    @staticmethod
    def forward(target_gpus, input):
        input_device = get_input_device(input)
        streams = None
        if input_device == -1 and target_gpus != [-1]:
            # Perform CPU to GPU copies in a background stream
            streams = [_get_stream(device) for device in target_gpus]
        # input shape=(1,3,100,100)
        # outputs shape=(1,3,100,100)
        outputs = scatter(input, target_gpus, streams)
        # Synchronize with the copy stream
        if streams is not None:
            synchronize_stream(outputs, target_gpus, streams)

        return tuple(outputs) # (3,100,100),
```
Because the batch dimension is missing, subsequent inferences will report an error.

innerlee:
@hhaAndroid do you know where the batch dim was added in gpu code?

hhaAndroid:
> @hhaAndroid do you know where the batch dim was added in gpu code?

The collate function output of dataloader includes the batch dimension.

innerlee:
Okay so we should make the behavior of collate consistent across cpu/gpu, right?

hhaAndroid:
> Okay so we should make the behavior of collate consistent across cpu/gpu, right?

No, your understanding is wrong. The `collate` does not consider CPU or GPU, the output always includes batch.

innerlee:
> The collate function output of dataloader includes the batch dimension.

Where is the extra dim added for gpu tensors?

hhaAndroid:
> > The collate function output of dataloader includes the batch dimension.
> 
> Where is the extra dim added for gpu tensors?

Sorry, I am late in replying, I will provide you with a detailed description document.

innerlee:
Cool, if you got extra time, please take a look at this bug https://github.com/open-mmlab/mmocr/pull/327. Checkout that branch and run this line https://github.com/open-mmlab/mmocr/pull/327/files#diff-2f44c053cce5ce1b42ec4cd7d50aca219b338cefe5aa8831813d6f4020fca740R14

HAOCHENYE:
Solved by #1621

PR #1289, opened 2021-08-23
fix parrots op bug

luopeichao:
Fix sync_bn bug in parrots. And when use parrots, open custom op exporting.

### DIFF
--- a/mmcv/onnx/info.py
+++ b/mmcv/onnx/info.py
@@ -1,6 +1,8 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import os
 
+import torch
+
 
 def is_custom_op_loaded():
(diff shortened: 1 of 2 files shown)


mmcv/onnx/info.py
```diff
@@ -16,4 +18,5 @@ def is_custom_op_loaded():
             flag = os.path.exists(ort_lib_path)
         except (ImportError, ModuleNotFoundError):
             pass
+    flag |= torch.__version__ == 'parrots'
```
zhouzaida, line 21: suggestion
```python
return flag or torch.__version__ == 'parrots'
```

zhouzaida (approved):
LGTM

PR #1337, opened 2021-09-15
[Feature] Add fps op from mmdet3d

DCNSW:
## Motivation

Add furthest point sample cuda operation from mmdet3d (branch: v1.0.0.dev0).

## Modification

Several files in mmcv/ops folder.

## BC-breaking (Optional)

No.

## Use cases (Optional)

from mmcv.ops import furthest_point_sample, furthest_point_sample_with_dist

PR #1338, opened 2021-09-15
[Feature] Add gather points op from mmdet3d

DCNSW:
## Motivation

Add gather points cuda operation from mmdet3d (branch: v1.0.0.dev0).

## Modification

Several files in mmcv/ops folder.

## BC-breaking (Optional)

No.

## Use cases (Optional)

from mmcv.ops import gather_points


grimoire:
Hi, I found that this ops can be implemented with PyTorch only:
```python

# method0, permute might take extra computation
def gather_point_pytorch0(features, idx):
    batch_size = features.shape[0]
    permute_features = features.permute(0, 2, 1)
    batch_idx = torch.arange(batch_size, device=features.device).unsqueeze(-1)
    permute_output = permute_features[batch_idx, idx, :]
    return permute_output.permute(0, 2, 1)

# method1, repeat might need more memory
def gather_point_pytorch1(features, idx):
    new_idx = idx.unsqueeze(1).repeat(1, 3, 1)
    return features.gather(2, new_idx)

```

DCNSW:
> Hi, I think this ops can be implemented with PyTorch only:
> 
> ```python
> # method0, permute might take extra computation
> def gather_point_pytorch0(features, idx):
>     batch_size = features.shape[0]
>     batch_idx = torch.arange(batch_size, device=features.device).unsqueeze(-1)
>     permute_output = features[batch_idx, :, idx]
>     return permute_output.permute(0, 2, 1)
> 
> # method1, repeat might need more memory
> def gather_point_pytorch1(features, idx):
>     new_idx = idx.unsqueeze(1).repeat(1, 3, 1)
>     return features.gather(2, new_idx)
> ```

We tested that the running time of the PyTorch version code is 17.3 times that of the Cuda version code, so it is necessary to implement with cuda.

grimoire:
> We tested that the running time of the PyTorch version code is 17.3 times that of the Cuda version code, so it is necessary to implement with cuda.

Here is my timer:
```python


def timer(func, num_test, num_warmup, *args, **kwargs):
    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)

    # warmup
    for _ in range(num_warmup):
        func(*args, **kwargs)

    torch.cuda.synchronize()

    start.record()
    for _ in range(num_test):
        output = func(*args, **kwargs)
    end.record()

    torch.cuda.synchronize()
    print(f'{func} take time {start.elapsed_time(end)/num_test}.')

    return output
```

With `warmup=5` and `num_test=10`, method0 is 3 times slower and method1 is 2 times slower. 
I believe custom cuda implement must be faster, but 17.3 times seems too much.

mmcv/ops/csrc/common/cuda/gather_points_cuda_kernel.cuh
```diff
@@ -0,0 +1,57 @@
...
+#define GATHER_POINTS_CUDA_KERNEL_CUH
+
+#ifdef MMCV_USE_PARROTS
+#include "parrots_cuda_helper.hpp"
+#else
+#include "pytorch_cuda_helper.hpp"
+#endif
+
+#define TOTAL_THREADS 1024
+#define DIVUP(m, n) ((m) / (n) + ((m) % (n) > 0))
```
zhouzaida, line 12: This line appears many times and does it need to be written in a public file?

DCNSW: refer to #1337

zhouzaida (approved):

ZwwWayne (approved):

PR #1360, opened 2021-09-22
[Enhancement] Raising warning when the different number of GPU is set for resuming

Junjun2016:
## Motivation

Raising warning when the different number of GPU is set for resuming

## Modification

The `resume` function in `iter_based_rnner`.

## BC-breaking (Optional)

No.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/iter_based_runner.py
+++ b/mmcv/runner/iter_based_runner.py
@@ -161,6 +161,20 @@ class IterBasedRunner(BaseRunner):
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
         self._inner_iter = checkpoint['meta']['iter']
+
+        # Raising warning when resuming models with different number of GPUs
+        if 'config' in checkpoint['meta']:
+            config = mmcv.Config.fromstring(
+                checkpoint['meta']['config'], file_format='.py')
+            previous_gpu_ids = config.get('gpu_ids', None)
+            if previous_gpu_ids and len(previous_gpu_ids) > 0 and len(
+                    previous_gpu_ids) != self.world_size:
+                warnings.warn(
+                    f'The number of GPU is {len(previous_gpu_ids)} before \
+                    resuming while the number of GPU is {len(self.world_size)}\
+                     after resuming. It is better to set the same number of \
+                    GPU for resuming.')
+
         if 'optimizer' in checkpoint and resume_optimizer:
             if isinstance(self.optimizer, Optimizer):
                 self.optimizer.load_state_dict(checkpoint['optimizer'])


mmcv/runner/iter_based_runner.py
```diff
@@ -161,6 +161,20 @@ def resume(self,
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
         self._inner_iter = checkpoint['meta']['iter']
+
+        # Raising warning when resuming models with different number of GPUs
```
zhouzaida, line 165: add a comment to explain why we recommend setting the same number of GPUs for resuming

mmcv/runner/iter_based_runner.py
```diff
@@ -161,6 +161,22 @@ def resume(self,
...
+            config = mmcv.Config.fromstring(
+                checkpoint['meta']['config'], file_format='.py')
+            previous_gpu_ids = config.get('gpu_ids', None)
+            if previous_gpu_ids and len(previous_gpu_ids) > 0 and len(
+                    previous_gpu_ids) != self.world_size:
+                warnings.warn(
+                    f'The number of GPU is {len(previous_gpu_ids)} before \
+                    resuming while the number of GPU is {len(self.world_size)}\
+                     after resuming. It is better to set the same number of \
+                    GPU for resuming.')
```
zhouzaida, line 178: Should we keep the consistent behavior with `BaseRunner` to modify the `_iter`
https://github.com/open-mmlab/mmcv/blob/b92ea0b5df9cfa4222fec704890bd6c8901fe6b8/mmcv/runner/base_runner.py#L371-L382

Junjun2016: No, there are no solid conclusions between optimizer status and batch size.

mmcv/runner/iter_based_runner.py
```diff
@@ -161,6 +161,22 @@ def resume(self,
         self._epoch = checkpoint['meta']['epoch']
         self._iter = checkpoint['meta']['iter']
         self._inner_iter = checkpoint['meta']['iter']
+
+        # Raising warning when resuming models with different number of GPUs
+        # since the optimizer status are relative with batch size
+        # (#GPUs x bs/GPU)
+        if 'config' in checkpoint['meta']:
+            config = mmcv.Config.fromstring(
```
Junjun2016, line 169: Can also check the difference of config between checkpoint and current.

Junjun2016: In the future.

mmcv/runner/iter_based_runner.py
MeowZheng: So? Should we rethink modifying `self._iter` in `BaseRunner`?
https://github.com/open-mmlab/mmcv/blob/e4b5348ebf407f0886bc11caa7ca74d2fbf693ac/mmcv/runner/base_runner.py#L369-L380

ZwwWayne:
Suggest only raising warnings to remind users because we are not sure the intention of users.

MeowZheng:
Might also raise a warning in `BaseRunner`? as people will not take log messages seriously

Issue #1339, opened 2021-09-16
'getpwuid(): uid not found: 20007'


ledaiduongvnth:
**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the FAQ documentation but cannot get the expected help.

I use mmcv inside of a docker container and the error occurs. Please help  
`  File "/app/mmdet/apis/train.py", line 150, in train_detector
    runner.run(data_loaders, cfg.workflow, cfg.total_epochs)
  File "/opt/conda/lib/python3.8/site-packages/mmcv/runner/epoch_based_runner.py", line 103, in run
    get_host_info(), work_dir)
  File "/opt/conda/lib/python3.8/site-packages/mmcv/runner/utils.py", line 16, in get_host_info
    return f'{getuser()}@{gethostname()}'
  File "/opt/conda/lib/python3.8/getpass.py", line 169, in getuser
    return pwd.getpwuid(os.getuid())[0]
KeyError: 'getpwuid(): uid not found: 20007'
`

torch 1.9.0
mmcv 1.3.3
mmdetection 2.7.0

zhouzaida:
hi @ledaiduongvnth , thanks for your feedback and we are fixing it. More details at #1374

PR #1374, opened 2021-09-26
[Enhancement] Add error handler when getting host failed

del-zhenwu:
Signed-off-by: del-zhenwu <zhenxiang.li@zilliz.com>

## Motivation

Try to fix issue #1339, add exception handler when get host failed

## Modification

Add exception handler in mmcv/runner/utils.py

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


mmcv/runner/utils.py
```diff
@@ -13,7 +14,13 @@
 
 
 def get_host_info():
-    return f'{getuser()}@{gethostname()}'
+    host = ''
```
zhouzaida, line 17: Suggest adding a comment to explain what situation the host can not be obtained

mmcv/runner/utils.py
```diff
@@ -13,7 +14,18 @@
 
 
 def get_host_info():
-    return f'{getuser()}@{gethostname()}'
+    """Get hostname and username.
+
+    Return empty string if exception raised, e.g. `getuser()` will lead to
+    error in docker container
```
zhouzaida, line 20: Maybe we can also mount `/etc/passwd` from our host when we run the docker with the following command to fix the exception. More details at https://stackoverflow.com/questions/48527958/docker-error-getting-username-from-password-database

```bash
docker run -v /etc/passwd:/etc/passwd ...
```

del-zhenwu: > Maybe we can also mount `/etc/passwd` from our host when we run the docker with the following command to fix the exception. More details at https://stackoverflow.com/questions/48527958/docker-error-getting-username-from-password-database
> 
> ```shell
> docker run -v /etc/passwd:/etc/passwd ...
> ```

This should be work, but from the user's point of view, it is even more inconvenient

zhouzaida (approved):

mmcv/runner/utils.py
```diff
@@ -13,7 +14,18 @@
...
+    """
+    host = ''
+    try:
+        host = f'{getuser()}@{gethostname()}'
+    except Exception as e:
+        warnings.warn('Host or user not found: %s' % (str(e)))
```
ZwwWayne, line 26: We usually use f-string, so please change it to `f'Host or user not found: {str(e)}'`

PR #1390, opened 2021-10-08
Fixed logging bug for object detection with validation set

joihn:
## Motivation

for object detection, the validation is done in 2 steps:
- validation bounding box Ap metric 
- validation Loss
the problem is that they are reported *separately* to wandb, with the *same* iter number, so wandb will ignore the second one (loss metric ) (undesired).

## Modification

This fix increment the iter number when reporting validation loss

## BC-breaking (Optional)

My fix might not be the best approach if the user is not using this lib for object detection, but should still work (untested in this case) (I tested only with MMdetection)

### DIFF
--- a/mmcv/runner/hooks/logger/wandb.py
+++ b/mmcv/runner/hooks/logger/wandb.py
@@ -45,8 +45,20 @@ class WandbLoggerHook(LoggerHook):
         tags = self.get_loggable_tags(runner)
         if tags:
             if self.with_step:
+                temp_iter = self.get_iter(runner)
+                #for object detection, the validation is done in 2 steps:
+                #validation Ap metric 
+                #validation Loss
+                # the problem is that they are reported separately to wandb, with the same iter number, so
+                #wandb will ignore the second one (loss) 
+                for key in tags:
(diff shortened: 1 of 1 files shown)


MengzhangLI:
Hi, related issue in MMSegmentation: https://github.com/open-mmlab/mmsegmentation/issues/1068.

Our user says it would not fix bug in MMSegmentation but he or she gave the alternative solution.

Best,

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -45,8 +45,20 @@ def log(self, runner):
         tags = self.get_loggable_tags(runner)
         if tags:
```
fcakyon, line 46: @joihn @ZwwWayne @MengzhangLI should be:

```python
if tags:
    tags['global_step'] = self.get_iter(runner)
    self.wandb.log(tags, commit=self.commit)
```

check open-mmlab/mmsegmentation#1068.

fcakyon (changes requested):

OpenMMLab-Assistant005:
Hi @joihn ！We are grateful for your efforts in helping improve this open-source project during your personal time. 

PR #1395, opened 2021-10-11
Fix potiential bug in `deprecated_api_warning`

jshilong:
## Motivation
We usually use `deprecated_api_warning` as a wapper to deprecated some arguments. There is a rare case that
deprecated arguments and new arguments exist in the config at the same time, we will override the value of the new argument with the value of deprecated arguments in our implementation, which is dangerous.

This usually happens when you specify new arguments and inherit an old config that contains deprecated arguments

## Modification

Assert this case and raise an error

## BC-breaking (Optional)

If this case exists in the downstream codebase, it would raise an error now.


tests/test_utils/test_misc.py
```diff
@@ -190,3 +191,17 @@ def foo1():
...
         mmcv.is_method_overridden('foo1', base_instance, sub_instance)
+
+
+def test_deprecated_api_warning():
+
+    @deprecated_api_warning(name_dict=dict(old_key='new_key'))
+    def dummy_func(new_key=1):
+        return new_key
+
+    # test can map old_key to new_key
```
zhouzaida, line 202:
```suggestion
    # replace `old_key` to `new_key`
```

zhouzaida (approved):

PR #1396, opened 2021-10-11
add world_size in if clause in load_from_http

tobiasfshr:
## Motivation

I'm using mmcv within another codebase and i have issues with loading a checkpoint from http in multi-GPU setting. 
Specifically i get the error: 

```
File "/mmcv/runner/checkpoint.py", line 290, in load_from_http
return checkpoint
UnboundLocalError: local variable 'checkpoint' referenced before assignment
```

This is because get_dist_info does not correctly get world_size in L280, but rank is correctly inferred via L281. Hence, the variable checkpoint will not be defined on return (rank > 0 but world_size == 1).

## Modification

I can mitigate the problem by modifying this line:
https://github.com/open-mmlab/mmcv/blob/f22c9eb4a409470b7e645f17fa1997fe85e27909/mmcv/runner/checkpoint.py#L282

to 

`if rank == 0 or world_size == 1:`

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -278,12 +278,7 @@ def load_from_http(filename, map_location=None, model_dir=None):
         dict or OrderedDict: The loaded checkpoint.
     """
     rank, world_size = get_dist_info()
-    rank = int(os.environ.get('LOCAL_RANK', rank))
-    # If LOCAL_RANK is set before ddp is initialized, the if condition
-    # 'if dist.is_available() and dist.is_initialized()' in get_dist_info will
-    # fail and lead to world_size = 1 while rank > 0 for non-master processes.
-    # Thus, check if either rank == 0 or world_size == 1 here.
-    if rank == 0 or world_size == 1:
+    if rank == 0:
         checkpoint = model_zoo.load_url(
             filename, model_dir=model_dir, map_location=map_location)
     if world_size > 1:
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -279,7 +279,7 @@ def load_from_http(filename, map_location=None, model_dir=None):
     """
     rank, world_size = get_dist_info()
     rank = int(os.environ.get('LOCAL_RANK', rank))
-    if rank == 0:
+    if rank == 0 or world_size == 1:
         checkpoint = model_zoo.load_url(
             filename, model_dir=model_dir, map_location=map_location)
     if world_size > 1:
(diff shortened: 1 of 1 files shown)


zhouzaida:
> Sometimes the environment variable LOCAL_RANK is set even before ddp is fully initialized by pytorch, so in get_dist_info the condition ` if dist.is_available() and dist.is_initialized():` will fail, and thus return world_size = 1 and rank = 0. However, in L281 rank is set to LOCAL_RANK and thus rank will be > 0 for the non-master processes.

Thanks for your explanation.

zhouzaida:
> Sometimes the environment variable LOCAL_RANK is set even before ddp is fully initialized by pytorch, so in get_dist_info the condition ` if dist.is_available() and dist.is_initialized():` will fail, and thus return world_size = 1 and rank = 0. However, in L281 rank is set to LOCAL_RANK and thus rank will be > 0 for the non-master processes.

Theoretically, `rank = int(os.environ.get('LOCAL_RANK', rank))` is redundant because the model state of rank 0 will broadcast to other processes before `forward pass`.

tobiasfshr:
yes one could also remove L281. But i assume there was a reason for adding this in the first place?

zhouzaida:
Your consideration is reasonable, I consulted the author, who felt that this line `rank = int(os.environ.get('LOCAL_RANK', rank))` could be left out. From the principle of DDP, this line is not necessary to be added.

tobiasfshr:
okay so i'll remove L281 and the explanation but keep the 'or world_size == 1' in the if condition, since this seems like a better behavior to me (if world_size == 1 it doesn't make sense to execute .barrier() regardless of the rank value). Sounds good?

zhouzaida:
> okay so i'll remove L281 and the explanation but keep the 'or world_size == 1' in the if condition, since this seems like a better behavior to me (if world_size == 1 it doesn't make sense to execute .barrier() regardless of the rank value). Sounds good?

In fact. if the world_size is 1, the rank should be 0 when we remove the L281.

tobiasfshr:
okay removed the check. My point was more that the additional check for world_size would make the code a little more robust to unexpected values of rank, but i get your point after removing L281 rank should always be 0 if world_size = 1.

zhouzaida (approved):

OpenMMLab-Coodinator:
Hi @tobiasfshr ！First of all, we want to express our gratitude for your significant PR in the MMCV project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

If you are Chinese or have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

PR #1403, opened 2021-10-14
[Feature] Add CRF module

459737087:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

add the crf block, and crf can improve the model capability

## Modification

add the crf module and test_crf.py

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

**Before PR**:

- [x] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.

PR #1424, opened 2021-10-21
增加crf fixes: 1391

459737087:
I have chenged the code and remove those wrong codes.
@zhouzaida


zhouzaida:
> OK I install the pre-commit, and setup.cfg is consistent with https://github.com/open-mmlab/mmcv/blob/master/setup.cfg @zhouzaida

hi, I cleaned the commit of #1403, and can we continue the PR in #1403?

459737087:
> > OK I install the pre-commit, and setup.cfg is consistent with https://github.com/open-mmlab/mmcv/blob/master/setup.cfg @zhouzaida
> 
> hi, the commit of #1424 is a little difficult to clean so I cleaned the commit of #1403, and can we continue the PR in #1403?

ok I'll chenge

zhouzaida:
Repeat with #1403

PR #1438, opened 2021-10-27
[Fix] Rename the init parameter task to name in PaviLoggerHook

ly015:

## Motivation

Fix some minor issues in PaviLoggerHook.

## Modification

- Use `name` instead of `task` according to the latest PAVI interface
- Call `writer.close()` in the deconstructor

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/hooks/logger/pavi.py
+++ b/mmcv/runner/hooks/logger/pavi.py
@@ -101,8 +101,7 @@ class PaviLoggerHook(LoggerHook):
                     iteration=iteration)
 
         # flush the buffer and send a task ending signal to Pavi
-        if hasattr(self, 'writer'):
-            self.writer.close()
+        self.writer.close()
 
     @master_only
     def before_epoch(self, runner):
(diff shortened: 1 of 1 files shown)


mmcv/runner/hooks/logger/pavi.py
```diff
@@ -100,6 +100,10 @@ def after_run(self, runner):
                     snapshot_file_path=ckpt_path,
                     iteration=iteration)
 
+        # flush the buffer and send a task ending signal to Pavi
+        if hasattr(self, 'writer'):
```
zhouzaida, line 104: What case the `self` object miss the attribute `writer`

ly015: This part was placed in `__del__` so we checked the `writer` in case that `before_run` has not run and built the writer.
Since it's in `after_run` now which is always called after `before_run`, maybe this check is no longer necessary?

zhouzaida: yet

zhouzaida (approved):

Issue #990, opened 2021-04-25
Detect checkpoint files more effectively


Yshuo-Li:
**Describe the feature**

Detect checkpoint files more effectively

**Motivation**

In `mmcv/runner/checkpoint.py`, line 259:
`if not osp.isfile(filename):
        raise IOError(f'{filename} is not a checkpoint file')`
Only files with absolute paths will be detected (python 3.7, torch 1.6).

**Additional context**

If necessary, I can try to fix this issue.

PR #1443, opened 2021-11-01
[Fix] fix checkpoint local files detect

teamwong111:

## Motivation

Related Issue: https://github.com/open-mmlab/mmcv/issues/990
Detect checkpoint files more effectively

## Modification

Add support for local files with relative path detect

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/checkpoint.py
+++ b/mmcv/runner/checkpoint.py
@@ -255,7 +255,7 @@ def load_from_local(filename, map_location):
     Returns:
         dict or OrderedDict: The loaded checkpoint.
     """
-    filename = osp.abspath(filename) if not osp.isabs(filename) else filename
+    filename = osp.expanduser(filename)
     if not osp.isfile(filename):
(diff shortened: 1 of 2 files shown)


mmcv/runner/checkpoint.py
```diff
@@ -255,7 +255,7 @@ def load_from_local(filename, map_location):
     Returns:
         dict or OrderedDict: The loaded checkpoint.
     """
-
+    filename = osp.abspath(filename) if not osp.isabs(filename) else filename
```
zhouzaida, line 258: `isfile` can handle the relative path like `../../path/of/your/checkpoint` but not `~/path/of/your/file`. In addition, `abspath('~/path/of/your/file')` returns an unexpected path so we should use [os.path.expanduser](https://docs.python.org/zh-cn/3/library/os.path.html#os.path.expanduser). BTW, a unit test should be added to test those cases.

Issue #1387, opened 2021-10-06
mmcv/parallel/collate.py issue when collecting filenames


cericyeh:
Hi guys, I've run into an issue attempting to collect filenames when running the test pipeline for my action recognition model, which as run via mmaction2 (main branch).  The portion this report refers to is on mmcv-full v1.3.14.

In my model configuration's pipeline, I added the 'filename' to the keys to collect, in order to include that information for debugging purposes.  I adjusted the "Collect" entries in the mmaction2 config file's pipeline entries, e.g.,

     dict(type='Collect', keys=['imgs', 'label', 'filename'], meta_keys=[]),

This results in a stack overflow around mmcv/parallel/collate.py L77, and a fragment of that is,

  /home/local/ceyFile "/home/local/cey/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/parallel/collate.py", line 77, in collate
    return [collate(samples, samples_per_gpu) for samples in transposed]
  File "/home/local/cey/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/parallel/collate.py", line 77, in <listcomp>
    return [collate(samples, samples_per_gpu) for samples in transposed]
  File "/home/local/cey/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/parallel/collate.py", line 77, in collate

This occurred when executing mmaction2/tools/test.py using the tsm_k400_pretrained_r50_1x1x8_25e_ucf101_rgb.video.py config against UCF101.

I believe the issue stems from the fact that the type check for the batch elements, on L77 of collate.py, does not distinguish between a list-like sequence versus a string,

  elif isinstance(batch[0], Sequence):

If you pass in a list of strings into the batch argument for collate() function, continuing the recursion on the transpose of that list will result in an infinite loop.  

I found that if you changed the check on L77 to check against strings,

  elif isinstance(batch[0], Sequence) and not(instance(batch[0], str)):

then the collation works, and the list of filenames bottoms out correctly to the default collate function.

Note at least for the default mmaction2/tools/test.py, I had to adjust the forward() call to avoid passing in the keyword argument for filenames for this to work, but that's a fairly easy workaround.

I have made no other changes to the code.

Please let me know if this indeed seems correct, and if it does not introduce any errors.  

Thanks,
Eric

Details about environment follow,
**Environment**
This was run from an Anaconda distribution on Ubuntu Linux.

The collect_env() results are,
[...]

PR #1445, opened 2021-11-01
[Fix] fix the bug caused by batch element is str

teamwong111:

## Motivation

Related Issue: https://github.com/open-mmlab/mmcv/issues/1387
Fix the bug caused by batch element is str

## Modification

Add a condition to check if the batch[0] is str

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/parallel/collate.py
+++ b/mmcv/parallel/collate.py
@@ -72,7 +72,7 @@ def collate(batch, samples_per_gpu=1):
                 stacked.append(
                     [sample.data for sample in batch[i:i + samples_per_gpu]])
         return DataContainer(stacked, batch[0].stack, batch[0].padding_value)
-    elif isinstance(batch[0], Sequence):
+    elif isinstance(batch[0], Sequence) and not isinstance(batch[0], str):
         transposed = zip(*batch)
         return [collate(samples, samples_per_gpu) for samples in transposed]
     elif isinstance(batch[0], Mapping):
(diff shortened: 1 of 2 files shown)


tests/test_parallel.py
```diff
@@ -64,3 +65,8 @@ def forward(self, *args, **kwargs):
 
     module_wraper = ModuleWrapper(model)
     assert is_module_wrapper(module_wraper)
+
+
+def test_collate():
```
zhouzaida, line 70: It seems like that the collate function has not been tested before. And we can add full unit tests for it.

mmcv/parallel/collate.py
```diff
@@ -72,7 +72,7 @@ def collate(batch, samples_per_gpu=1):
                 stacked.append(
                     [sample.data for sample in batch[i:i + samples_per_gpu]])
         return DataContainer(stacked, batch[0].stack, batch[0].padding_value)
-    elif isinstance(batch[0], Sequence):
+    elif isinstance(batch[0], Sequence) and not isinstance(batch[0], str):
```
zhouzaida, line 75: Add a comment to describe why we need to check whether `batch[0]` is a string. In addition, the issue link can be provided here like `More details at https://github.com/open-mmlab/mmcv/issues/1387.`

zhouzaida:
Please @274869388 have a look.

PR #1471, opened 2021-11-09
Add feature : config file dump to wandb server

hyun06000:

## Motivation

WandB has many features to record ML experiments. It can record python scripts or any files on WandB repository as well as plots of training and measurements. So I want to add a feature to upload a config file of mmdetection or mmsegmentation etc. The config file is the most important part of open-mmlab. This pull-request can support a model comparison and reconstruction at the same time at wandb.  

## Modification

file:   mmcv/runner/hooks/logger/wandb.py

line: 2 / 20 / 27 / 47 / 48

## Use cases (Optional)

image
This is an example of config file ofr wandb logger. If user can set a work-dir before setting config file, user can write down the dumped config file path directly.

image
every mm{library}/tools/train.py includes a line for dump the config file of each experiment. So user can set a line to write a config path for wandb under the line.

```
config_dump_path = osp.join(cfg.work_dir, osp.basename(args.config))
cfg.dump(config_dump_path)
cfg.log_config.hooks[2].config_path = config_dump_path

```

This is a result.
image

image

## Checklist

**Before PR**:

**After PR**:


mmcv/runner/hooks/logger/wandb.py
```diff
@@ -39,6 +43,9 @@ def before_run(self, runner):
             self.wandb.init(**self.init_kwargs)
         else:
             self.wandb.init()
+        
+        if self.config_path is not None:
+            os.system(f"cp -r {self.config_path} {self.wandb.run.dir}")
```
zhouzaida, line 48: The following lines also work in the Windows environment.
```python
import shutil
shutuil.copy2(self.config_path, self.wandb.run.dir)
```

hyun06000: Okay. I thank you for recommendation.

hyun06000: It is changed in under commit.

zhouzaida:
> cfg.log_config.hooks[2].config_path = config_dump_path

The above line may not be the best way to set the `config_path`. Maybe we can manually set the `config_path` in config files or get the value from the runner but we need to add it to `runner` in the downstream codebases. I prefer the former.

hyun06000:
> > > cfg.log_config.hooks[2].config_path = config_dump_path
> > 
> > 
> > The above line may not be the best way to set the `config_path`. Maybe we can manually set the `config_path` in config files or get the value from the runner but we need to add it to `runner` in the downstream codebases. I prefer the former.
> 
> But the former way is not a convenient way to set the config_path which requires us to manually set it every time when we launch another experiment.

I agree that to do something to `runner` for this feature is not effecient way and the former is better method.

Actually I set the config for each experiment when there was no hyperparameter auto tunning. But when I used a hyperparameter tunning tool (wandb sweep), the tool was worked with a train script which includes adjusting and dumping config. In my opinion, this feature is more useful with hyperparameter tunning. That is why I write the line `cfg.log_config.hooks[2].config_path = config_dump_path`.

hyun06000:
> > > cfg.log_config.hooks[2].config_path = config_dump_path
> > 
> > 
> > The above line may not be the best way to set the `config_path`. Maybe we can manually set the `config_path` in config files or get the value from the runner but we need to add it to `runner` in the downstream codebases. I prefer the former.
> 
> But the former way is not a convenient way to set the config_path which requires us to manually set it every time when we launch another experiment.

Actually I set the config for each experiment when there was no hyperparameter auto tunning. But when I used a hyperparameter tunning tool (wandb sweep), the tool was worked with a train script which includes adjusting and dumping config. In my opinion, this feature is more useful with hyperparameter tunning. That is why I write the line `

zhouzaida:
> Actually I set the config for each experiment when there was no hyperparameter auto tunning. But when I used a hyperparameter tunning tool (wandb sweep), the tool was worked with a train script which includes adjusting and dumping config. In my opinion, this feature is more useful with hyperparameter tunning. That is why I write the line `cfg.log_config.hooks[2].config_path = config_dump_path`.

Got it. We need to add a unit test for the feature.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -14,13 +16,15 @@ def __init__(self,
                  reset_flag=False,
                  commit=True,
                  by_epoch=True,
-                 with_step=True):
+                 with_step=True,
+                 config_path=None):
```
zhouzaida, line 20: We can add a docstring for the two parameters.

hyun06000: The line 19 is just added comma. Can I know how I can make test and docstring?

zhouzaida: https://github.com/open-mmlab/mmcv/blob/792473c960ed91e334164debc6877a96fdac9558/tests/test_runner/test_hooks.py#L1113-L1130

zhouzaida:
```python
class WandbLoggerHook(LoggerHook):
    """
    Args:
        with_step (bool): xxxx. Default: True.
        config_path (str, optional): xxxx. Default: None.
    """
```

hyun06000: I can't find `assert_called_with()` attribute in wandb docs. And `join` is deprecated. Can I know it is the edge of version or working test?

hyun06000: [this](https://github.com/wandb/client/issues/1991) is an issue of wandb unit test.

zhouzaida: We use `MagicMock` to mock the `wandb`. More details about `MagicMock` at https://docs.python.org/3/library/unittest.mock.html#quick-guide

hyun06000: My mistake. I just get the code. It is attribute of `mock.py`. I will make a unit test in that format. But `join` is deprecated attribute. It was moved to `finish`. please check it.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -46,7 +47,20 @@ def before_run(self, runner):
...
+                        base_path=self.wandb.run.dir,
+                        policy="now"
+                    )
+            else:
+                self.wandb.save(
+                    glob_str=self.wandb.run.dir + "/*",
+                    base_path=self.wandb.run.dir,
+                    policy="now"
+                )
+
```
hyun06000, line 76: This part is added to use wandb.save(). we can upload our config with just copy of config dir, but in this case the uploading occurs at the end of the process. Now, using this lines, we can upload config at the start of training.

tests/test_runner/test_hooks.py
```diff
@@ -1121,6 +1121,11 @@ def test_wandb_hook():
     shutil.rmtree(runner.work_dir)
 
     hook.wandb.init.assert_called_with()
+    hook.wandb.save.assert_called_with(
+        glob_str=hook.wandb.run.dir + "/*",
+        base_path=hook.wandb.run.dir,
+        policy="now"
+    )
```
hyun06000, line 1128: This part is the unit test of above lines with mock.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -14,13 +17,28 @@ def __init__(self,
...
+        Args:
+            
+            with_step (bool): whether making a log in each step or not.
+                        Default: True.
+            config_path (str, optional): The path of final config of each experiment.
+                        It can be either a path of final config file or a directory of 
+                        config files. The config is uploaded to wandb server if it is
+                        not None. Default: None.
+        """
+
```
hyun06000, line 36: This part is the docstring of wandb logger class. I wrote docs for only two arguments because the author of this class is not I. I worried that I can't wrote the exact function.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -39,7 +57,23 @@ def before_run(self, runner):
...
+                        glob_str=path + "/*",
+                        base_path=self.wandb.run.dir,
+                        policy="now"
+                    )
+            else:
+                self.wandb.save(
+                    glob_str=self.wandb.run.dir + "/*",
+                    base_path=self.wandb.run.dir,
+                    policy="now"
+                )
```
hyun06000, line 75: This part is added to use wandb.save(). we can upload our config with just copy of config dir, but in this case the uploading occurs at the end of the process. Now, using this lines, we can upload config at the start of training.

tests/test_runner/test_hooks.py
hyun06000, line 1128: This part is the unit test of above lines with mock.

tests/test_runner/test_hooks.py
```diff
@@ -1121,7 +1121,7 @@ def test_wandb_hook():
     
     hook.wandb.init.assert_called_with()
     
-    hook.wandb.run.dir = runner.work_dir
+    hook.wandb.run.dir.return_value = runner.work_dir
     hook.wandb.save.assert_called_with(
         glob_str=hook.wandb.run.dir + '/*',
         base_path=hook.wandb.run.dir,
```
hyun06000: In this test, `hook.wandb` is not `wandb.run`.  `wandb.init()` is a code to make an `wandb.run` object, so after initialization of wandb, every `wandb` is alias of  `wandb.run` (more details are [here](https://docs.wandb.ai/ref/python/init)). Therefore, I should use `wandb.run`  to call `wandb.run.dir`, but upper line defines `wandb` is MagicMock. Instad of modification of whole lines in this project for wandb, I just use `return_value` to return the work_dir of runner. This is the same work with `shutil.copy2` and `wandb.save` in wandb logger script.

mmcv/runner/hooks/logger/wandb.py
zhouzaida: Yet, two arguments are enough here. BTW, the convention of mmcv is that put the docstring on top of `__init__` method so the docstring should be moved to line 12.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -14,13 +17,28 @@ def __init__(self,
...
+        """
+        Args:
+            
+            with_step (bool): whether making a log in each step or not.
+                        Default: True.
+            config_path (str, optional): The path of final config of each experiment.
+                        It can be either a path of final config file or a directory of 
+                        config files. The config is uploaded to wandb server if it is
+                        not None. Default: None.
+        """
```
zhouzaida, line 35:
```suggestion
        """
        Args:
            with_step (bool): whether making a log in each step or not.
                Default: True.
            config_path (str, optional): The path of the final config of each
                experiment. It can be either a path of the final config file or
                a directory of config files. The config is uploaded to wandb
                server if it is not None. Default: None.
                `New in version 1.3.18.` 
        """
```

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -39,7 +57,23 @@ def before_run(self, runner):
             self.wandb.init(**self.init_kwargs)
         else:
             self.wandb.init()
-
+        
+        if self.config_path is not None:
+            shutil.copy2(self.config_path, self.wandb.run.dir)
```
zhouzaida, line 62: Can the `config_path` also be a path of a directory?

zhouzaida: Note that `shutil.copy2` only works for copying a file rather than a directory.

hyun06000: You are right. I should add a line to copy a tree for files such as `shutil.copytree`. 
I thought about just case of hyperparam tunning, but I got a recommendation from my peer about the usage of config dirs. In general case, many users want to manage their configurations in a separated from such as model.py, optimizer.py, logger.py etc.

hyun06000: In the under commit, I use copy_tree function in distutils.dir_util because shutil.copytree doesn't work with an existing path. In recent versions of shutil.copytree have an argument to work with existing path but not for old python.

mmcv/runner/hooks/logger/wandb.py
hyun06000: Thank you for the correction.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -59,20 +60,26 @@ def before_run(self, runner):
             self.wandb.init()
         
         if self.config_path is not None:
-            shutil.copy2(self.config_path, self.wandb.run.dir)
             if os.path.isdir(self.config_path):
-                for path, _, _ in os.walk(self.wandb.run.dir):
+                copy_tree(self.config_path, self.wandb.run.dir)
```
zhouzaida, line 64: Is it necessary to copy the directory to `self.wandb.run.dir` first and then save (upload) the `self.wandb.run.dir` to wandb?

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -59,20 +60,26 @@ def before_run(self, runner):
...
                         policy='now'
                     )
             else:
-                self.wandb.save(
-                    glob_str=self.wandb.run.dir + '/*',
-                    base_path=self.wandb.run.dir,
-                    policy='now'
-                )
+                if os.path.isfile(self.config_path):
+                    shutil.copy2(self.config_path, self.wandb.run.dir)
```
zhouzaida, line 73: Same as above

hyun06000: This method was the best way in my experience. When I want to rebuild the same model with a specific config file, I sometimes experienced a hard time to find what I want to use because of a directory name. wandb `name` is a hashing value as default and not same as the `work-dir`. On the other hands, the copy version of config in `wandb.run.dir` could always have a high reliability.

hyun06000:
@zhouzaida I believed the malfunction was from my mistakes, however, as the results of some tests, I found this function does not work with a path above `wandb.run.dir`. The `wandb.save` needs `glob_str` and `base_path` to find a file to upload. I found that the `glob_str` is not real `global path` for user's file system. The `wandb.run.dir` is root for `wandb.run`.
image

This image is the result of `wandb.save` with a `glob_str` above the `root` for `wandb.run`(I mean `wandb.run.dir`). When I set the path of a tartget file with `os.path.dirname()`, the function doesn't save any file. Also, without `base_path`, the function showed the error like below.
image

Therefore, we have two choice whether to move `wandb.run.dir` or to make copy in `wandb.run.dir`. In my opinion, the latter is better than the former in general cases.

fcakyon:
@hyun06000 @zhouzaida this feature is so much needed, cant wait to see this merged!

zhouzaida:
> 

Sorry for the late reply

zhouzaida:
> @hyun06000 @zhouzaida this feature is so much needed, cant wait to see this merged!

Got it.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -39,7 +58,29 @@ def before_run(self, runner):
...
+                    self.wandb.save(
+                        glob_str=os.path.join(self.wandb.run.dir,'*'),
+                        base_path=self.wandb.run.dir,
+                        policy='now'
+                    )
+                else:
+                    raise FileNotFoundError(
+                        "No such file or directory: " + self.config_path
+                    )
```
zhouzaida, line 82: Should we move those lines to `before_run` because it will make `import_wandb` simpler and clearer.

mmcv/runner/hooks/logger/wandb.py
```diff
@@ -14,13 +18,28 @@ def __init__(self,
...
+        """
+        Args:
+            with_step (bool): whether making a log in each step or not.
+                Default: True.
+            config_path (str, optional): The path of the final config of each
+                experiment. It can be either a path of the final config file or
+                a directory of config files. The config is uploaded to wandb
+                server if it is not None. Default: None.
+                `New in version 1.3.18.` 
+        """
```
zhouzaida, line 36: suggest moving those lines to line13.

zhouzaida:
> @zhouzaida I believed the malfunction was from my mistakes, however, as the results of some tests, I found this function does not work with a path above `wandb.run.dir`. The `wandb.save` needs `glob_str` and `base_path` to find a file to upload. I found that the `glob_str` is not real `global path` for user's file system. The `wandb.run.dir` is root for `wandb.run`. image
> 
> This image is the result of `wandb.save` with a `glob_str` above the `root` for `wandb.run`(I mean `wandb.run.dir`). When I set the path of a tartget file with `os.path.dirname()`, the function doesn't save any file. Also, without `base_path`, the function showed the error like below. image
> 
> Therefore, we have two choice whether to move `wandb.run.dir` or to make copy in `wandb.run.dir`. In my opinion, the latter is better than the former in general cases.

Yet. As you said, If we do not provide the `base_path` parameter, any folder will not be created which means all files will be uploaded to the same level directory `root`. It is not an expected behavior. So we can do it ourselves. If the `config_path` is a directory, we can set the `base_path` as `config_path` and the `glob_str` as `*`. If the `config_path` is a file, we can set the `base_path` as `dirname(config_path)` and `glob_str` as `basename(config_path)`.

More details about `wandb.save` can be found at https://github.com/wandb/client/blob/v0.12.7/wandb/sdk/wandb_run.py#L1349-L1438.

mmcv/runner/hooks/logger/wandb.py
hyun06000: It is already in `before_run` function. Please check it. When we run all these lines, we need the `wandb.run.dir`. `wandb.run.dir` is defined at `wandb.init()`. So I have no other opinion to move.

mmcv/runner/hooks/logger/wandb.py
hyun06000: You are right. It is moved.

hyun06000:
> > @zhouzaida I believed the malfunction was from my mistakes, however, as the results of some tests, I found this function does not work with a path above `wandb.run.dir`. The `wandb.save` needs `glob_str` and `base_path` to find a file to upload. I found that the `glob_str` is not real `global path` for user's file system. The `wandb.run.dir` is root for `wandb.run`. image
> > This image is the result of `wandb.save` with a `glob_str` above the `root` for `wandb.run`(I mean `wandb.run.dir`). When I set the path of a tartget file with `os.path.dirname()`, the function doesn't save any file. Also, without `base_path`, the function showed the error like below. image
> > Therefore, we have two choice whether to move `wandb.run.dir` or to make copy in `wandb.run.dir`. In my opinion, the latter is better than the former in general cases.
> 
> Yet. As you said, If we do not provide the `base_path` parameter, any folder will not be created which means all files will be uploaded to the same level directory `root`. It is not an expected behavior. So we can do it ourselves. If the `config_path` is a directory, we can set the `base_path` as `config_path` and the `glob_str` as `*`. If the `config_path` is a file, we can set the `base_path` as `dirname(config_path)` and `glob_str` as `basename(config_path)`.
> 
> More details about `wandb.save` can be found at https://github.com/wandb/client/blob/v0.12.7/wandb/sdk/wandb_run.py#L1349-L1438.

First of all, I'm sorry for the late response.
The link was very useful for me to understand `wandb.save`. Thanks.

mmcv/runner/hooks/logger/wandb.py
hyun06000: @zhouzaida After moving it, docs reader make build fail. Should I indicate the docs is for `__init__`? Please let me know about it.

hyun06000: @zhouzaida Sorry, it was version issue, my mistake. I didn't update my forked repo so long time...

zhouzaida:
Hi, the CI failed.

zhouzaida:
PR #1616 has supported uploading files to the wandb server so should I close the PR.

OpenMMLab-Assistant005:
Hi @hyun06000 ！We are grateful for your efforts in helping improve this open-source project during your personal time. 

Issue #976, opened 2021-04-22
[Feature Request]: Support to add information of git hash in version file during MMCV installation


ZwwWayne:
**Describe the feature**

This issue requests to support to add information of git hash in version file during MMCV installation.

**Motivation**
The current `collect_env()` or version file does not provide a way to store the information of git hash of MMCV or mm repos.
It is inconvenient when debug or benchmarking on a specific branch and the developers need to manually check the MMCV version.
The information is also helpful to reproduce the experiments or bugs when necessary.

**Additional context**
A naive way is to add the git hash information during installation, e.g., write it in a file and provide an API to read it. Any implementation strategy is acceptable as long as it is elegant.

PR #1493, opened 2021-11-15
Add version info

teamwong111:

## Motivation

Related issue: https://github.com/open-mmlab/mmcv/issues/976
Add GCC version, Cuda version, torch version and git hash in version.py. Honestly, I am not quite sure about stability on Linux, Windows and macOS.

## Modification

Impl related functions. 

## BC-breaking (Optional)

## Use cases (Optional)

When run `pip install [option] .` or `python setup.py [option]`, some version info will be added in `version.py`. It may look like this
```python
GCC = '7.5.0'
CUDA = '11.2.78'
PyTorch = '1.5.1'
CommitID = 'ad71e62'
```

According to our Github CI, we can see it is more easy to get these info.
```shell
Run python -c "import mmcv; print(f'GCC {mmcv.version.gcc_version}', f'CUDA {mmcv.version.cuda_version}', f'PyTorch {mmcv.version.torch_version}', f'CommitID {mmcv.version.commit_id}', sep='\n')"
GCC 7.5.0
CUDA None
PyTorch 1.5.1
CommitID ad71e62
```

## Checklist

**Before PR**:

**After PR**:


zhouzaida:
We can print the version information in CI.

- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L46
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L80
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L119
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L186
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L244
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L276
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L305
- https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/.github/workflows/build.yml#L349

setup.py
```diff
@@ -19,6 +24,187 @@
...
+
+        return sha
+
+    mmcv_root = Path(__file__).parent
+    version_path = mmcv_root / 'mmcv' / 'version.py'
+    gcc_version = get_gcc()
+    cuda_version = get_cuda()
+    git_hash = get_git_hash()
+    if EXT_TYPE:
+        torch_version = torch.__version__
```
zhouzaida, line 193: The value `torch.__version__` is different between `pip install torch` and `conda install torch`. More details at https://github.com/open-mmlab/mim/blob/62bd756d7045c3c591d442415c72a8308e778207/mim/utils/utils.py#L321

setup.py
```diff
@@ -19,6 +24,187 @@
...
+    else:
+        torch_version = None
+    with open(version_path, 'a') as f:
+        f.write(f'\n'
+                f"gcc_version = '{gcc_version}'\n"
+                f"cuda_version = '{cuda_version}'\n"
+                f"torch_version = '{torch_version}'\n"
+                f"git_version = '{git_hash}'\n")
```
zhouzaida, line 201:
```suggestion
                f"commit_id = '{git_hash}'\n")
```

.github/workflows/build.yml
```diff
@@ -346,7 +356,7 @@ jobs:
           rm -rf .eggs
           CC=clang CXX=clang++ CFLAGS='-stdlib=libc++' pip install -e .
       - name: Validate the installation
-        run: python -c "import mmcv"
+        run: python -c "import mmcv; print(f'GCC {mmcv.version.gcc_version}', f'CUDA {mmcv.version.cuda_version}', f'PyTorch {mmcv.version.torch_version}', f'Git {mmcv.version.commit_id}', sep='\n')"
```
zhouzaida, line 359: Git -> CommitID

setup.py
```diff
@@ -19,6 +24,189 @@
...
+        except OSError:
+            sha = fallback
+
+        return sha
+
+    mmcv_root = Path(__file__).parent
+    version_path = mmcv_root / 'mmcv' / 'version.py'
+    gcc_version = get_gcc()
+    cuda_version = get_cuda()
+    git_hash = get_git_hash()
```
zhouzaida, line 191:
```suggestion
    git_hash = get_git_hash(digits=7)
```
We just need to keep the first 7 characters.

setup.py
```diff
@@ -19,6 +24,189 @@
...
+            sha = out.strip().decode('ascii')
+            if digits is not None:
+                sha = sha[:digits]
+        except OSError:
+            sha = fallback
+
+        return sha
+
+    mmcv_root = Path(__file__).parent
+    version_path = mmcv_root / 'mmcv' / 'version.py'
```
zhouzaida, line 188:
```python
import os.path as osp
...
version_path = osp.join(mmcv_root, 'mmcv', 'version.py')
```

And replace all the `os.path` with `osp`.

setup.py
```diff
@@ -19,6 +24,189 @@
...
+            for name in fnames:
+                if os.access(name,
+                             os.F_OK | os.X_OK) and not os.path.isdir(name):
+                    return name
+        return None
+
+    def find_nvcc():
+        nvcc = which('nvcc')
+        if nvcc is not None:
+            return os.path.dirname(nvcc)
```
zhouzaida, line 62:
```python
import os.path as osp
```

Replace all the `os.path` with `osp`.

setup.py
```diff
@@ -19,6 +24,189 @@
     print('Skip building ext ops due to the absence of torch.')
 
 
+def add_version_info():
+    """Modified from
+    https://github.com/pytorch/pytorch/blob/v1.3.1/tools/setup_helpers."""
```
zhouzaida, line 29:
```suggestion
    # Modified from
    # https://github.com/pytorch/pytorch/blob/v1.3.1/tools/setup_helpers.
```

setup.py
```diff
@@ -19,6 +24,189 @@
...
+                         default).upper() in ['ON', '1', 'YES', 'TRUE', 'Y']
+
+    def check_negative_env_flag(name, default=''):
+        return os.getenv(name,
+                         default).upper() in ['OFF', '0', 'NO', 'FALSE', 'N']
+
+    def which(thefile):
+        path = os.environ.get('PATH', os.defpath).split(os.pathsep)
+        for d in path:
+            fname = os.path.join(d, thefile)
```
zhouzaida, line 48:
```suggestion
            fname = osp.join(d, thefile)
```

setup.py
```diff
@@ -19,6 +24,189 @@
...
+        if check_negative_env_flag('USE_CUDA') or check_env_flag('USE_ROCM'):
+            cuda_version = None
+        else:
+            if IS_LINUX or IS_DARWIN:
+                CUDA_HOME = os.getenv('CUDA_HOME', LINUX_HOME)
+            else:
+                CUDA_HOME = os.getenv('CUDA_PATH', '').replace('\\', '/')
+                if CUDA_HOME == '' and len(WINDOWS_HOME) > 0:
+                    CUDA_HOME = WINDOWS_HOME[0].replace('\\', '/')
+            if not os.path.exists(CUDA_HOME):
```
zhouzaida, line 118:
```suggestion
            if not osp.exists(CUDA_HOME):
```

setup.py
```diff
@@ -19,6 +24,189 @@
...
+                # We use nvcc path on Linux and cudart path on macOS
+                if IS_LINUX or IS_WINDOWS:
+                    cuda_path = find_nvcc()
+                else:
+                    cudart_path = ctypes.util.find_library('cudart')
+                    if cudart_path is not None:
+                        cuda_path = os.path.dirname(cudart_path)
```
zhouzaida, line 125:
```suggestion
                        cuda_path = osp.dirname(cudart_path)
```

setup.py
```diff
@@ -19,6 +24,189 @@
...
+                    else:
+                        cuda_path = None
+                if cuda_path is not None:
+                    CUDA_HOME = os.path.dirname(cuda_path)
```
zhouzaida, line 129:
```suggestion
                    CUDA_HOME = osp.dirname(cuda_path)
```

setup.py
```diff
@@ -19,6 +24,189 @@
...
+        torch_version = None
+    with open(version_path, 'a') as f:
+        f.write(f'\n'
+                f"gcc_version = '{gcc_version}'\n"
+                f"cuda_version = '{cuda_version}'\n"
+                f"torch_version = '{torch_version}'\n"
+                f"commit_id = '{git_hash}'\n")
+
+
+if 'egg_info' not in sys.argv:
```
zhouzaida, line 206: Please add a comment to describe which command will make this condition true?

.github/workflows/build.yml
teamwong111: done

setup.py
teamwong111: done

setup.py
teamwong111: done

setup.py
teamwong111: done

setup.py
teamwong111: done

zhouzaida (approved):

setup.py
```diff
@@ -19,6 +25,191 @@
...
+
+    mmcv_root = Path(__file__).parent
+    version_path = mmcv_root / 'mmcv' / 'version.py'
+    gcc_version = get_gcc()
+    cuda_version = get_cuda()
+    git_hash = get_git_hash(digits=7)
+    if EXT_TYPE:
+        torch_version = torch.__version__
+        if '+' in torch_version:  # eg. 1.8.1+cu111 -> 1.8.1
+            torch_version = torch_version.split('+')[0]
```
zhouzaida, line 195: Maybe the torch_version should be the format like `1.8.1+cu111`.

.github/workflows/build.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: Build and install
         run: rm -rf .eggs && pip install -e .
       - name: Validate the installation
-        run: python -c "import mmcv"
+        run: python -c "import mmcv; print(f'GCC {mmcv.version.gcc_version}', f'CUDA {mmcv.version.cuda_version}', f'PyTorch {mmcv.version.torch_version}', f'CommitID {mmcv.version.commit_id}', sep='\n')"
```
ZwwWayne, line 46: we can directly call collect_env.py for simplicity

ZwwWayne:
May post the results in the PR message, e.g., what is the new verion.py looks like? Furthermore, can we do it more easily by using the existing collect_env?

.github/workflows/build.yml
teamwong111: We run this command just to check whether the version info we get is correct on different OS.

teamwong111:
> May post the results in the PR message, e.g., what is the new verion.py looks like? Furthermore, can we do it more easily by using the existing collect_env?

The results have been posted in the PR message. We cannot use the existing collect_env because collect_env relies on C++ or CUDA which will work after build.

zhouzaida:
@ZwwWayne 

We need to record the compilation information in `version.py`, because it doesn't make sense to record it in `mmcv.ops`. Users may encounter `undefine symbol` errors when importing operators from `mmcv.ops`, and this error is usually due to inconsistencies between the cuda version used to compile mmcv-full and the runtime version, so we need the user to provide the compile-time cuda information, which is not available from mmcv.ops due to the `undefine symbol` errors. So it is necessary to record the cuda information in `version.py`.

So did you mean that "Can we simplify the way to get the gcc and cuda versions using `from mmcv.ops import get_compiler_version, get_compiling_cuda_version`"

setup.py
teamwong111: done

ZwwWayne:
> 

yes

zhouzaida:
**Windows**
```
  File "setup.py", line 220, in <module>
    add_version_info()
  File "setup.py", line 189, in add_version_info
    gcc_version = get_gcc()
  File "setup.py", line 138, in get_gcc
    stderr=subprocess.PIPE)
  File "D:\miniconda3\envs\py37-pt18-cu111\lib\subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "D:\miniconda3\envs\py37-pt18-cu111\lib\subprocess.py", line 1207, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] 系统找不到指定的文件。
```

setup.py
```diff
@@ -20,6 +25,201 @@
...
+                torch_version += '+cu'
+                torch_version += (torch.version.cuda).replace('.', '')
+            else:
+                torch_version += '+cpu'
+    else:
+        torch_version = None
+    with open(version_path, 'r+') as f:
+        lines = f.readlines()
+        last_line = lines[-1]
+        if '__all__' not in last_line:
```
zhouzaida, line 205: It maybe better to put those information after `__version__` rather than `__all__`

setup.py
```diff
@@ -20,6 +25,201 @@
...
+                    CUDA_HOME = osp.dirname(cuda_path)
+                else:
+                    CUDA_HOME = None
+            cuda_version = find_cuda_version(CUDA_HOME)
+        return cuda_version
+
+    def get_gcc():
+        proc = subprocess.Popen(['gcc', '--version'],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
```
zhouzaida, line 138: image

Issue #1163, opened 2021-06-30
[bug] val loss log not show using IterBasedRunner


hhaAndroid:
As LoggerHook does not implement the after_val_iter method, the val loss cannot be displayed, please fix it as soon as possible.
please refer to https://github.com/open-mmlab/mmdetection/issues/5382

PR #1495, opened 2021-11-16
[Fix] fix the bug val loss log not show using IterBasedRunner

teamwong111:

## Motivation

Related Issue: https://github.com/open-mmlab/mmcv/issues/1163
Fix the bug val loss log not show using `IterBasedRunner`

## Modification

Impl `after_val_iter` function in `LoggerHook`. It should be noted that this function only works when the corresponding value of `workflow.val` is greater than the log config `interval` or `len(data_loader)`

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/hooks/logger/base.py
+++ b/mmcv/runner/hooks/logger/base.py
@@ -153,6 +153,18 @@ class LoggerHook(Hook):
             if self.reset_flag:
                 runner.log_buffer.clear_output()
 
+    def after_val_iter(self, runner):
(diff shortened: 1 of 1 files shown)


CptCaptain:
Does this PR have any chance of getting merged anytime soon?
Having to choose between using an iteration based schedule and logging validation losses seems... suboptimal, especially when the rest of the library is so great to use.

Issue #988, opened 2021-04-24
Progress bar does not produce a new line after it finishes


wangruohui:
Currently it looks like this. Two line linked together.

```
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 715/715, 21.7 task/s, elapsed: 33s, ETA:     0s2021-04-25 01:52:58,022 - mmcls - INFO - Epoch(val) [1][89] accuracy_top-1: 92.4476, accuracy_top-5: 100.0000
```

Expected:

```
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 715/715, 21.7 task/s, elapsed: 33s, ETA:     0s
2021-04-25 01:52:58,022 - mmcls - INFO - Epoch(val) [1][89] accuracy_top-1: 92.4476, accuracy_top-5: 100.0000
```

teamwong111:
Hi, @wangruohui , apologies for the delay. It seems to have no problem just from the code. Do you use ProgressBar class and customize some functions? Or could you please provide more details?

wangruohui:
No customization in my memory. 
We can close the issue, if updated code does not show this error. Thank you.

PR #1496, opened 2021-11-16
[Fix] fix progress bar lacks a wrap

teamwong111:

## Motivation

Related closed issue: https://github.com/open-mmlab/mmcv/issues/988.
Fix the progress bar lacks a wrap.

## Modification

Add wrap in the corresponding function.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/engine/test.py
+++ b/mmcv/engine/test.py
@@ -38,6 +38,7 @@ def single_gpu_test(model, data_loader):
         batch_size = len(result)
         for _ in range(batch_size):
             prog_bar.update()
+    prog_bar.file.write('\n')
     return results
 
 
@@ -79,6 +80,7 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
                 batch_size_all = len(dataset) - prog_bar.completed
             for _ in range(batch_size_all):
                 prog_bar.update()
+    prog_bar.file.write('\n')
 
     # collect results from all ranks
     if gpu_collect:


zhouzaida (approved):

mmcv/engine/test.py
```diff
@@ -64,21 +64,20 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
...
         results.extend(result)
-
-        if rank == 0:
-            batch_size = len(result)
-            batch_size_all = batch_size * world_size
-            if batch_size_all + prog_bar.completed > len(dataset):
-                batch_size_all = len(dataset) - prog_bar.completed
-            for _ in range(batch_size_all):
-                prog_bar.update()
+    if rank == 0:
```
ZwwWayne, line 72: Originally the there is only one progressbar, in the new code, it seems that progressbar is created by multiple times?

ZwwWayne:
Does this PR introduces BC-breaking? For example, there are many code in downstream repos using the old style like
```
if rank == 0:
        prog_bar = mmcv.ProgressBar(len(dataset))
for I in xxx:
    prog_bar.update()
```
What will happen if we do not modify our usage of progressbar?

mmcv/engine/test.py
teamwong111: I think the new and old will create progressbar the same number of times.

teamwong111:
> Does this PR introduces BC-breaking? For example, there are many code in downstream repos using the old style like
> 
> ```
> if rank == 0:
>         prog_bar = mmcv.ProgressBar(len(dataset))
> for I in xxx:
>     prog_bar.update()
> ```
> 
> What will happen if we do not modify our usage of progressbar?

Honestly, I am not quite sure about this PR introduces BC-breaking because some downstream repos impl their own `single_gpu_test` and `multi_gpu_test` but maybe other downstream repos use these func in mmcv. However, we should ask them to use the progressbar according to the new Impl of progressbar.

mmcv/engine/test.py
zhouzaida: In fact, the new code is not right.
Old code updates the progress bar after one iteration. But new code updates it at the end of iterating the whole dataset which is not an expected behavior.

zhouzaida:
> Does this PR introduces BC-breaking? For example, there are many code in downstream repos using the old style like
> 
> ```
> if rank == 0:
>         prog_bar = mmcv.ProgressBar(len(dataset))
> for I in xxx:
>     prog_bar.update()
> ```
> 
> What will happen if we do not modify our usage of progressbar?

I think it will not introduce BC-breaking. In fact, the PR just adds a `\n` character aotumatically if we use the `with` clause to manage the context. If we do not use the `with` clause, everything is the same as before.

mmcv/engine/test.py
teamwong111: I will fix it.

ZwwWayne (approved):

ZwwWayne:
Ready to merge, we can choose either adding documentations in this PR or update it in another PR.

teamwong111:
> Ready to merge, we can choose either adding documentations in this PR or update it in another PR.

What kind of docs, about what we modified and how downstream modify? And where should we put the docs, compatibility.md?

ZwwWayne:
> 

Sorry for the vague comments. Shall we update this documentation https://github.com/open-mmlab/mmcv/blob/master/docs/en/understand_mmcv/utils.md#progressbar to tell users about the new and old usages of progressbar?

teamwong111:
> > 
> 
> Sorry for the vague comments. Shall we update this documentation https://github.com/open-mmlab/mmcv/blob/master/docs/en/understand_mmcv/utils.md#progressbar to tell users about the new and old usages of progressbar?

In the above docs we only provide the usages of `track_progress`, `track_parallel_progress` and `track_iter_progress`, their usages and the previous remain the same.

PR #1504, opened 2021-11-18
Add extension ops in parrots

luopeichao:
Add extension ops (group_points, iou3d, roiaware_pool3d and voxelize) in parrots.


grimoire (approved):
LGTM.

Did parrots has some limit about function? like no return or tensor args first? We will try avoid when we add new ops in future.

luopeichao:
> LGTM.
> 
> Did parrots has some limit about function? like no return or tensor args first? We will try avoid when we add new ops in future.

Yes. Now we need no return or return is only a Tensor. And we need input tensor args fisrt, then output tensor args, finally other args. Other args need to be given by using kwargs, like f_op(input, output, ndim=x) rather than f_op(input, output, x).

ZwwWayne:
Lint failed. Can be merged after fixing lint issues

PR #1524, opened 2021-11-24
[Fix] Fix iou3d python bind

grimoire:
## Motivation

`iou3d_nms_forward` and `iou3d_nms_normal_forward` have different function declaration between `pybind.cpp` and `iou3d.cpp`.

## Modification

Remove return value of `iou3d_nms_forward` and `iou3d_nms_normal_forward`.


zhouzaida (approved):

zhouzaida:
Verified on Windows

PR #1525, opened 2021-11-25
[Fix] Fix build with cuda9.0

grimoire:
## Motivation

Can not build mmcv-full with cuda9.0.

Reference
https://github.com/pytorch/extension-cpp/issues/35

## Modification

Replace `torch/extension.h` with `torch/types.h`

### DIFF
--- a/mmcv/ops/csrc/common/cuda/correlation_cuda.cuh
+++ b/mmcv/ops/csrc/common/cuda/correlation_cuda.cuh
@@ -14,7 +14,7 @@
 
 #include <cuda.h>
 #include <cuda_runtime.h>
-#include <torch/extension.h>
+#include <torch/types.h>
 
 #include <iostream>
 #include <vector>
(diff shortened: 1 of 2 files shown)


teamwong111:
Checked.

zhouzaida (approved):

ZwwWayne (approved):
shall we leave a note for that in case we cause the similar bug in the future？

grimoire:
@ZwwWayne sure. Where should I put the note? In code?

grimoire:
By the way, I prefer `<torch/extension.h>` since this seems like the official way to add custom ops.
https://pytorch.org/tutorials/advanced/cpp_extension.html

ZwwWayne:
We can add a comment just in the line above line 17. I also thought we should use torch.extension.h, so I think we should notify other developers in the code. An example can be as below:
```
Using <torch/extension.h> is recommended in the official documentation in https://pytorch.org/tutorials/advanced/cpp_extension.html#writing-the-c-op.
However, we use torch/types.h for compatibility with CUDA 9.0
```

PR #1527, opened 2021-11-26
[Feature] Add multi file backends to imread/imwrite.

274869388:
## Motivation

The motivation of the PR is to read image from different file backends and write image to different file backends but do not need to modify the existing interface.

It includes two parts:
- Add `FileClient` to image io function to support different file backends.
- Add unittest to `tests/test_image/test_io.py`.

## Modification

1. imread
    - Add a new parameter `file_client_args` to load which makes it support loading object from different file backends, but the parameter is optional
    - Add `FileClient` to read image bytes of different file backends and use `imfrombytes` to decode image.
    - Add `tifffile` backends to `imfrombytes`.
 
2. imwrite
    - Add a new parameter `file_client_args` to load which makes it support loading object from different file backends, but the parameter is optional
    - Add `cv2.imencode` to encode image and use `FileClient` to write image bytes to different file backends.
    - Warn when image writing fails.

3. test_io.py
    - Add `MagicMock` to test `imread` and `imwrite`.

## BC-breaking (Optional)

No

## Use cases (Optional)

```python
import mmcv

disk_path = '/path/of/your/file'
img = mmcv.imread(disk_path)
mmcv.imwrite(img, disk_path)

s3_path = 's3://your_bucket/img.png'
img = mmcv.imread(s3_path)
img = mmcv.imread(s3_path, file_client_args={'backend': 'petrel'})
img = mmcv.imread(s3_path, file_client_args={'prefix': 's3'})
mmcv.imwrite(img, s3_path)
mmcv.imwrite(img, s3_path, file_client_args={'backend': 'petrel'})
mmcv.imwrite(img, s3_path, file_client_args={'prefix': 's3'})

http_path = 'http://path/of/your/file'
img = mmcv.imread(http_path)
img = mmcv.imread(http_path, file_client_args={'backend': 'http'})
img = mmcv.imread(http_path, file_client_args={'prefix': 'http'})
mmcv.imwrite(img, http_path)
...
```

## Test
- [x] disk backend works as expected
- [x] petrel backend works as expected

## Checklist

**Before PR**:

**After PR**:


mmcv/image/io.py
```diff
@@ -137,9 +139,16 @@ def _pillow2array(img, flag='color', channel_order='bgr'):
...
-def imread(img_or_path, flag='color', channel_order='bgr', backend=None):
+def imread(img_or_path,
+           flag='color',
+           channel_order='bgr',
+           backend=None,
+           file_client_args=None):
     """Read an image.
 
+    Note:
+        In v1.3.17 and later, add `file_client_args` parameters.
```
teamwong111, line 150: Maybe v1.3.17 --> v1.3.19

mmcv/image/io.py
```diff
@@ -157,44 +166,29 @@ def imread(img_or_path, flag='color', channel_order='bgr', backend=None):
...
-                cv2.cvtColor(img, cv2.COLOR_BGR2RGB, img)
-            return img
+        file_client = FileClient.infer_client(file_client_args, img_or_path)
+        # TODO Some file_backends is not supporting to exists function.
+        # If you want to use petrel_client, you must ensure that
+        # the version >= v2.2.0.
+        if hasattr(file_client.client,
+                   'exists') and not file_client.exists(img_or_path):
+            raise FileNotFoundError(f'img file does not exist: {img_or_path}')
+        img_bytes = file_client.get(img_or_path)
```
teamwong111, line 190: What will happen if I use `file_client` which does not support `exists` function and `img_or_path` which does not exist?

mmcv/image/io.py
```diff
@@ -239,20 +238,46 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
-def imwrite(img, file_path, params=None, auto_mkdir=True):
+def imwrite(img,
+            file_path,
+            params=None,
+            auto_mkdir=True,
+            file_client_args=None):
     """Write image to file.
 
+    Note:
+        In v1.3.17 and later, add `file_client_args` parameters.
```
teamwong111, line 249: Maybe v1.3.17 --> v1.3.19

teamwong111 (commented):
LGTM but with some slight doubt.
1. Is it necessary to add some novel unittests for `imwrite` with other file backend?
2. Shall we refactor test file which looks like not elegant now?
3. Shall we impl `exists` function for all File backend before merging this PR?

mmcv/image/io.py
274869388: Yes, I will modify this problem in the next commit.

mmcv/image/io.py
274869388: Yes, I will modify this problem in the next commit.

mmcv/image/io.py
274869388: It will skip judgment and throw some unknown errors.For example, `http_client` may throw ConnectionError.

274869388:
> LGTM but with some slight doubt.
> 
> 1. Is it necessary to add some novel unittests for `imwrite` with other file backend?
> 2. Shall we refactor test file which looks like not elegant now?
> 3. Shall we impl `exists` function for all File backend before merging this PR?

1. I think there is no need to test other file backends because they have been fully tested in test_fileio.py.
2. I may not be ready to refactor the test file for now.
3. We plan to impl the `exists` function for all file backend in the future,but not now.

mmcv/image/io.py
274869388: I think we should throw exceptions in the `get` method instead of manually judging various exceptions.so we deleted this judgment.

teamwong111 (approved):
LGTM

zhouzaida:
Maybe we can also add multi-file backends support for the [flowread](https://github.com/open-mmlab/mmcv/blob/7207397de3c2ffaef878628a55fa2ab37ac0bbc7/mmcv/video/optflow.py#L12) and [flowwrite](https://github.com/open-mmlab/mmcv/blob/7207397de3c2ffaef878628a55fa2ab37ac0bbc7/mmcv/video/optflow.py#L61)

MeowZheng:
> Maybe we can also add multi-file backends support for the [flowread](https://github.com/open-mmlab/mmcv/blob/7207397de3c2ffaef878628a55fa2ab37ac0bbc7/mmcv/video/optflow.py#L12) and [flowwrite](https://github.com/open-mmlab/mmcv/blob/7207397de3c2ffaef878628a55fa2ab37ac0bbc7/mmcv/video/optflow.py#L61)

Of course, but I think we should create another pr for it. I will take this work very recently.

MeowZheng (approved):

mmcv/image/io.py
```diff
@@ -157,44 +166,40 @@ def imread(img_or_path, flag='color', channel_order='bgr', backend=None):
...
+        >>> img_path = '/path/to/img.jpg'
+        >>> img = mmcv.imread(img_path)
+        >>> img = mmcv.imread(img_path, flag='color', channel_order='rgb',
+        ...     backend='cv2')
+        >>> img = mmcv.imread(img_path, flag='color', channel_order='bgr',
+        ...     backend='pillow')
+        >>> s3_img_path = 's3://bucket/img.jpg'
+        >>> img = mmcv.imread(s3_img_path)
+        >>> img = mmcv.imread(s3_img_path, file_client_args={
+        ...     'backend': 'petrel'})
```
zhouzaida, line 187:
```suggestion
        >>> s3_img_path = 's3://bucket/img.jpg'
        >>> # infer the file backend by the prefix s3
        >>> img = mmcv.imread(s3_img_path)
        >>> # manually set the file backend petrel
        >>> img = mmcv.imread(s3_img_path, file_client_args={
        ...     'backend': 'petrel'})
```

mmcv/image/io.py
```diff
@@ -239,20 +258,52 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+        >>> ret = mmcv.imwrite(img, 's3://bucket/img.jpg', file_client_args={
+        ...     'backend': 'petrel'})
     """
-    if auto_mkdir:
-        dir_name = osp.abspath(osp.dirname(file_path))
-        mkdir_or_exist(dir_name)
-    return cv2.imwrite(file_path, img, params)
+    assert is_str(file_path)
+    file_client = FileClient.infer_client(file_client_args, file_path)
+    if not hasattr(file_client.client, 'put'):
```
zhouzaida, line 292: replace hasattr with hasmethod https://github.com/open-mmlab/mmcv/blob/72182747a17231075b26768d37694781bd992daf/mmcv/utils/misc.py#L367

mmcv/image/io.py
```diff
@@ -239,20 +258,52 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+        raise AttributeError(
+            f"{file_client.name} doesn't contain the `put` method")
+    try:
+        img_ext = osp.splitext(file_path)[-1]
+        _, img_buff = cv2.imencode(img_ext, img, params)
+        # The HardDiskBackend automatically mkdir by default,it needs to be
+        # judged individually
+        if not auto_mkdir and file_client.name == 'HardDiskBackend':
```
zhouzaida, line 300: Note that `file_client.put` should follow a Protocol that makes directory automatically if needed.

tests/test_image/test_io.py
```diff
@@ -315,6 +352,9 @@ def test_imwrite(self):
             img, './non_exist_path/mmcv_test.jpg', auto_mkdir=False)
         assert ret is False
```
zhouzaida, line 354: A test for petrel should be added.

mmcv/image/io.py
```diff
@@ -207,28 +212,42 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+        >>> img = mmcv.imfrombytes(img_buff, backend='pillow')
+        >>> img = mmcv.imfrombytes(img_buff, backend='cv2')
     """
 
     if backend is None:
         backend = imread_backend
     if backend not in supported_backends:
-        raise ValueError(f'backend: {backend} is not supported. Supported '
-                         "backends are 'cv2', 'turbojpeg', 'pillow'")
+        raise ValueError(
```
zhouzaida, line 235: A test for the line should be added in `test_imfrombytes`

mmcv/image/io.py
274869388: ok

mmcv/image/io.py
274869388: yes,i will fix it.

tests/test_image/test_io.py
274869388: ok

mmcv/image/io.py
274869388: It is tested in test_io.py 350 line.

mmcv/image/io.py
```diff
@@ -239,20 +261,59 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+        >>> ret = mmcv.imwrite(img, 's3://bucket/img.jpg')
+        >>> # manually set the file backend petrel
+        >>> ret = mmcv.imwrite(img, 's3://bucket/img.jpg', file_client_args={
+        ...     'backend': 'petrel'})
     """
-    if auto_mkdir:
-        dir_name = osp.abspath(osp.dirname(file_path))
-        mkdir_or_exist(dir_name)
-    return cv2.imwrite(file_path, img, params)
+    assert is_str(file_path)
```
zhouzaida, line 300: file_path can also be a Path object.

mmcv/image/io.py
```diff
@@ -239,20 +261,59 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+    if auto_mkdir is not None:
```
zhouzaida, line 301: This line probably is redundant. The optional values of `auto_mkdir` are `True` or `False`.

mmcv/image/io.py
```diff
@@ -239,20 +261,59 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+            f"{file_client.name} doesn't contain the `put` method")
+    try:
+        img_ext = osp.splitext(file_path)[-1]
+        # Encode image according to image suffix.
+        # For example, if image path is '/path/your/img.jpg', the encode
+        # format is '.jpg'.
+        _, img_buff = cv2.imencode(img_ext, img, params)
+        file_client.put(img_buff.tobytes(), file_path)
+    except Exception as e:
+        warnings.warn(f"'{file_path}' writing failed, msg is '{e}'")
```
zhouzaida, line 317:
```suggestion
    except Exception as error:
        warnings.warn(error)
```

mmcv/image/io.py
```diff
@@ -239,20 +261,59 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
-    return cv2.imwrite(file_path, img, params)
+    assert is_str(file_path)
+    if auto_mkdir is not None:
+        warnings.warn(
+            'The parameter `auto_mkdir` will be deprecated in the future and '
+            'every file clients will make directory automatically.')
+    file_client = FileClient.infer_client(file_client_args, file_path)
+    if not has_method(file_client.client, 'put'):
+        raise AttributeError(
+            f"{file_client.name} doesn't contain the `put` method")
```
zhouzaida, line 308: We can remove the checking for `put`. `Fileclient` will throw an error if the file backend does not implement the method.

mmcv/image/io.py
274869388: ok

mmcv/image/io.py
274869388: This judgment is to confirm whether the user has entered the `auto_mkdir` parameter, and only output a warning if the user actively enters it.

mmcv/image/io.py
274869388: yes

mmcv/image/io.py
zhouzaida: got it

tests/test_image/test_io.py
```diff
@@ -311,9 +357,15 @@ def test_imwrite(self):
         os.remove(out_file)
         self.assert_img_equal(img, rewrite_img)
 
-        ret = mmcv.imwrite(
```
zhouzaida, line 314: why the test is removed?

274869388: Because auto_mkdir is deprecated, every file client will make directory automatically, this test is meaningless.

zhouzaida: got it

zhouzaida (approved):

mmcv/image/io.py
```diff
@@ -239,20 +260,57 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
+    file_client = FileClient.infer_client(file_client_args, file_path)
+    try:
+        img_ext = osp.splitext(file_path)[-1]
+        # Encode image according to image suffix.
+        # For example, if image path is '/path/your/img.jpg', the encode
+        # format is '.jpg'.
+        _, img_buff = cv2.imencode(img_ext, img, params)
+        file_client.put(img_buff.tobytes(), file_path)
+    except Exception as error:
+        warnings.warn(error)
```
zhouzaida, line 314: please @ZwwWayne have a look. If the `imwrite` occurred exception, is it useful to help users debug just printing warning information?

ZwwWayne: I do not think we should handle that because we should let the exact place to replace its error rather than leave it in later with potentially misleading error messages.

274869388: Because this interface specifies return the bool type, and it does not throw an exception, it may cause the interface inconsistent issue.

ZwwWayne: Then should it raise a warning of error and directly return False?

274869388: yes

274869388: Sorry, in cv2.imwrite, if file_path extension is error, it will throw an exception. 
```python
>>> cv2.imwrite('1.jpgg', np.zeros((2,2)))
  File "<stdin>", line 1, in <module>
cv2.error: OpenCV(4.5.4) /tmp/pip-req-build-3129w7z7/opencv/modules/imgcodecs/src/loadsave.cpp:728: error: (-2:Unspecified error) could not find a writer for the specified extension in function 'imwrite_'
```
I will fix it error.

mmcv/image/io.py
```diff
@@ -239,20 +260,53 @@ def imfrombytes(content, flag='color', channel_order='bgr', backend=None):
...
-        mkdir_or_exist(dir_name)
-    return cv2.imwrite(file_path, img, params)
+    assert is_filepath(file_path)
+    file_path = str(file_path)
+    if auto_mkdir is not None:
+        warnings.warn(
+            'The parameter `auto_mkdir` will be deprecated in the future and '
+            'every file clients will make directory automatically.')
+    file_client = FileClient.infer_client(file_client_args, file_path)
+    img_ext = osp.splitext(file_path)[-1]
```
zhouzaida, line 306: We should add a unit test for the case that the extension of the file is not expected.

274869388: done

PR #1541, opened 2021-11-30
[Feature]Detect anomalous parameters 

jshilong:
## Motivation

Sometimes there are some anomalous parameters in the training phase, mainly two cases 

- Some parameters are  not be used in a forward pass due to our wrong implementation, which causes the error when doing `loss.backward()`
- Some parameters are not be used to produce loss, which causes the deadlock when doing `loss.backward()`
 
In both cases，anomalous parameters are not included in the computational graph that is with `loss` as the root.

## Modification

I add a debug option named `detect_anomalous_params`  to `OptimizerHook`, which can help you find anomalous parameters  

## BC-breaking (Optional)
None

### DIFF
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -24,6 +24,7 @@ class OptimizerHook(Hook):
 
     Args:
         grad_clip (dict): A config dict to control the clip_grad.
+            Default: None.
         detect_anomalous_params (bool):  Detect anomalous parameters
             of the model. Such as
 
@@ -31,9 +32,10 @@ class OptimizerHook(Hook):
                   forward pass.
                 - Parameters were not used to produce
                   loss.
+            Default: False.
     """
 
-    def __init__(self, grad_clip=None, detect_anomalous_params=True):
+    def __init__(self, grad_clip=None, detect_anomalous_params=False):
 
         self.grad_clip = grad_clip
         self.detect_anomalous_params = detect_anomalous_params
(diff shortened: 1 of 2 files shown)


mmcv/runner/hooks/optimizer.py
```diff
@@ -41,6 +62,32 @@ def after_train_iter(self, runner):
...
+                visited.add(grad_fn)
+            else:
+                return
+            if hasattr(grad_fn, 'variable'):
+                parameters_in_graph.add(grad_fn.variable)
+            parents = grad_fn.next_functions
+            if parents is not None:
+                for parent in parents:
+                    grad_fn = parent[0]
+                    traverse(grad_fn)
```
MeowZheng, line 83: insert these lines in `if grad_fn not in visited` statement after `visited.add(grad_fn)`

MeowZheng (commented):
Add unit tests

MeowZheng (approved):
LGTM

mmcv/runner/hooks/optimizer.py
```diff
@@ -20,9 +20,27 @@
 
 @HOOKS.register_module()
 class OptimizerHook(Hook):
+    """A hook contains custom operations for the optimizer.
+
+    Args:
+        grad_clip (dict): A config dict to control the clip_grad.
+            Default: None.
+        detect_anomalous_params (bool):  Detect anomalous parameters
+            of the model which is not included in
```
RangiLyu, line 29:
```suggestion
            of the model which are not included in
```

mmcv/runner/hooks/optimizer.py
```diff
@@ -41,6 +62,30 @@ def after_train_iter(self, runner):
...
+                parents = grad_fn.next_functions
+                if parents is not None:
+                    for parent in parents:
+                        grad_fn = parent[0]
+                        traverse(grad_fn)
+
+        traverse(loss.grad_fn)
+        for n, p in runner.model.named_parameters():
+            if p not in parameters_in_graph and p.requires_grad:
+                logger.info(f'{n} with shape {p.size()} is not '
```
RangiLyu, line 86: What if some params are not used on other ranks except rank0?

RangiLyu (approved):

PR #1547, opened 2021-12-01
[Feature] Detect anomalous parameters

jshilong:
## Motivation

Sometimes there are some anomalous parameters in the training phase, mainly two cases 

- Some parameters are  not be used in a forward pass due to our wrong implementation, which causes the error when doing `loss.backward()`
- Some parameters are not be used to produce loss, which causes the deadlock when doing `loss.backward()`
 
In both cases，anomalous parameters are not included in the computational graph that is with `loss` as the root.

## Modification

I add a debug option named `detect_anomalous_params`  to `OptimizerHook`, which can help you find anomalous parameters  

## BC-breaking (Optional)
None

### DIFF
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -26,8 +26,9 @@ class OptimizerHook(Hook):
     Args:
         grad_clip (dict, optional): A config dict to control the clip_grad.
             Default: None.
-        detect_anomalous_params (bool):  Detect anomalous parameters
-            of the model which are not included in
+        detect_anomalous_params (bool): This option is only used for
+            debugging which will slow down the training speed.
+            Detect anomalous parameters that are not included in
             the computational graph with `loss` as the root.
             There are two cases
 
--- a/mmcv/runner/hooks/optimizer.py
+++ b/mmcv/runner/hooks/optimizer.py
@@ -24,6 +24,7 @@ class OptimizerHook(Hook):
 
     Args:
         grad_clip (dict): A config dict to control the clip_grad.
+            Default: None.
         detect_anomalous_params (bool):  Detect anomalous parameters
             of the model. Such as
 
@@ -31,9 +32,10 @@ class OptimizerHook(Hook):
                   forward pass.
                 - Parameters were not used to produce
                   loss.
+            Default: False.
     """
 
-    def __init__(self, grad_clip=None, detect_anomalous_params=True):
+    def __init__(self, grad_clip=None, detect_anomalous_params=False):
 
         self.grad_clip = grad_clip
         self.detect_anomalous_params = detect_anomalous_params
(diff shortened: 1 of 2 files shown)


jshilong:
Ping @ZwwWayne

tests/test_runner/test_hooks.py
```diff
@@ -39,6 +39,85 @@
 sys.modules['petrel_client.client'] = MagicMock()
 
 
+def test_optimizerhook():
+
+    class Model(nn.Module):
```
ZwwWayne, line 44: The unit test only test the case when conv3 is not used in the forwarwd pass. We should also test the case when conv3 is forwarded but not used for computing loss.

jshilong: This case has been tested when the loss is `x1.sum()` then conv2 would be detected. Line 98

mmcv/runner/hooks/optimizer.py
```diff
@@ -20,9 +21,27 @@
 
 @HOOKS.register_module()
 class OptimizerHook(Hook):
+    """A hook contains custom operations for the optimizer.
+
+    Args:
+        grad_clip (dict): A config dict to control the clip_grad.
+            Default: None.
```
MeowZheng, line 28: (dict) -> (dict, optional)
when the default value is `None`,  the type of keyword argument can be `optional`.

zhouzaida (approved):

zhouzaida:
After resolving the above comments and the CI passes, the PR can be merged.

MeowZheng (approved):

ZwwWayne:
Can we check whether it slows down the training speed if we set detect_anomalous_params=True?

zhouzaida:
> Can we check whether it slows down the training speed if we set detect_anomalous_params=True? This PR can be merged after checking the influence on training speed and updating this information in docstring.

It probably slows down the training speed but the flag should only be opened for debugging.

zhouzaida:
Please merge the upstream master branch to this PR which will resolve the error when building documentation.

jshilong:
> > Can we check whether it slows down the training speed if we set detect_anomalous_params=True? This PR can be merged after checking the influence on training speed and updating this information in docstring.
> 
> It probably slows down the training speed but the flag should only be opened for debugging.



> Can we check whether it slows down the training speed if we set detect_anomalous_params=True? This PR can be merged after checking the influence on training speed and updating this information in docstring.

It is only used for debugging

jshilong:
> Can we check whether it slows down the training speed if we set detect_anomalous_params=True? This PR can be merged after checking the influence on training speed and updating this information in docstring.

I will add a notice in the docstr to indicate this is only used for debugging

zhouzaida:
> I will add a notice in the docstr to indicate this is only used for debugging

Hi @jshilong , we will merge the PR after you add a notice in the docstring.

ZwwWayne (approved):

jshilong:
> > I will add a notice in the docstr to indicate this is only used for debugging
> 
> Hi @jshilong , we will merge the PR after you add a notice in the docstring.

Done

zhouzaida (approved):

Issue #403, opened 2020-07-09
Add support for early stopping and saving best model?


sandylaker:
Hi,

currently the checkpoints are save at the end of every epoch, and only checkpoints of the nearest k epochs are kept.
Could you please add hooks for early stopping, and saving the top k model based on the validation metrics? This would be very helpful.

Thanks

hellock:
Good suggestion.

SilvioGiancola:
Hi! Is there any plan on implementing such feature? Thank you! :)

zhouzaida:
> Hi! Is there any plan on implementing such feature? Thank you! :)

hi @SilvioGiancola , we have supported saving the best model. Please refer to https://github.com/open-mmlab/mmcv/blob/7540cf73ac7e5d1e14d0ffbd9b6759e83929ecfc/mmcv/runner/hooks/evaluation.py#L33.

zhouzaida:
We are interested in supporting early stopping. Are you interested in implementing it?

SilvioGiancola:
Thank you for your amazingly fast answer :) 
Do you also have a hook for early stopping? I would be interested in such a feature. If you have any idea/guidance on how to implement it, I don't mind giving a hand.

zhouzaida:
Thanks for your contribution.
We don't have this hook yet. But you may add a new hook to implement it. We can monitor a metric and we stop training when the metric stop improving. The metric can be val_loss or val_acc and so on.  Here is an implementation of early stopping. https://github.com/Bjarten/early-stopping-pytorch/blob/master/pytorchtools.py

zhouzaida:
> Thank you for your amazingly fast answer :)
> Do you also have a hook for early stopping? I would be interested in such a feature. If you have any idea/guidance on how to implement it, I don't mind giving a hand.

hi @SilvioGiancola , is there any progress?

24hours:
Hi, Im interested on this feature and I can implement it. 

I propose we implement `should_stop` flag in runner toggle by EarlyStoppingHook following the implementation from 
https://github.com/PyTorchLightning/pytorch-lightning/blob/3fcfd0214cc95fa7de49f6404421b5c6cd113c57/pytorch_lightning/callbacks/early_stopping.py#L204

I will make a pull request if its ok.

zhouzaida:
> Hi, Im interested on this feature and I can implement it.
> 
[...]

PR #1602, opened 2021-12-21
Add EarlyStopHook

24hours:
## Motivation
Allow user to specific stopping condition depending on loss value progress. 
following discussion in this issue #403 

## Modification
Added EarlyStopHook and `should_stop` flag in runner to trigger stop.

## BC-breaking (Optional)
No obvious BC-breaking 

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -103,6 +103,7 @@ class BaseRunner(metaclass=ABCMeta):
         self.optimizer = optimizer
         self.logger = logger
         self.meta = meta
+        self.should_stop = False
         # create work_dir
         if mmcv.is_str(work_dir):
             self.work_dir = osp.abspath(work_dir)
--- a/mmcv/runner/epoch_based_runner.py
+++ b/mmcv/runner/epoch_based_runner.py
@@ -107,7 +107,7 @@ class EpochBasedRunner(BaseRunner):
                          self._max_epochs)
         self.call_hook('before_run')
 
-        while self.epoch < self._max_epochs:
+        while self.epoch < self._max_epochs and not self.should_stop:
             for i, flow in enumerate(workflow):
                 mode, epochs = flow
                 if isinstance(mode, str):  # self.train()
--- a/mmcv/runner/hooks/early_stopping.py
+++ b/mmcv/runner/hooks/early_stopping.py
@@ -1,0 +1,222 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+# Adapted from
+# https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/callbacks/early_stopping.py
+
+import warnings
+from typing import Callable, Dict, Optional
+
+import numpy as np
+
+from ..hooks.hook import HOOKS, Hook
+
+
+@HOOKS.register_module()
+class EarlyStoppingHook(Hook):
+    r"""Stop training when a monitored metric has stopped improving.
+
+    Args:
+    monitor: quantity to be monitored.
+    phase: runner mode when executing model.Evaluation will only take place
+        with corresponding runner mode.
+    min_delta: minimum change in the monitored quantity to qualify as an
+        improvement, i.e. an absolute change of less than or equal to
+        `min_delta`, will count as no improvement.
+    patience: number of checks with no improvement
+        after which training will be stopped. Under the default configuration,
+        one check happens after every training epoch. However, the frequency
+        of validation can be modified by setting various parameters on the
+        ``Runner``, for example ``[(train, 10), (val, 1)]``.
+        .. note::
+            It must be noted that the patience parameter counts the number of
+            validation checks with no improvement, and not the number of
+            training epochs. Therefore, with parameters
+            ``[(train, 10), (val, 1)]`` and ``patience=3``, the runner
+            will perform at least 40 training epochs before being stopped.
+    verbose: verbosity mode.
+    mode: one of ``'min'``, ``'max'``. In ``'min'`` mode, training will stop
+        when the quantity monitored has stopped decreasing and in ``'max'``
+        mode it will stop when the quantity monitored has stopped increasing.
+    strict: whether to crash the training if `monitor` is not found in the
+        validation metrics.
+    check_finite: When set ``True``, stops training when the monitor becomes
+        NaN or infinite.
+    stopping_threshold: Stop training immediately once the monitored quantity
+        reaches this threshold.
+    divergence_threshold: Stop training as soon as the monitored quantity
+        becomes worse than this threshold.
+    """
+
+    mode_dict = {'min': np.less, 'max': np.greater}
+
+    def __init__(self,
+                 monitor: str,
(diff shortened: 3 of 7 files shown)


teamwong111 (approved):
LGTM

mzr1996:
The docstring is broken.
图片

mmcv/runner/hooks/early_stopping.py
```diff
@@ -0,0 +1,222 @@
...
+        when the quantity monitored has stopped decreasing and in ``'max'``
+        mode it will stop when the quantity monitored has stopped increasing.
+    strict: whether to crash the training if `monitor` is not found in the
+        validation metrics.
+    check_finite: When set ``True``, stops training when the monitor becomes
+        NaN or infinite.
+    stopping_threshold: Stop training immediately once the monitored quantity
+        reaches this threshold.
+    divergence_threshold: Stop training as soon as the monitored quantity
+        becomes worse than this threshold.
```
zhouzaida, line 46:
```suggestion
    Args:
        monitor: quantity to be monitored.
        phase: runner mode when executing model.Evaluation will only take place
            with corresponding runner mode.
        min_delta: minimum change in the monitored quantity to qualify as an
            improvement, i.e. an absolute change of less than or equal to
            `min_delta`, will count as no improvement.
        patience: number of checks with no improvement
            after which training will be stopped. Under the default configuration,
            one check happens after every training epoch. However, the frequency
            of validation can be modified by setting various parameters on the
            ``Runner``, for example ``[(train, 10), (val, 1)]``.
            .. note::
                It must be noted that the patience parameter counts the number of
                validation checks with no improvement, and not the number of
                training epochs. Therefore, with parameters
                ``[(train, 10), (val, 1)]`` and ``patience=3``, the runner
                will perform at least 40 training epochs before being stopped.
        verbose: verbosity mode.
        mode: one of ``'min'``, ``'max'``. In ``'min'`` mode, training will stop
            when the quantity monitored has stopped decreasing and in ``'max'``
            mode it will stop when the quantity monitored has stopped increasing.
        strict: whether to crash the training if `monitor` is not found in the
            validation metrics.
        check_finite: When set ``True``, stops training when the monitor becomes
            NaN or infinite.
        stopping_threshold: Stop training immediately once the monitored quantity
            reaches this threshold.
        divergence_threshold: Stop training as soon as the monitored quantity
            becomes worse than this threshold.
```

mmcv/runner/base_runner.py
```diff
@@ -103,6 +103,7 @@ def __init__(self,
         self.optimizer = optimizer
         self.logger = logger
         self.meta = meta
+        self.should_stop = False
```
zhouzaida, line 106: Suggest adding a comment to describe the function of the flag `self.should_stop`

mmcv/runner/epoch_based_runner.py
```diff
@@ -107,7 +107,7 @@ def run(self, data_loaders, workflow, max_epochs=None, **kwargs):
                          self._max_epochs)
         self.call_hook('before_run')
 
-        while self.epoch < self._max_epochs:
+        while self.epoch < self._max_epochs and not self.should_stop:
```
zhouzaida, line 110:
```suggestion
        while self.epoch < self._max_epochs:
             if self.should_stop:
                 self.logger.info('...')
                 break
```

zhouzaida:
Hi @24hours , thanks for your contribution.

There are two points to note:

- There are two ways to evaluate a dataset, one is to use [EvalHook](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/evaluation.py) and the other is to set the workflow like `workflow=[('train', 1), ('val', 1)]`. We need to consider both of them.
- The priority of hooks matters so we need to determine a reasonable priority for `EarlyStopHook`
https://github.com/open-mmlab/mmcv/blob/e4b5348ebf407f0886bc11caa7ca74d2fbf693ac/mmcv/runner/base_runner.py#L503-L531

mmcv/runner/epoch_based_runner.py
zhouzaida: It is helpful for users if we print a proper log here.

mmcv/runner/iter_based_runner.py
```diff
@@ -119,7 +119,7 @@ def run(self, data_loaders, workflow, max_iters=None, **kwargs):
 
         self.call_hook('before_epoch')
 
-        while self.iter < self._max_iters:
+        while self.iter < self._max_iters and not self.should_stop:
```
MeowZheng, line 122: kind reminder to provide log like `EpochBasedRunner`

mmcv/runner/hooks/early_stopping.py
MeowZheng, line 46: Might give some examples in docstring

mmcv/runner/epoch_based_runner.py
24hours: > Hi @24hours , thanks for your contribution.
> 
> There are two points to note:
> 
> * There are two ways to evaluate a dataset, one is to use [EvalHook](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/evaluation.py) and the other is to set the workflow like `workflow=[('train', 1), ('val', 1)]`. We need to consider both of them.
> * The priority of hooks matters so we need to determine a reasonable priority for `EarlyStopHook`
>   https://github.com/open-mmlab/mmcv/blob/e4b5348ebf407f0886bc11caa7ca74d2fbf693ac/mmcv/runner/base_runner.py#L503-L531



> 

Evalhook is registered with `priority=LOW` is most algorithm. 
https://github.com/open-mmlab/mmclassification/blob/master/mmcls/apis/train.py#L196
https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/train.py#L197
If we registery EarlyStopping to `VERY_LOW` then it would be possible to monitor EvalHook output. 

Since EarlyStoppingHook just monitor `runner.log_buffer.output[name]` it should work fine. 
https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/evaluation.py#L365
However, one would need to set `strict` flag to false. 

There are 2 issue need to clarify in this case. 
1. is `VERY_LOW` priority appropriate for EarlyStopping hook ?
    - I think it's ok because it doesnt seems to conflict with any other hook. 
    - However, it does feel weird that EShook is same rank with loggerhook. 
2. Should `strict` flag set to `false` by default ?
    - I opt for `strict = true` because strict pevent user giving errorneous log name. 
    - EShook should provide instruction to user upon validation check failed. Notify user to set `strict = false`

zhouzaida:
Sorry for the late reply. Have you ever used it in your own experiments?

24hours:
Hi, 

We have the requirement for internal training. However, it's not being rigorously tested in production environment yet. 
If you wish to, I could share our training log.

zhouzaida:
> 

Thanks, it would be better to provide the training log.

twmht:
@24hours 

Any update on this?

PR #1651, opened 2022-01-10
Support using variables in base config directly as normal variables.

mzr1996:
## Motivation

A config feature improvement draft. You can directly use the variables in the base config as normal variables virtually in the Python format config file.

## Modification

Parse the base config before loading the entire config file. And use `eval` instead of `import_module` to load configs.

## BC-breaking (Optional)

Should be no BC-breaking. It can pass the original unit tests. But please check it carefully.

## Use cases (Optional)

`base.py`:
```python
list_variable = [0, 1, 2]
tuple_variable = (0, 1)
dict_variable = dict(
    a=dict(first=1, second=2),
    b=dict(first=1, second=2)
)
```
`final.py`:
```python
_base_ = './base.py'
list_variable[1] = 5
x, y = tuple_variable
dict_variable = dict(
    a=dict(third=3, _delete_=True),
    b=dict(third=3)
)
```
```python
>>> from mmcv import Config
>>> cfg = Config.fromfile("final.py")
>>> print(cfg.pretty_text)
list_variable = [0, 5, 2]
tuple_variable = (0, 1)
dict_variable = dict(
    a=dict(third=3),
    b=dict(first=1, second=2, third=3)
)
x = 0
y = 1
```

## Checklist

**Before PR**:

**After PR**:


innerlee:
Need to add test to show case the new features

mmcv/utils/config.py
```diff
@@ -199,19 +222,24 @@ def _file2dict(filename, use_predefined_variables=True):
...
-                    for name, value in mod.__dict__.items()
-                    if not name.startswith('__')
-                }
-                # delete imported module
-                del sys.modules[temp_module_name]
+                cfg_dict = {}
+                with open(temp_config_file.name, 'r') as f:
+                    content = f.read()
+                codeobj = compile(content, '', mode='exec')
+                eval(codeobj, base_cfg_dict, cfg_dict)
```
innerlee, line 242: This is a clever trick.

One thing to note is that the eval will introduce many things into the global dict. See the following example. Whether side effects will be caused by these extra variables should be double checked.

image

mmcv/utils/config.py
```diff
@@ -199,19 +222,24 @@ def _file2dict(filename, use_predefined_variables=True):
             base_var_dict = Config._pre_substitute_base_vars(
```
innerlee, line 222: It seems that this code of substitution of base vars is no longer needed.

innerlee: Actually no. The base vars substitution  support yaml and json.

mmcv/utils/config.py
mzr1996: Right, these interval variables will be removed after merged in to the cfg_dict in line 272-276
```python
        cfg_dict = {
            k: v
            for k, v in cfg_dict.items() if not k.startswith('__')
        }
```

mzr1996:
> Need to add test to show case the new features

Done

zhouzaida:
```python
_base_ = './base.py'
list_variable[1] = 5
x, y = tuple_variable
dict_variable = dict(
    a=dict(third=3, _delete_=True),
    b=dict(third=3)
)
```

Does the usage of `list_variable[1] = 5` confuse the users? 

Would the follow design be more appropriate?
```python
_base_ = './base.py'
_base_.list_variable[1] = 5
x, y = tuple_variable
dict_variable = dict(
    a=dict(third=3, _delete_=True),
    b=dict(third=3)
)
```

mzr1996:
> 

Sounds good, I will try it.

mzr1996:
> 



> ```python
> _base_ = './base.py'
> list_variable[1] = 5
> x, y = tuple_variable
> dict_variable = dict(
>     a=dict(third=3, _delete_=True),
>     b=dict(third=3)
> )
> ```
> 
> Does the usage of `list_variable[1] = 5` confuse the users?
> 
> Would the follow design be more appropriate?
> 
> ```python
> _base_ = './base.py'
> _base_.list_variable[1] = 5
> x, y = tuple_variable
> dict_variable = dict(
>     a=dict(third=3, _delete_=True),
>     b=dict(third=3)
> )
> ```

Done, and the usage and the unit tests are updated.

zhouzaida:
We also need to add the example in the following documentation.
- https://github.com/open-mmlab/mmcv/blob/master/docs/en/understand_mmcv/config.md
- https://github.com/open-mmlab/mmcv/blob/master/docs/zh_cn/understand_mmcv/config.md

mmcv/utils/config.py
innerlee: Note the `abs/all/any/...`

innerlee:
In the doc, don't forget to mark it as python-config-only feature.

PR #1660, opened 2022-01-11
[Fix] fix a bug of CUDA illegal mem access when training s2anet

zhanggefan:
Fix a bug in the op `active_rotated_filter`. The buggy code incurs "CUDA illegal memory access" and `nan` loss after training s2anet for several iterations.

### DIFF
--- a/mmcv/ops/csrc/pytorch/cuda/active_rotated_filter_cuda.cu
+++ b/mmcv/ops/csrc/pytorch/cuda/active_rotated_filter_cuda.cu
@@ -14,7 +14,7 @@ void ActiveRotatedFilterForwardCUDAKernelLauncher(const Tensor input,
   int kW = input.size(4);
   int num_rotations = indices.size(3);
   int nEntry = num_orientations * kH * kW;
-  int output_size = output.numel();
+  int output_size = input.numel();
 
   at::cuda::CUDAGuard device_guard(input.device());
   cudaStream_t stream = at::cuda::getCurrentCUDAStream();


mmcv/ops/csrc/pytorch/cuda/active_rotated_filter_cuda.cu
```diff
@@ -14,7 +14,7 @@ void ActiveRotatedFilterForwardCUDAKernelLauncher(const Tensor input,
   int kW = input.size(4);
   int num_rotations = indices.size(3);
   int nEntry = num_orientations * kH * kW;
-  int output_size = output.numel();
+  int output_size = input.numel();
```
zhouzaida, line 17: Should we also fix the unit test of the op? https://github.com/open-mmlab/mmcv/blob/master/tests/test_ops/test_active_rotated_filter.py

zhanggefan: Hi, @zhouzaida 
It is really hard to design unit test that can efficiently detect out of bound array access. The fact is, even though the unit test is logically correct, the buggy code could possibly also pass the unit test as long as it does not access addresses beyond the process's memory resource border.
There are indeed inefficient solutions -- repeatedly malloc the input and ground truth and run the test several times until the ground truth is corrupted by out of bounds write, or make the unit test input large enough such that the test code can read out of the process's memory resource bound. But then the code would become unacceptably ugly. LOL
@zhouzaida @zytx121
Do you have suggestions?

zhouzaida: It is really hard.

zhouzaida, line 17: Why should the output_size equal `input.numel()` rather than `output.numel()`?

grimoire (approved):
LGTM

mmcv/ops/csrc/pytorch/cuda/active_rotated_filter_cuda.cu
zhanggefan: We are just following the original implementation, which uses variable name output_size as well. https://github.com/csuhan/s2anet/blob/d2870963815b10c8f7796577e30b7c3b14e47ac6/mmdet/ops/orn/src/cuda/ActiveRotatingFilter_cuda.cu#L94
Actually the numel of `output` tensor is 8 times the numel of `input`. Because the `output` feature tensor is orientation-sensitive and it encodes the features for all 8 discretized orientations.

zhouzaida: Got it.

zhouzaida (approved):

PR #1639, opened 2022-01-01
[Fix] Fix the cast_tensor_type

zhouzaida:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Fix the `cast_tensor_type` when the type of inputs is not the same as src_type.

## Modification

- Fix the cast_tensor_type in the `mmcv/runner/fp16_utils.py`
- Add two unit tests in the `tests/test_runner/test_fp16.py`

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

**Before PR**:

- [x] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to create this PR.
- [x] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) are used to fix the potential lint issues.
- [x] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [x] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [x] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls.
- [x] CLA has been signed and all committers have signed the CLA in this PR.

PR #1676, opened 2022-01-18
[Fix] refactor fp16 utils code so that the integer tensor will not be auto-cast into fp16

zhanggefan:
The original `cast_tensor_type` function sometimes unexpectedly casts integer tensors into fp16 tensors. For almost every situation during fp16 mix precision training, auto-casting integer tensors into fp16 does not make any sense.

PyTorch Tensors' definition is somewhat obscure. According to the official doc here: 
[torch.Tensor is an alias for the default tensor type (torch.FloatTensor).](https://pytorch.org/docs/stable/tensors.html)
But the following statement always returns `True`
```isinstance(torch.LongTensor([1,2,3]), torch.Tensor)```

So I add another branch(`if inputs.is_floating_point()`) to protect integer tensors from being cast.

### DIFF
--- a/mmcv/runner/fp16_utils.py
+++ b/mmcv/runner/fp16_utils.py
@@ -35,7 +35,7 @@ def cast_tensor_type(inputs, src_type, dst_type):
     if isinstance(inputs, nn.Module):
         return inputs
     elif isinstance(inputs, torch.Tensor):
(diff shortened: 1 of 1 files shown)


zhanggefan:
duplicate to #1639. close.

PR #1681, opened 2022-01-20
impl visual model

teamwong111:
## Motivation

Add model visualization in TensorboardLoggerHook.

## Modification

`mmcv/runner/hooks/logger/tensorboard.py` and `tests/test_runner/test_hooks.py` for model visualization.
`mmcv/runner/hooks/logger/pavi.py` for fixing docstring by the way.

## BC-breaking (Optional)

- Downstream repos need to modify a few codes of `forward_test`, like [mmcls](https://github.com/open-mmlab/mmclassification/blob/master/mmcls/models/classifiers/base.py#L69). Because although `add_graph` can pass in `List[Tensor]`, `torch/jit/_trace.py` has a loop and the parameter into `Module.forward` is a `tuple(Tensor)`.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/hooks/logger/tensorboard.py
+++ b/mmcv/runner/hooks/logger/tensorboard.py
@@ -1,7 +1,13 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import os.path as osp
+from functools import partial
 
-from mmcv.utils import TORCH_VERSION, digit_version
+import torch
+from torch.utils.data import IterableDataset
+
+from ....parallel import DataContainer
+from ....parallel.utils import is_module_wrapper
+from ....utils import TORCH_VERSION, digit_version
 from ...dist_utils import master_only
 from ..hook import HOOKS
 from .base import LoggerHook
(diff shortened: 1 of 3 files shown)


mmcv/runner/hooks/logger/tensorboard.py
```diff
@@ -1,15 +1,20 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import os.path as osp
+from functools import partial
 
-from mmcv.utils import TORCH_VERSION, digit_version
+import torch
+
+from ....parallel import DataContainer
+from ....parallel.utils import is_module_wrapper
+from ....utils import TORCH_VERSION, digit_version
```
mzr1996, line 9: Relative package path with more than two `.` is weird. Why don't use absolte path?

tests/test_runner/test_hooks.py
```diff
@@ -21,12 +21,15 @@
 from torch.utils.data import DataLoader
 
 from mmcv.fileio.file_client import PetrelBackend
+# yapf: disable
```
mzr1996, line 24: Why disable yapf here?

mzr1996 (changes requested):

mmcv/runner/hooks/logger/tensorboard.py
teamwong111: I will fix it.

tests/test_runner/test_hooks.py
teamwong111: Because isort and yapf are conflict.

PR #1683, opened 2022-01-22
[Fix] DDP duplicate logs handling

sleep3r:
## Motivation

Essentially, PyTorch DDP attaches StreamHandler directly after first forward() and by default logger.root.handlers may be empty list

## Modification

Added manual append to root.handlers list of StreamHandler with level ERROR in case of empty list

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/utils/logging.py
+++ b/mmcv/utils/logging.py
@@ -48,6 +48,10 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
     for handler in logger.root.handlers:
         if type(handler) is logging.StreamHandler:
             handler.setLevel(logging.ERROR)
+    else:
+        stream_handler = logging.StreamHandler()
+        stream_handler.setLevel(logging.ERROR)
+        logger.root.handlers.append(stream_handler)
 
     stream_handler = logging.StreamHandler()
     handlers = [stream_handler]


zhouzaida:
Hi @sleep3r ,thanks for your contribution. What is your PyTorch version?

sleep3r:
@zhouzaida 
Tested with **PyTorch 1.10.1**

I tried your logging utils with my trainloop and figured out some strange logging behaviour, let me show you some example.

sleep3r:
Снимок экрана 2022-01-23 в 14 59 03

sleep3r:
Here you could see, that by default the logger has not any handlers and DDP attaches one after forward of batch with idx 1. Therefore, the following logs are duplicated

sleep3r:
I may be missing something in case of influence on downstream or other related projects, but it seems for me to be a good notice for someone with a possible similar problem

zhouzaida:
Thank you very much for the description. BTW, is it convenient for you to test that torch<1.10.1 has similar behavior

sleep3r:
@zhouzaida 
I've just tested. So, seems like this behaviour starts from torch==1.9.0. 

1.8.0 and 1.8.1 are fine

zhouzaida:
> @zhouzaida I've just tested. So, seems like this behaviour starts from torch==1.9.0.
> 
> 1.8.0 and 1.8.1 are fine

Got it.

HAOCHENYE:
LGTM. But it seems that the propagate attribute of Logger causes the duplicate log. Can set `logger.propagate` to False solve the problem of duplicate logs?  This way we don't need the change attribute of the root logger any more。

sleep3r:
@HAOCHENYE
Well, yes logger.propagate=False solves this issue

sleep3r:
But this fix doesn't inherit the stated one in utils/logging.py  ¯\_(ツ)_/¯

sleep3r:
I mean, is it legit to replace all those StreamHandler setting stuff just by logger.propagate = False?

HAOCHENYE:
The `init_process_group` function in Pytorch 1.8 adds a `StreamHandler` to the root logger, so if `get_logger` is called after `init_process_group`, the log level of the root logger will be `ERROR` and there will be no duplicate printing. Pytorch 1.10 and Pytorch1.7 do not attach a StreamHandler to the root logger during the init_process_group phase but during the forward phase of the DDP (`Logging.info` is called, which adds the default `StreamHandler`(`INFO`) to the root logger. `Logging.info `is also called in Pytorch1.8, while the log level of `StreamHandler` was set to `ERROR` by `get_logger`). The logging level of the `StreamHandler` at this point is `INFO`, so there will be a duplicate log caused by the propagate mechanism. Since the duplicate log is caused by the propagate property of the child logger, should we set the property to False? Here is the logger setting for Detectron2 https://github.com/facebookresearch/detectron2/blob/cbbc1ce26473cb2a5cc8f58e8ada9ae14cb41052/detectron2/ utils/logger.py

Translated with www.DeepL.com/Translator (free version)

HAOCHENYE:
LGTM. I‘ll do further testing in the downstream codebase.

HAOCHENYE (approved):

mmcv/utils/logging.py
```diff
@@ -38,16 +39,7 @@ def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
...
-    # Starting in 1.8.0, PyTorch DDP attaches a StreamHandler <stderr> (NOTSET)
-    # to the root logger. As logger.propagate is True by default, this root
-    # level handler causes logging messages from rank>0 processes to
-    # unexpectedly show up on the console, creating much unwanted clutter.
-    # To fix this issue, we set the root logger's StreamHandler, if any, to log
-    # at the ERROR level.
-    for handler in logger.root.handlers:
-        if type(handler) is logging.StreamHandler:
-            handler.setLevel(logging.ERROR)
+    logger.propagate = False
```
zhouzaida, line 42: We need to add a comment.

sleep3r: @zhouzaida 
Please, check is it will be ok comment before 42th string:

# Handling of duplicate console logs.
# PyTorch DDP attaches a StreamHandler <stderr> (NOTSET)
# to the root logger. As logger.propagate is True by default, this root
# level handler causes logging messages from rank>0 processes to
# unexpectedly show up on the console, creating much unwanted clutter.

zhouzaida: Append the sentence 'More details can be found at PR #1683.'

HAOCHENYE: Hi @sleep3r , is there any progress?

teamwong111:
Hi @sleep3r , is there any progress?

mmcv/utils/logging.py
sleep3r: @HAOCHENYE @zhouzaida 
Sorry guys for this delay. I was shocked a bit due to Russian invasion of Ukraine :(

I've updated the comment.

sleep3r: But now, tests aren't passing...

Снимок экрана 2022-05-11 в 23 53 36

zhouzaida:
Hi @sleep3r , tests failed. https://app.circleci.com/pipelines/github/open-mmlab/mmcv/582/workflows/fe9adf11-579c-45c9-8802-4a15f74bb9d9/jobs/1207?invite=true#step-110-124

sleep3r:
@zhouzaida 
Yep, I mentioned this failed test above. But I haven't any thoughts how to fix it, unfortunately. Essentially, my new commit contains only some comment changes

sleep3r:
I could try the former fix without .propagate

sleep3r:
Lucky, it works

HAOCHENYE:
@sleep3r thanks for your reply~.
If we set default StreamHandler for `logging.root` in `get_logger`, It seems there will be no message shown in the terminal by calling logging.warning('xxx'):
```python
import logging
sh = logging.StreamHandler()
sh.setLevel(logging.ERROR)
logging.root.addHandler(sh)
``` 
Is this behavior reasonable? Besides, what will happen if we set `propagate=False`, maybe there is some corner case I've not concerned.

sleep3r:
@HAOCHENYE 
I can't reproduce the behaviour with warning logging. Setting `propagate=False` leads to fallen tests, and I don't have much about this question too.

HAOCHENYE:
> @HAOCHENYE I can't reproduce the behaviour with warning logging. Setting `propagate=False` leads to fallen tests, and I don't have much about this question too.



> @HAOCHENYE I can't reproduce the behaviour with warning logging. Setting `propagate=False` leads to fallen tests, and I don't have much about this question too.

Hi~ sorry for my late reply. This is the original issue
https://github.com/open-mmlab/mmcv/issues/1691 
I think it could be better if we do not need to change the root logger

OpenMMLab-Assistant003:
Hi @sleep3r！We are grateful for your efforts in helping improve this open-source project during your personal time. 
Thank you again for your contribution❤ @sleep3r

PR #1685, opened 2022-01-24
ConvModule has no respect to `init_cfg`

imyhxy:
## Motivation

`ConvModule` has no respect to `init_cfg` and always initialized with it's own `init_weights` function.

open-mmlab/mmcv#772 implemented the `BaseModule` class which will recursively call `init_weights` of sub-module. When a `ConvModule` is an attribution of one `BaseMocule`, it will initilize the `Conv2d` layer with `kaiming_init` regardless of the `init_cfg`. This requires the downstream application should carefully separate the `ConvModule` with the `nn.Module`, `nn.ModuleList` or `nn.Sequential`, which will break the `init_weights` call chain. Or, overwirte the `init_weights` function of user module.

I checked the mmclassification repo and it avoids the impact with the above trick, but there are several models in mmdetection that are already affected by this bug, although the initialization in `init_weights` of `ConvModule` is robust enough to train a good result.

## Modification

The `init_weights` of `ConvModule` is intended to be called once in the `__init__` function, so I think it is better to rename it to other name rather than `init_weights`.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/cnn/bricks/conv_module.py
+++ b/mmcv/cnn/bricks/conv_module.py
@@ -163,7 +163,7 @@ class ConvModule(nn.Module):
             self.activate = build_activation_layer(act_cfg_)
 
         # Use msra init by default
-        self.init_weights()
+        self._init_default_weights()
 
     @property
     def norm(self):
@@ -172,7 +172,7 @@ class ConvModule(nn.Module):
         else:
             return None
 
-    def init_weights(self):
+    def _init_default_weights(self):
         # 1. It is mainly for customized conv layers with their own
         #    initialization manners by calling their own ``init_weights()``,
         #    and we do not want ConvModule to override the initialization.


imyhxy:
@MeowZheng Take a look at the following examples, example1 fails to initialize the `conv` layer with user's `init_cfg`, while example2 do success. This is because when `ConvModule` is a direct member of a `BaseModule`-derived class, its `init_weights()` method will always override the user's `init_cfg`. In my opinion, the `init_weights()` of `ConvModule` may be conflicting with the `BaseModule`.

Example1: 
```python
class Stem1(BaseModule):
    def __init__(self):
        super().__init__()
        self.m = ConvModule(1, 2, 1, 1)

class Model1(BaseModule):
    def __init__(self):
        super().__init__(dict(type='Constant', val=1, layer=['Conv2d']))
        self.stem = Stem1()
        self.conv = ConvModule(1, 3, 1, 1)
model1 = Model1()
model1.init_weights()
print(model1.stem.m.conv.weight, model1.conv.conv.weight)
```

```bash
(Parameter containing:
 tensor([[[[-0.2800]]],
 
 
         [[[-0.2912]]]], requires_grad=True),
 Parameter containing:
 tensor([[[[-0.8592]]],
 
 
         [[[-2.0264]]],
 
 
         [[[ 0.6006]]]], requires_grad=True))
```

Example2:
```python
class Stem2(nn.Module):
    def __init__(self):
        super().__init__()
        self.m = ConvModule(1, 2, 1, 1)
        
class Model2(BaseModule):
    def __init__(self):
        super().__init__(dict(type='Constant', val=1, layer=['Conv2d']))
        self.stem = Stem2()
        self.m = nn.Sequential(ConvModule(1, 3, 1, 1))
        
model2 = Model2()
model2.init_weights()
print(model2.stem.m.conv.weight, model2.m[0].conv.weight)
```

```bash
(Parameter containing:
 tensor([[[[1.]]],
 
 
         [[[1.]]]], requires_grad=True),
 Parameter containing:
 tensor([[[[1.]]],
 
 
         [[[1.]]],
 
 
         [[[1.]]]], requires_grad=True))
```

MeowZheng:
> @MeowZheng Take a look at the following examples, example1 fails to initialize the `conv` layer with user's `init_cfg`, while example2 do success. This is because when `ConvModule` is a direct member of a `BaseModule`-derived class, its `init_weights()` method will always override the user's `init_cfg`. In my opinion, the `init_weights()` of `ConvModule` may be conflicting with the `BaseModule`.
> 
> Example1:
> 
> ```python
> class Stem1(BaseModule):
>     def __init__(self):
>         super().__init__()
>         self.m = ConvModule(1, 2, 1, 1)
> 
> class Model1(BaseModule):
>     def __init__(self):
>         super().__init__(dict(type='Constant', val=1, layer=['Conv2d']))
>         self.stem = Stem1()
>         self.conv = ConvModule(1, 3, 1, 1)
> model1 = Model1()
> model1.init_weights()
> print(model1.stem.m.conv.weight, model1.conv.conv.weight)
> ```
> 
> ```shell
> (Parameter containing:
>  tensor([[[[-0.2800]]],
>  
>  
>          [[[-0.2912]]]], requires_grad=True),
>  Parameter containing:
>  tensor([[[[-0.8592]]],
>  
>  
>          [[[-2.0264]]],
>  
>  
>          [[[ 0.6006]]]], requires_grad=True))
> ```
> 
> Example2:
> 
> ```python
> class Stem2(nn.Module):
>     def __init__(self):
>         super().__init__()
>         self.m = ConvModule(1, 2, 1, 1)
>         
> class Model2(BaseModule):
>     def __init__(self):
>         super().__init__(dict(type='Constant', val=1, layer=['Conv2d']))
>         self.stem = Stem2()
>         self.m = nn.Sequential(ConvModule(1, 3, 1, 1))
>         
> model2 = Model2()
> model2.init_weights()
> print(model2.stem.m.conv.weight, model2.m[0].conv.weight)
> ```
> 
> ```shell
> (Parameter containing:
>  tensor([[[[1.]]],
>  
>  
>          [[[1.]]]], requires_grad=True),
>  Parameter containing:
>  tensor([[[[1.]]],
>  
>  
>          [[[1.]]],
>  
>  
>          [[[1.]]]], requires_grad=True))
> ```

I see, but I have a little concern about this modification of is PR. Users may only use `ConvModule` but not use `BaseModule`, and this modification will cause a break when `ConvModule` users call `init_weights`

We plan to refractory weights initialization, and collaborations are appreciated If you are interested in it. We may design a new init api together offline and my email is meowzheng@outlook.com.

Issue #1691, opened 2022-01-26
mmcv.utils.get_logger should not change logging.root.


ypwhs:
# Environment

```
[...]

PR #1692, opened 2022-01-26
Fix mmcv.utils.get_logger changes logging.root in any condition.

ypwhs:
See: https://github.com/open-mmlab/mmcv/issues/1691


HAOCHENYE:
Being discussed in #1683.

zhouzaida:
Being discussed in https://github.com/open-mmlab/mmcv/pull/1683.

PR #1702, opened 2022-02-06
[Fix] Fix a bug in min_area_polygons

zytx121:
Fix a bug in the op ` min_area_polygons`. The buggy code incurs error output. The unit test cannot cover the rotated boxes, so we modified the unit test.


grimoire (approved):
LGTM

PR #1746, opened 2022-02-25
[Fix] Fix bugs in Voxelization op

wHao-Wu:

## Motivation

The unittest of mmdet3d fails after importing the Voxeliaztion op from mmcv. I found that there exists an inconsistency of Voxelization op between mmdet3d and mmcv, which causes the failure of mmdet3d.

## Modification

Replace `memcpy` with loop in the Voxelization op.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
+++ b/mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
@@ -101,7 +101,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
   CUDA_1D_KERNEL_LOOP(index, num_points) {
     auto coor_offset = coor + index * NDim;
     // skip invalid points
-    if ((index >= num_points) || (coor_offset[0] == -1)) return;
+    if (coor_offset[0] == -1) return;
 
     int num = 0;
     int coor_x = coor_offset[0];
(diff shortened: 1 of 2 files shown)


mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
```diff
@@ -101,7 +101,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
   CUDA_1D_KERNEL_LOOP(index, num_points) {
     auto coor_offset = coor + index * NDim;
     // skip invalid points
-    if (coor_offset[0] == -1) return;
+    if ((index >= num_points) || (coor_offset[0] == -1)) return;
```
grimoire, line 104: Not necessary since `CUDA_1D_KERNEL_LOOP` will keep index in range(0, num_points)

mmcv/ops/csrc/pytorch/cpu/voxelization.cpp
```diff
@@ -26,13 +26,16 @@ void dynamic_voxelize_forward_cpu_kernel(
       coor[ndim_minus_1 - j] = c;
     }
 
-    if (failed)
-      memset(&coors[i][0], -1, NDim * sizeof(T_int));
-    else
-      memcpy(&coors[i][0], &coor[0], NDim * sizeof(T_int));
+    for (int k = 0; k < NDim; ++k) {
+      if (failed)
```
grimoire, line 30: Could move the if/else out of for loop.

grimoire (approved):

mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
zhanggefan: shall we use `continue` here to avoid early return? if the number of input points is so large that multiple points shall be processed by one thread? (though practically impossible)

grimoire: Yes, continue is required.

PR #1748, opened 2022-02-25
[Fix] Fix bugs in DynamicScatter op

wHao-Wu:

## Motivation

The unittest of mmdet3d fails after importing the DynamicScatter op from mmcv. I found that there exists an inconsistency of DynamicScatter op between mmdet3d and mmcv, which causes the failure of mmdet3d.

## Modification

Recover the `if` statement.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
+++ b/mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
@@ -26,10 +26,15 @@ std::vector<at::Tensor> DynamicPointToVoxelForwardCUDAKernelLauncher(
   std::tie(out_coors, coors_map, reduce_count) =
       at::unique_dim(coors_clean, 0, true, true, true);
 
-  // the first element of out_coors is always (-1,-1,-1) and should be removed
-  out_coors = out_coors.slice(0, 1);
-  reduce_count = reduce_count.slice(0, 1).to(torch::kInt32);
-  coors_map = coors_map.to(torch::kInt32) - 1;
+  if (out_coors.index({0, 0}).lt(0).item<bool>()) {
+    // the first element of out_coors (-1,-1,-1) and should be removed
+    out_coors = out_coors.slice(0, 1);
+    reduce_count = reduce_count.slice(0, 1);
+    coors_map = coors_map - 1;
+  }
+
+  coors_map = coors_map.to(torch::kInt32);
+  reduce_count = reduce_count.to(torch::kInt32);
 
   auto reduced_feats =
       at::empty({out_coors.size(0), num_feats}, feats.options());
--- a/mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
+++ b/mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
@@ -26,6 +26,7 @@ std::vector<at::Tensor> DynamicPointToVoxelForwardCUDAKernelLauncher(
   std::tie(out_coors, coors_map, reduce_count) =
       at::unique_dim(coors_clean, 0, true, true, true);
 
+  // deal with the condition when the input may contain out-of-range coors (full with -1).
   if (out_coors.index({0, 0}).lt(0).item<bool>()) {
     // the first element of out_coors (-1,-1,-1) and should be removed
     out_coors = out_coors.slice(0, 1);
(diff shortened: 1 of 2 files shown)


ZwwWayne:
Invite @zhanggefan to take a look at this PR

mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
```diff
@@ -26,10 +26,15 @@ std::vector<at::Tensor> DynamicPointToVoxelForwardCUDAKernelLauncher(
   std::tie(out_coors, coors_map, reduce_count) =
       at::unique_dim(coors_clean, 0, true, true, true);
 
-  // the first element of out_coors is always (-1,-1,-1) and should be removed
-  out_coors = out_coors.slice(0, 1);
-  reduce_count = reduce_count.slice(0, 1).to(torch::kInt32);
-  coors_map = coors_map.to(torch::kInt32) - 1;
+  if (out_coors.index({0, 0}).lt(0).item<bool>()) {
```
ZwwWayne, line 29: is there any other cases that need this judgment?

zhanggefan: This judgment is necessary if the input may contain out-of-range `coors` (full with -1). If we could guarantee that the input coors are all valid, then we may safely remove this branch.

tests/test_ops/test_scatter_points.py
```diff
@@ -11,7 +11,7 @@ def test_dynamic_scatter():
     feats = torch.rand(
         size=(200000, 3), dtype=torch.float32, device='cuda') * 100 - 50
     coors = torch.randint(
-        low=-1, high=20, size=(200000, 3), dtype=torch.int32, device='cuda')
+        low=0, high=20, size=(200000, 3), dtype=torch.int32, device='cuda')
```
zhanggefan, line 15: why should we change this? I personally think the original test that simulates input with invalid coors (have value -1) is actually beneficial. We actually could not guarantee that the output coors of dynamic voxelization are all valid according to the logics here:
https://github.com/open-mmlab/mmcv/blob/a0127e1712cd58b22736d20ed4e5531e1d277f5e/mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh#L22-L49

mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
zhanggefan: I just check the code of dynamic voxelization and found that we could not rule out the possibility that the input contains out-of-range coors.

ZwwWayne: sure, then we can suggest @wHao-Wu to leave a comment here as a reminder.

tests/test_ops/test_scatter_points.py
wHao-Wu: Sorry for the mistake. I found that some of the unittest in mmdet3d has been deleted here. The op cannot pass the unittest after the recovery of the unittest before this PR. Then I need to recover the unittest to implement the strict testing for the DynamicScatter op.

mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
wHao-Wu: OK. I will leave a comment here. Thanks for your help.

tests/test_ops/test_scatter_points.py
zhanggefan: LGTM

zhanggefan (approved):

mmcv/ops/csrc/pytorch/cuda/scatter_points_cuda.cu
```diff
@@ -26,10 +26,16 @@ std::vector<at::Tensor> DynamicPointToVoxelForwardCUDAKernelLauncher(
   std::tie(out_coors, coors_map, reduce_count) =
       at::unique_dim(coors_clean, 0, true, true, true);
 
-  // the first element of out_coors is always (-1,-1,-1) and should be removed
-  out_coors = out_coors.slice(0, 1);
-  reduce_count = reduce_count.slice(0, 1).to(torch::kInt32);
-  coors_map = coors_map.to(torch::kInt32) - 1;
+  // deal with the condition when the input may contain out-of-range coors (full with -1).
+  if (out_coors.index({0, 0}).lt(0).item<bool>()) {
+    // the first element of out_coors (-1,-1,-1) and should be removed
```
ZwwWayne, line 31: rephrase the comment

ZwwWayne (approved):

grimoire (approved):

PR #1750, opened 2022-02-27
deprecate model deployment

grimoire:
## Motivation

This PR add deprecate warning to codes about model deployment.

## Modification

Add warning in codes and docs.

## BC-breaking (Optional)

Codes include c++ and cuda ops implement will be deprecated in the future.


teamwong111:
What time does the "future" mean? I think we should discuss the specific time to remove the support for deployment.

RunningLeon (approved):
LGTM

ZwwWayne (approved):

PR #1757, opened 2022-03-01
[Fix] Pad boader fix

Harold-lkk:
## Motivation
When using `Pad` to pad image, we support the type of pad_val is int or dict. If the type of pad_val is int, `Pad` should pad all the channels of the image with pad_val. The pad_val should have the same dim with an image.

### DIFF
--- a/tests/test_transforms/test_transforms_processing.py
+++ b/tests/test_transforms/test_transforms_processing.py
@@ -172,7 +172,7 @@ class TestPad:
         assert (results['img'] == np.ones((1333, 1333, 3))).all()
 
         # test pad_val is dict
-        # test rgb image
+        # test rgb image, size=(2000, 2000)
         trans = Pad(
             size=(2000, 2000),
             pad_val=dict(img=(12, 12, 12), seg=(10, 10, 10)))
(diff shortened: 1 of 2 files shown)


tests/test_transforms/test_transforms_processing.py
```diff
@@ -172,11 +172,16 @@ def test_pad(self):
...
+        results = trans(copy.deepcopy(data_info))
+        assert (results['img'][1333:2000, 800:2000, :] == 12).all()
+        # test gray image
+        new_img = np.random.random((1333, 800))
         data_info['img'] = new_img
-        trans = Pad(pad_to_square=True, pad_val=0)
+        trans = Pad(size=(2000, 2000), pad_val=12)
         results = trans(copy.deepcopy(data_info))
-        assert (results['img'] == np.zeros((1333, 1333, 3))).all()
+        assert (results['img'][1333:2000, 800:2000] == 12).all()
```
plyfager, line 184: might also check the value of `gt_semantic_seg`.

tests/test_transforms/test_transforms_processing.py
```diff
@@ -172,11 +172,16 @@ def test_pad(self):
...
         # test pad_val
-        new_img = np.zeros((1333, 800, 3))
+        # test rgb image
+        trans = Pad(size=(2000, 2000), pad_val=12)
+        results = trans(copy.deepcopy(data_info))
+        assert (results['img'][1333:2000, 800:2000, :] == 12).all()
+        # test gray image
+        new_img = np.random.random((1333, 800))
         data_info['img'] = new_img
-        trans = Pad(pad_to_square=True, pad_val=0)
```
ZwwWayne, line 177: should we also test the case of pad to square？

ZwwWayne (approved):

PR #1772, opened 2022-03-06
Implement corner pool with python for torch<1.5

grimoire:
## Motivation

Corner pool did not have it's own cuda kernel. This PR replace the cpp implementation with a python implementation.

## Modification

Remove code of corner pool in ops/csrc (both pytorch and parrots).
Add python implementation in `corner_pool.py`.

Has not tested on parrots.


grimoire:
Note that since MMCV requires pytorch>=1.5.0. What if parrots provide support of  `torch.cummax`, we can remove all these codes and make it much cleaner.

grimoire:
@zhouzaida benchmark might not be necessary since the python implementation is exactly the same as the c++ one(this ops has no cuda kernel). And the implementation is only triggered when torch<1.5, which is not supported now.

zhouzaida:
> @zhouzaida benchmark might not be necessary since the python implementation is exactly the same as the c++ one(this ops has no cuda kernel). And the implementation is only triggered when torch<1.5, which is not supported now.

got it

luopeichao (approved):
It is passed in parrots. We will add torch.cummax in parrots in the future.

ZwwWayne (approved):

PR #1776, opened 2022-03-07
[WIP] fix checkpoint hook logic

BIGWangYuDong:
## Motivation

If use EpochBasedRunner and save ckpt as IterBase (such as `checkpoint_config = dict(by_epoch=False, interval=100)`), this may cause a potential error: the ckpt will name as `epoch_*.pth`.

## Modification

change the logic in checkpoint_hook.

 Maybe we can only set the code in BaseRunner?

### DIFF
--- a/mmcv/runner/base_runner.py
+++ b/mmcv/runner/base_runner.py
@@ -2,6 +2,8 @@
 import copy
 import logging
 import os.path as osp
+import platform
+import shutil
 import warnings
 from abc import ABCMeta, abstractmethod
 
@@ -10,7 +12,7 @@ from torch.optim import Optimizer
 
 import mmcv
 from ..parallel import is_module_wrapper
-from .checkpoint import load_checkpoint
+from .checkpoint import load_checkpoint, save_checkpoint
 from .dist_utils import get_dist_info
 from .hooks import HOOKS, Hook
 from .log_buffer import LogBuffer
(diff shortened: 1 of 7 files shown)


mmcv/runner/epoch_based_runner.py
```diff
@@ -134,7 +134,8 @@ def save_checkpoint(self,
                         filename_tmpl='epoch_{}.pth',
                         save_optimizer=True,
                         meta=None,
-                        create_symlink=True):
+                        create_symlink=True,
+                        by_epoch=True):
```
AllentDan, line 138: Lack corresponding docstring here.

mmcv/runner/iter_based_runner.py
```diff
@@ -181,7 +181,8 @@ def save_checkpoint(self,
                         filename_tmpl='iter_{}.pth',
                         meta=None,
                         save_optimizer=True,
-                        create_symlink=True):
+                        create_symlink=True,
+                        by_epoch=False):
```
AllentDan, line 185: Lack docstring

AllentDan (commented):
May add unit tests

BIGWangYuDong:
- Move the code to `BaseRunner` and delete the `save_checkpoint` function in `EpochBasedRunner` and `IterBasedRunner`
- update related test unit
- add a warning if the checkpoint filename template does not match the saving checkpoints way

mmcv/runner/base_runner.py
```diff
@@ -195,14 +197,70 @@ def val(self):
     def run(self, data_loaders, workflow, **kwargs):
         pass
 
-    @abstractmethod
     def save_checkpoint(self,
                         out_dir,
-                        filename_tmpl,
+                        filename_tmpl='epoch_{}.pth',
```
AllentDan, line 202: Using `epoch_{}.pth` as the default checkpoint name for `iter_based_runner` may be a little confusing for users. A possible method is using None as default, but getting changed to `iter_{}.pth` or `epoch_{}.pth` according to `by_epoch` if it is None. But I am not sure if it is the best solution here.

PR #1792, opened 2022-03-10
Make sure padding size is not negative

Charlyo:
If Resize maintains aspect ratio, could be the case that image shape is bigger than resize value.


zhouzaida:
Hi, @Charlyo , thanks for your contribution. Maybe unit tests for this case should also be added. https://github.com/open-mmlab/mmcv/blob/master/tests/test_image/test_geometric.py

Charlyo:
Updated!

zhouzaida:
Hi @Charlyo , the lint failed. You could follow [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to fix the lint error by installing `pre-commit` tool.

zhouzaida (approved):

PR #1802, opened 2022-03-15
[fix] config ignore imported modules and functions

SuTanTank:
## Motivation

When use `import` like `import os.path as osp` in config. The parsed config's `_cfg_dict` contains an entry which looks like `osp: <module ....`. 

Then when calling `pretty_text()`, this dict is converted back to string, which is different from the original config., Then this string is formatted by `yapf` and error occurs. 

```python 
# config: pretty_text()
        cfg_dict = self._cfg_dict.to_dict()
        text = _format_dict(cfg_dict, outest_level=True)
        # copied from setup.cfg
        yapf_style = dict(
            based_on_style='pep8',
            blank_line_before_nested_class_or_def=True,
            split_before_expression_after_opening_paren=True)
        text, _ = FormatCode(text, style_config=yapf_style, verify=True)
```

## Modification

When parsing a config file to dict, modules and functions are ignored. 

## BC-breaking (Optional)

No

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -7,6 +7,7 @@ import platform
 import shutil
 import sys
 import tempfile
+import types
 import uuid
 import warnings
 from argparse import Action, ArgumentParser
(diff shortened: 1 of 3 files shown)


HAOCHENYE (approved):
LGTM.

ZwwWayne:
Hi @SuTanTank ,
Thanks for the kind PR. Would you like to also update the unit tests to test the updated code?

SuTanTank:
> 

What function should I test? There is no new API introduced.

PR #1813, opened 2022-03-17
Add annotation for ninja build in setup.py

imabackstabber:
## Motivation

Add annotation, which explains why code (the part for adjusting number of cpu used in ninja build) will inevitably goes into Except block due to inherent design on MacOS,in setup.py.

## Modification

Add annotation.

## BC-breaking (Optional)

No

## Use cases (Optional)

No

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -241,6 +241,11 @@ def get_extensions():
             num_cpu = len(psutil.Process().cpu_affinity())
             cpu_use = max(4, num_cpu - 1)
         except (ModuleNotFoundError, AttributeError):
+            # for macos, due to psutil.Process() has no method 'cpu_affinity',
+            # due to its own design which aims to
+            # let kernel manage cache affinity well.
+            # Because of that, AttributeError will be invoked,
+            # and here set cpu_use = 4 by default.
             cpu_use = 4
 
         os.environ.setdefault('MAX_JOBS', str(cpu_use))


setup.py
```diff
@@ -241,6 +241,11 @@ def get_extensions():
             num_cpu = len(psutil.Process().cpu_affinity())
             cpu_use = max(4, num_cpu - 1)
         except (ModuleNotFoundError, AttributeError):
+            # for macos, due to psutil.Process() has no method 'cpu_affinity',
```
ZwwWayne, line 244: rephrase as the following:

```python
            # In macos, psutil.Process() does not have method 'cpu_affinity'.
            # In this case, AttributeError will be invoked 
            # and we set cpu_use = 4 by default.

```

setup.py
```diff
@@ -241,6 +241,9 @@ def get_extensions():
             num_cpu = len(psutil.Process().cpu_affinity())
             cpu_use = max(4, num_cpu - 1)
         except (ModuleNotFoundError, AttributeError):
+            # In macos, psutil.Process() does not have method 'cpu_affinity'.
+            # In this case, AttributeError will be invoked 
+            # and we set cpu_use = 4 by default.
```
hellock, line 246: `multiprocessing.cpu_count()` is applicable on macOS.

hellock: should also work on other OS (not tested).

imabackstabber: > `multiprocessing.cpu_count()` is applicable on macOS.

not sure process.cpu_affinity() will return same result as cpu_count() do, or is it ok we just use maximum num of cpu for ninja build?If so we can change to multiprocessing.cpu_count() which works on linux,mac and windows.

imabackstabber: just tested several cases on ubuntu and windows,``multiprocessing.cpu_count()`` and  ``psutil.Process().cpu_affinity()`` returned same value.Change to ``multiprocessing.cpu_count()`` then.

imabackstabber: Also,notice that linux make ``taskset`` command exposed to user,which means user can manually set number of cpu used for specific process,so I think maybe we should leave ``cpu_affinity()`` intact.
Any comment? @hellock

ZwwWayne: > > `multiprocessing.cpu_count()` is applicable on macOS.
> 
> not sure process.cpu_affinity() will return same result as cpu_count() do, or is it ok we just use maximum num of cpu for ninja build?If so we can change to multiprocessing.cpu_count() which works on linux,mac and windows.

we can check if they are the same. If it works, we can uniformly use `multiprocessing.cpu_count()`. We should not use the maximum num of CPU for ninja build to avoid occupying all CPU resources which slows the machine.

imabackstabber: > > > `multiprocessing.cpu_count()` is applicable on macOS.
> > 
> > 
> > not sure process.cpu_affinity() will return same result as cpu_count() do, or is it ok we just use maximum num of cpu for ninja build?If so we can change to multiprocessing.cpu_count() which works on linux,mac and windows.
> 
> we can check if they are the same. If it works, we can uniformly use `multiprocessing.cpu_count()`. We should not use the maximum num of CPU for ninja build to avoid occupying all CPU resources which slows the machine.

Usually they will be the same,but it's no longer same if user manually specify cpu_affinity by using [taskset](https://man7.org/linux/man-pages/man1/taskset.1.html) command on linux,or user can set cpu affinity by using [task manager](https://winbuzzer.com/2020/07/03/how-to-set-cpu-affinity-in-windows-10-with-task-manager-or-system-properties-xcxwbt/) on windows.
So for linux and windows,using `Process().cpu_affinity()` is fine.For macos,probably using `num_cpu - 1` will be fine.(just like original code,and yes, I don't think we should use maximum num of cpu)

PR #1822, opened 2022-03-21
[Bug] Fix voxelization bug

VVsssssk:
Hello, when I deploy the `CenterPoint` model in mmdeploy, I find a bug in Voxelization op. In point_to_voxelidx_kernel func, if find invalid points or point num out of the boundary, func  `return` will make the following point not processed, and then affect the model results.

Before:
企业微信截图_1647831647752

After fix:
企业微信截图_1647831547777

### DIFF
--- a/mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
+++ b/mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
@@ -101,7 +101,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
   CUDA_1D_KERNEL_LOOP(index, num_points) {
     auto coor_offset = coor + index * NDim;
     // skip invalid points
-    if (coor_offset[0] == -1) return;
+    if (coor_offset[0] == -1) continue;
 
     int num = 0;
     int coor_x = coor_offset[0];
@@ -122,7 +122,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
           point_to_pointidx[index] = i;
         } else if (num >= max_points) {
           // out of boundary
-          return;
+          continue;
         }
       }
     }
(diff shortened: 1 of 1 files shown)


mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
```diff
@@ -122,7 +122,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
           point_to_pointidx[index] = i;
         } else if (num >= max_points) {
           // out of boundary
-          return;
+          break;
```
grimoire, line 125: break is not enough, you should mark it with a flag and continue in the outer loop.

VVsssssk: Right, but if num >= maxpoints and break loop, and then check `if (num == 0)` and `if (num < max_points)` is invalid, so I think break is enough.

grimoire: Sure, I see.

grimoire (approved):
LGTM

mmcv/ops/csrc/common/cuda/voxelization_cuda_kernel.cuh
```diff
@@ -101,7 +101,7 @@ __global__ void point_to_voxelidx_kernel(const T_int* coor,
   CUDA_1D_KERNEL_LOOP(index, num_points) {
     auto coor_offset = coor + index * NDim;
     // skip invalid points
-    if (coor_offset[0] == -1) return;
+    if (coor_offset[0] == -1) continue;
```
ZwwWayne, line 104: shall we leave comments here to remind future developers?

zhanggefan (approved):
LGTM as well

PR #1824, opened 2022-03-22
NMS with CUDA only

grimoire:
This PR add a cuda kernel for nms to avoid computation on cpu.

I am not sure if I should call this an "optimization" cause on small input data, cpu performance might even better than gpu. And the data distribution will also affect the performance.

envs:
- Device: RTX 2070super
- CUDA: 11.3
- nvidia driver: 465.19.01
- CPU: Intel(R) Core(TM) i7-9700KF CPU @ 3.60GHz

Test data comes from [faster rcnn](https://github.com/open-mmlab/mmdetection/blob/master/configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py) and [demo image](https://github.com/open-mmlab/mmdetection/blob/master/demo/demo.jpg). Both nms in rpn and nms in bbox head are tested.

| - | old | new |
| --- | --- | --- |
| rpn data(4741 boxes) | 0.98435ms | **0.88798ms** |
| rcnn data(583 boxes) | 0.23121ms | **0.22827ms** |
| random data(500 boxes) | **0.23056ms** | 0.29762ms |
| random data(1000 boxes) | **0.26279ms** | 0.41685ms |
| random data(5000 boxes) | **1.27405ms** | 1.34129ms |
| random data(10000 boxes) | 3.62377ms | **2.74194ms** |
| random data(20000 boxes) | 21.07619ms | **4.80819ms** |

Real data might have clustered bboxes, which reduce gpu computations. I guess that is why real data performance is better than random data.

### DIFF
--- a/mmcv/ops/csrc/common/cuda/nms_cuda_kernel.cuh
+++ b/mmcv/ops/csrc/common/cuda/nms_cuda_kernel.cuh
@@ -72,4 +72,65 @@ __global__ void nms_cuda(const int n_boxes, const float iou_threshold,
     }
   }
 }
+
+__global__ void gather_keep_from_mask_parallize(
+    bool *keep, const unsigned long long *dev_mask, const int n_boxes) {
+  const int col_blocks = (n_boxes + threadsPerBlock - 1) / threadsPerBlock;
+  const int tid = threadIdx.x;
+
+  extern __shared__ unsigned long long remv[];
+  __shared__ const unsigned long long *p[1];
+  __shared__ bool finish[1];
+  __shared__ int nblock_ptr[1];
+
+  for (int i = tid; i < col_blocks; i += blockDim.x) {
+    remv[i] = 0;
+  }
+  if (tid == 0) {
+    finish[0] = false;
+  }
+  __syncthreads();
+
+  int nblock = 0;
+  int inblock = 0;
+  int i = 0;
+  bool do_sync = false;
+
+  while (!finish[0]) {
+    if (tid == 0) {
+      do_sync = false;
+      for (; nblock < col_blocks; ++nblock) {
+#pragma unroll
+        for (; inblock < threadsPerBlock; ++inblock) {
+          if (i < n_boxes && !(remv[nblock] & (1ULL << inblock))) {
+            keep[i] = true;
+            p[0] = dev_mask + i * col_blocks;
+            i += 1;
+            inblock = (inblock + 1) % threadsPerBlock;
+            nblock_ptr[0] = nblock;
+            nblock += inblock == 0 ? 1 : 0;
+            do_sync = true;
+            break;
+          } else {
+            i += 1;
+          }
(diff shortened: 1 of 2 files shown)


mmcv/ops/csrc/pytorch/cuda/nms_cuda.cu
```diff
@@ -24,31 +24,13 @@ Tensor NMSCUDAKernelLauncher(Tensor boxes, Tensor scores, float iou_threshold,
...
-      unsigned long long* p = mask_host + i * col_blocks;
-      for (int j = nblock; j < col_blocks; j++) {
-        remv[j] |= p[j];
-      }
-    }
-  }
-
+  at::Tensor keep_t = at::zeros(
+      {boxes_num}, boxes.options().dtype(at::kBool).device(at::kCUDA));
+  gather_keep_from_mask_parallize<<<1, min(col_blocks, THREADS_PER_BLOCK),
```
ZwwWayne, line 29: It might not be necessary to indicate this function is parallel through `parallize`, since it launches a CUDA kernel and it should always be parallel. Therefore, I prefer `gather_keep_from_mask` as the function name which is clear and short.

mmcv/ops/csrc/pytorch/cuda/nms_cuda.cu
```diff
@@ -24,31 +24,13 @@ Tensor NMSCUDAKernelLauncher(Tensor boxes, Tensor scores, float iou_threshold,
...
-      keep[i] = true;
-      // set every overlap box with bit 1 in remv
-      unsigned long long* p = mask_host + i * col_blocks;
-      for (int j = nblock; j < col_blocks; j++) {
-        remv[j] |= p[j];
-      }
-    }
-  }
-
+  at::Tensor keep_t = at::zeros(
```
ZwwWayne, line 27: It would be better if some explanation/comments could be added above this line since this differs our implementation from the old/more commonly used implementation in other repos.

mmcv/ops/csrc/common/cuda/nms_cuda_kernel.cuh
```diff
@@ -72,4 +72,40 @@ __global__ void nms_cuda(const int n_boxes, const float iou_threshold,
     }
   }
 }
+
+__global__ void gather_keep_from_mask_parallize(
+    bool *keep, const unsigned long long *dev_mask, const int n_boxes) {
+  const int col_blocks = (n_boxes + threadsPerBlock - 1) / threadsPerBlock;
+  const int tid = threadIdx.x;
+
+  extern __shared__ unsigned long long remv[];
```
ZwwWayne, line 81: remv is not a very clear shortname (I cannot get its meaning at the first glance). Better to rename it or add an explanation to tell readers whose abbreviation it is.

ZwwWayne (approved):

PR #1827, opened 2022-03-23
[Fix] config handles string with escape correctly

SuTanTank:
## Motivation

`runner.resume()` recompile configs loaded from `['meta']['config']`. In downstream repos, this meta config text is created by `config.pretty_text()`, which convert confg from dict to string. During this conversion, escapes like '\\', '\U' is not handled. 

## Modification

I change the way dict is converted to string. When key value is a string, convert `\` to `\\` etc. by `repr()`. 

## BC-breaking (Optional)

No

## Use cases (Optional)

Any config with string type value, including '\''s and other escaped characters. 

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tests/data/config/l.py
+++ b/tests/data/config/l.py
@@ -1,11 +1,4 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import os.path as osp
-
-
-def func(x):
-    return x
-
 _base_ = ['./l1.py', './l2.yaml', './l3.json', './l4.py']
 item3 = False
 item4 = 'test'
-item5 = osp.expanduser('~')
--- a/tests/data/config/l.py
+++ b/tests/data/config/l.py
@@ -8,3 +8,4 @@ def func(x):
 _base_ = ['./l1.py', './l2.yaml', './l3.json', './l4.py']
 item3 = False
 item4 = 'test'
+item5 = osp.expanduser('~')
--- a/tests/data/config/n.py
+++ b/tests/data/config/n.py
@@ -3,7 +3,7 @@ import os.path as osp
 
 
 def func(x):
-    return x
+    return '\\' + x
 
 test_item1 = [1, 2]
 bool_item2 = True
@@ -21,3 +21,5 @@ dict_item4 = dict(
     f=dict(a='69'))
 dict_item5 = {'x/x': {'a.0': 233}}
 dict_list_item6 = {'x/x': [{'a.0': 1., 'b.0': 2.}, {'c/3': 3.}]}
+dict_item7 = osp.expanduser('~')
+dict_item8 = func('a')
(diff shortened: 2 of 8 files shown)


tests/data/config/l.py
```diff
@@ -8,3 +8,4 @@ def func(x):
 _base_ = ['./l1.py', './l2.yaml', './l3.json', './l4.py']
 item3 = False
 item4 = 'test'
+item5 = osp.expanduser('~')
```
HAOCHENYE, line 11: Hi, Thanks for your contribution! `test_merge_from_multiple_bases` should also be modified.

HAOCHENYE, line 11: Hi, Thanks for your contribution! `test_merge_from_multiple_bases` should also be modified.

HAOCHENYE (approved):
LGTM. But it seems there is some problem in workflow `build_cu102`. @zhouzaida

tests/data/config/n.py
```diff
@@ -21,3 +21,5 @@ def func(x):
     f=dict(a='69'))
 dict_item5 = {'x/x': {'a.0': 233}}
 dict_list_item6 = {'x/x': [{'a.0': 1., 'b.0': 2.}, {'c/3': 3.}]}
+dict_item7 = osp.expanduser('~')
+dict_item8 = func('a')
```
HAOCHENYE, line 25: It could be better to add some unit tests to test could `Config` dumps the windows path successfully.

SuTanTank:
I moved the unrelated changes to other PRs and wrote a new unitest for this use case.

tests/data/config/w.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+# config now can have imported modules and defined functions for convenience
+import os.path as osp
+def func():
+    return 'string with \"escape\" characters'
+
+str_item_1 = osp.join(osp.expanduser('~'), 'folder') # with backslash in Windows
+str_item_2 = func()
```
HAOCHENYE, line 9:
```suggestion
str_item_2 = func()

```
Although lint will not check the test file, a new line at end of file may be properer.

tests/data/config/w.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+# config now can have imported modules and defined functions for convenience
+import os.path as osp
+def func():
+    return 'string with \"escape\" characters'
```
HAOCHENYE, line 6: It seems `'string with \"escape\" characters' == 'string with "escape" characters'` is True and str_items2 maybe not with an escape character?

SuTanTank: Yes, python string need no escape for quotes. I'll change to another example.

HAOCHENYE (approved):

SuTanTank:
Since this change has been applied to mmengine, close.

PR #1834, opened 2022-03-24
[Enhancement] Support None in DictAction

### LINKED ISSUE #1112: Support None in DictAction
Some times there are needs to override some key in the config to be `None`, e.g., the load_from is assigned to some string and you want to override it through `--cfg-options load_from=None`, it will pass a string of None rather than the real `None` value to the config and make checkpoint loader load checkpoint from a path named `'None'`

vakker:
## Motivation

Resolves #1112

## Modification

Adds `--cfg-options key=none` functionality.

## BC-breaking (Optional)

None that I'm aware of.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -638,6 +638,8 @@ class DictAction(Action):
             pass
         if val.lower() in ['true', 'false']:
             return True if val.lower() == 'true' else False
+        if val.lower() == 'none':
+            return None
         return val
 
     @staticmethod
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -638,7 +638,7 @@ class DictAction(Action):
             pass
         if val.lower() in ['true', 'false']:
             return True if val.lower() == 'true' else False
-        if val.lower() == 'none':
+        if val == 'None':
             return None
         return val
 
--- a/tests/test_utils/test_config.py
+++ b/tests/test_utils/test_config.py
@@ -472,14 +472,15 @@ def test_dict_action():
     # Normal values
     args = parser.parse_args([
         '--options', 'item2.a=1', 'item2.b=0.1', 'item2.c=x', 'item3=false',
-        'item4=none'
+        'item4=none', 'item5=None'
     ])
     out_dict = {
         'item2.a': 1,
         'item2.b': 0.1,
         'item2.c': 'x',
         'item3': False,
-        'item4': None,
+        'item4': 'none',
+        'item5': None,
     }
     assert args.options == out_dict
     cfg_file = osp.join(data_path, 'config/a.py')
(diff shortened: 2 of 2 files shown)


HAOCHENYE (commented):
LGTM.

HAOCHENYE (approved):

vakker:
@HAOCHENYE is there anything else needed for this PR?

mmcv/utils/config.py
```diff
@@ -638,6 +638,8 @@ def _parse_int_float_bool(val):
             pass
         if val.lower() in ['true', 'false']:
             return True if val.lower() == 'true' else False
+        if val.lower() == 'none':
+            return None
```
zhouzaida, line 642: suggest moving those lines to line 693

vakker: Contextually it belongs here, because the code around 693 handles the different delimiters until it finds a single value, then this function handles the rest. 
Probably the name is not accurate any more, because it should be `_parse_int_float_bool_none` now.

zhouzaida: Agree to rename it to `_parse_int_float_bool_none`.

ZwwWayne: Sometimes string "none" is an available option, e.g., in losses like below
https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/models/losses/iou_loss.py#L250

When setting `reduction="none"` in loss, will line 641 change it from 'none' to None? This might cause failure.

vakker: That's a good point. However, this only has an impact if it's passed in from the command line, e.g. `--cfg-options loss.reduction=none`.

But to cover that case, I think either the parser should be refactored to only parse inputs as string if they're quoted, which might be a rather complex change and would lead to confusion, or just force the parser to only accept `None` as python `None` and `none` and string `"none"`, but then it's inconsistent with the bool case.

ZwwWayne: Yes, we could force the parser to only accept None as python None. We do not need to make it consistent with the bool case.

vakker: Done

ZwwWayne (approved):

zhouzaida (approved):

OpenMMLab-gd:
Hi @vakker！First of all, we want to express our gratitude for your significant PR in the MMCV project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

If you are Chinese or have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）
Thank you again for your contribution❤ @vakker

PR #1837, opened 2022-03-26
[Fix] Fix dump method of Config

jmercat:
## Motivation: 

When loading a config from a dict, the dump function does not work. Dump should not depend on how config was loaded but only where it is dumped.

## Modification:

 Changed checks for type in dump. It used to check the config source filename and I changed it to the destination filename. When there is no input file argument, it does not save the config in a file but returns a string. I am not sure this is the desired behavior but the format of this string depends on self.filename (format matching source format). I personally don’t think this is the best choice but I think this is the closest behavior to the previous version.

## Use cases (Optional)

It would offer the ability for a config loaded from a dict to be dumped.


mmcv/utils/config.py
```diff
@@ -556,7 +556,7 @@ def __setstate__(self, state):
 
     def dump(self, file=None):
         cfg_dict = super(Config, self).__getattribute__('_cfg_dict').to_dict()
-        if self.filename.endswith('.py'):
+        if file is not None and file.endswith('.py'):
             if file is None:
                 return self.pretty_text
             else:
```
HAOCHENYE, line 562: Thanks for your contribution and the original dump logic is actually weird. The argument `file` is judged twice here, maybe the logic can be refined?

mmcv/utils/config.py
```diff
@@ -565,7 +565,7 @@ def dump(self, file=None):
         else:
             import mmcv
             if file is None:
-                file_format = self.filename.split('.')[-1]
+                file_format = file.split('.')[-1]
```
HAOCHENYE, line 568: `None ` object does not have `split ` method.

HAOCHENYE (commented):
It could be better if you can provide a corresponding unit test.

jmercat:
oops thanks for the review, I didn’t think this through I’ll give it another try...

jmercat:
When there is no input file argument, it does not save the config in a file but returns a string. I am not sure this is the desired behavior but the format of this string depends on self.filename (format of the string matching source format). I personally don’t think this is the best choice but I think this is the closest behavior to the previous version. (To me the dump string format should always be the same regardless of the source of the config.)

jmercat:
When there is no input file argument, it does not save the config in a file but returns a string. I am not sure this is the desired behavior but the format of this string depends on self.filename (format of the string matching source format). I personally don’t think this is the best choice but I think this is the closest behavior to the previous version. (To me the dump string format should always be the same regardless of the source of the config.)

mmcv/utils/config.py
```diff
@@ -555,20 +555,29 @@ def __setstate__(self, state):
...
+            The formatting of this returned string is defined by the extension of self.filename.
+            If self.filename is not defined, returns a string representation of a dict (lowercased and using ' for strings).
+        """
+        import mmcv
         cfg_dict = super(Config, self).__getattribute__('_cfg_dict').to_dict()
-        if self.filename.endswith('.py'):
-            if file is None:
+        if file is None:
+            if self.filename is None:
+                return mmcv.dump(cfg_dict)
```
HAOCHENYE, line 568: `mmcv.dump` will raise a `ValueError` if `file_format` and `file` is None at the same time. May we can directly dump a `.py` config file.

tests/test_utils/test_config.py
```diff
@@ -427,7 +427,25 @@ def test_dict():
         assert cfg.item1 == 0
         cfg.update(dict(item2=dict(a=1)))
         assert cfg.item2.a == 1
-
+        
+        
+@pytest.mark.parametrize(
+    "file", ["a.json", "b.py", "c.yaml"]
+)
+def test_fromdict(file):
```
HAOCHENYE, line 435: `test_dump_from_dict` could be more comfortable.

HAOCHENYE:
> When there is no input file argument, it does not save the config in a file but returns a string. I am not sure this is the desired behavior but the format of this string depends on self.filename (format of the string matching source format). I personally don’t think this is the best choice but I think this is the closest behavior to the previous version. (To me the dump string format should always be the same regardless of the source of the config.)

Besides, it seems the original implementation will return python-format pretty_text regardless of the file format of `file`.  It's a little weird, Maybe @zhouzaida could have a look.

HAOCHENYE:
@jmercat Hi.

> When there is no input file argument, it does not save the config in a file but returns a string. I am not sure this is the desired behavior but the format of this string depends on self.filename (format of the string matching source format). I personally don’t think this is the best choice but I think this is the closest behavior to the previous version. (To me the dump string format should always be the same regardless of the source of the config.)

Hi~ After discussing with my partners, your code implementation makes more sense and will not raise BC breaking risk. But there exists a small bug, If `file` and `self.filename` are `None` at the same time,  `mmcv.dump` will raise an error.

mmcv/utils/config.py
```diff
@@ -555,20 +555,27 @@ def __setstate__(self, state):
         super(Config, self).__setattr__('_text', _text)
 
     def dump(self, file=None):
+        """
+            If a file argument is given, saves the config to that file using the format defined by the file argument extension.
+            Otherwise, returns a string representing the config.
+            The formatting of this returned string is defined by the extension of self.filename.
+            If self.filename is not defined, returns a string representation of a dict (lowercased and using ' for strings).
+        """
+        import mmcv
```
HAOCHENYE, line 564: Thanks for your docstring! It could be better to add an overall description and remove some indents. Besides, The `If a file argument is given.....` is too long, maybe we could start a new line~
```suggestion
        """Dump config to file or return config text.
        
        If a file argument is given, saves the config to that file using the format defined by the file argument extension.
        Otherwise, returns a string representing the config.
        The formatting of this returned string is defined by the extension of self.filename.
        If self.filename is not defined, returns a string representation of a dict (lowercased and using ' for strings).
        """
        import mmcv
```

mmcv/utils/config.py
```diff
@@ -555,20 +555,27 @@ def __setstate__(self, state):
         super(Config, self).__setattr__('_text', _text)
 
     def dump(self, file=None):
+        """
+            If a file argument is given, saves the config to that file using the format defined by the file argument extension.
+            Otherwise, returns a string representing the config.
+            The formatting of this returned string is defined by the extension of self.filename.
+            If self.filename is not defined, returns a string representation of a dict (lowercased and using ' for strings).
+        """
```
HAOCHENYE, line 563: Thanks for your docstring! It could be better to add an overall description and remove some indents. Besides, The `If a file argument is given.....` is too long, maybe we could start a new line~
```suggestion
        """Dump config to file or return config text.
        
        If a file argument is given, saves the config to that file using the format defined by the file argument extension.
        Otherwise, returns a string representing the config.
        The formatting of this returned string is defined by the extension of self.filename.
        If self.filename is not defined, returns a string representation of a dict (lowercased and using ' for strings).
        """
```

HAOCHENYE:
@jmercat Thanks for your docstring. But it seems the format is a little weird. Have you committed your code with pre-commit(https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md)? Besides, 
image
should be signed~

zhouzaida:
Hi @jmercat , is there any progress?

jmercat:
I added other email addresses to my account. I think it should solve the CLA signing problem.
I installed and run the pre-commit and made changes in the docstring.

jmercat:
This is a good learning experience for me but I think that for such a small change, the formatting, cla, and all add an overhead that is larger than the contribution. Next time I’ll probably just write an issue, propose a code change there and let a bigger contributor implement it. It might even be faster for you, then you won’t need all these discussions as we had here. You might want to warn about this in the [contribution guidelines.](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md)

zhouzaida:
> This is a good learning experience for me but I think that for such a small change, the formatting, cla, and all add an overhead that is larger than the contribution. Next time I’ll probably just write an issue, propose a code change there and let a bigger contributor implement it. It might even be faster for you, then you won’t need all these discussions as we had here. You might want to warn about this in the [contribution guidelines.](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md)

Hi @jmercat, thanks for your contribution. You are welcome to open an issue to report anything.

tests/test_utils/test_config.py
```diff
@@ -429,6 +429,26 @@ def test_dict():
         assert cfg.item2.a == 1
 
 
+@pytest.mark.parametrize('file', ['a.json', 'b.py', 'c.yaml', 'd.yml', None])
+def test_dump_from_dict(file):
```
zhouzaida, line 433: Suggest renaming `test_dump_from_dict` to `test_dump` and add a unit test that its `self.filename` is not None.

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
...
+    assert cfg.item3 == cfg_dict['item3']
+    assert cfg.item4 == cfg_dict['item4']
+    assert cfg._filename is None
+    if file is not None:
+        # dump without a filename argument is only returning pretty_text.
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            dump_file = osp.join(temp_config_dir, file)
+            cfg.dump(dump_file)
+            dumped_cfg = Config.fromfile(dump_file)
+            assert dumped_cfg._cfg_dict == cfg._cfg_dict
```
zhouzaida, line 454:
```suggestion
        with tempfile.TemporaryDirectory() as temp_dir:
            path = osp.join(temp_dir, file)
            cfg.dump(path)
            dumped_cfg = Config.fromfile(path)
            assert dumped_cfg._cfg_dict == cfg._cfg_dict
```

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
...
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            dump_file = osp.join(temp_config_dir, file)
+            cfg.dump(dump_file)
+            dumped_cfg = Config.fromfile(dump_file)
+        if not file.endswith('.json'):
+            assert dumped_cfg._cfg_dict == cfg._cfg_dict
+        else:
+            check_dump_json_cfg(cfg._cfg_dict, dumped_cfg._cfg_dict)
+    else:
+        cfg.dump() == cfg.pretty_text
```
zhouzaida, line 480:
```suggestion
        assert cfg.dump() == cfg.pretty_text
```

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
...
+        # dump without a filename argument is only returning pretty_text.
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            dump_file = osp.join(temp_config_dir, file)
+            cfg.dump(dump_file)
+            dumped_cfg = Config.fromfile(dump_file)
+            assert dumped_cfg._cfg_dict == cfg._cfg_dict
+    else:
+        assert cfg.dump() == cfg.pretty_text
+
+    # The key of json must be string, key `1` will be converted to `'1'`.
```
zhouzaida, line 458:
```suggestion
    # The key of json must be a string so key `1` will be converted to `'1'`.
```

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
         assert cfg.item2.a == 1
 
 
+@pytest.mark.parametrize('file', ['a.json', 'b.py', 'c.yaml', 'd.yml', None])
+def test_dump(file):
+    # Test dump config load from dict.
```
zhouzaida, line 440:
```suggestion
    # config loaded from dict
```

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
...
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            dump_file = osp.join(temp_config_dir, file)
+            cfg.dump(dump_file)
+            dumped_cfg = Config.fromfile(dump_file)
+            assert dumped_cfg._cfg_dict == cfg._cfg_dict
+    else:
+        assert cfg.dump() == cfg.pretty_text
+
+    # The key of json must be string, key `1` will be converted to `'1'`.
+    def check_dump_json_cfg(ori_cfg, dumped_json_cfg):
```
zhouzaida, line 459:
```suggestion
    def compare_json_config(ori_cfg, dumped_json_cfg):
```

tests/test_utils/test_config.py
```diff
@@ -435,6 +435,51 @@ def test_dict():
...
+                check_dump_json_cfg(value, dumped_json_cfg[str(key)])
+
+    # Test dump config.
+    cfg_file = osp.join(data_path, 'config/n.py')
+    cfg = Config.fromfile(cfg_file)
+    if file is not None:
+        with tempfile.TemporaryDirectory() as temp_config_dir:
+            dump_file = osp.join(temp_config_dir, file)
+            cfg.dump(dump_file)
+            dumped_cfg = Config.fromfile(dump_file)
```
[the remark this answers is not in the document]
HAOCHENYE: renamed to `cfg_file` consist with the previous convention。

mmcv/utils/config.py
```diff
@@ -561,20 +562,41 @@ def __setstate__(self, state):
...
+            >>> cfg_dict = dict(item1=[1, 2], item2=dict(a=0),
+            ...     item3=True, item4='test')
+            >>> cfg = Config(cfg_dict=cfg_dict)
+            >>> dump_file = "a.py"
+            >>> cfg.dump(dump_file)
+
+        Args:
+            file (optional string): path of the output file where the config
+                should be written
+        """
```
zhouzaida, line 586:
```suggestion
        """Dumps config into a file or returns a string representation of the
        config.

        If a file argument is given, saves the config to that file using the
        format defined by the file argument extension.

        Otherwise, returns a string representing the config. The formatting of
        this returned string is defined by the extension of `self.filename`. If
        `self.filename` is not defined, returns a string representation of a dict
        (lowercased and using ' for strings).

        Examples:
            >>> cfg_dict = dict(item1=[1, 2], item2=dict(a=0),
            ...     item3=True, item4='test')
            >>> cfg = Config(cfg_dict=cfg_dict)
            >>> dump_file = "a.py"
            >>> cfg.dump(dump_file)

        Args:
            file (str, optional): Path of the output file where the config
                will be dumped. Defaults to None.
        """
```

mmcv/utils/config.py
```diff
@@ -26,6 +26,8 @@
 else:
     import re
 
+import mmcv
+
```
zhouzaida, line 30:
```suggestion
```

mmcv/utils/config.py
```diff
@@ -561,20 +562,41 @@ def __setstate__(self, state):
...
+            >>> cfg_dict = dict(item1=[1, 2], item2=dict(a=0),
+            ...     item3=True, item4='test')
+            >>> cfg = Config(cfg_dict=cfg_dict)
+            >>> dump_file = "a.py"
+            >>> cfg.dump(dump_file)
+
+        Args:
+            file (str, optional): Path of the output file where the config
+                will be dumped. Defaults to None.
+        """
```
zhouzaida, line 586:
```suggestion
        """
        import mmcv
```

mmcv/utils/config.py
```diff
@@ -561,20 +561,42 @@ def __setstate__(self, state):
...
+        """Dumps config into a file or returns a string representation of the
+        config.
+
+        If a file argument is given, saves the config to that file using the
+        format defined by the file argument extension.
+
+        Otherwise, returns a string representing the config. The formatting of
+        this returned string is defined by the extension of `self.filename`. If
+        `self.filename` is not defined, returns a string representation of a dict
+        (lowercased and using ' for strings).
```
zhouzaida, line 573:
```suggestion
        `self.filename` is not defined, returns a string representation of a
         dict (lowercased and using ' for strings).
```

zhouzaida (approved):

PR #1845, opened 2022-03-29
remove redundant code

www516717402:
1. In template template `<typename F, F f>`， F type deduce from decltype(&(key)) and applay to `FunctionType funcs_[MAX_DEVICE_TYPES];`. However  `f` not used and only as explain code.

2. `static key##_##device##_registerer` is redundant since inst is static type and return reference.
``` c++
 static DeviceRegistry& instance() {
    static DeviceRegistry inst;
    return inst;
  }
```

### DIFF
--- a/mmcv/ops/csrc/common/pytorch_device_registry.hpp
+++ b/mmcv/ops/csrc/common/pytorch_device_registry.hpp
@@ -22,13 +22,9 @@ inline std::string GetDeviceStr(const at::Device& device) {
 }
 
 // Registry
-template <typename F>
-class DeviceRegistry;
-
-template <typename Ret, typename... Args>
-class DeviceRegistry<Ret (*)(Args...)> {
+template <typename FunctionType>
+class DeviceRegistry {
  public:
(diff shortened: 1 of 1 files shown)


grimoire:
Thanks for your contribution. But I am afraid we should not do these changes.
1. `f` is used to distinguish different functions. Let's say what if we have two functions with the same parameters and outputs. They will have the same function pointer type, which means they will share the same registry. That is not what we expected. (Some unit tests would fail `pytest tests/test_ops`)
2. There is a potential risk to remove `static`. We want to limit these `_register` inside their own compile unit, so code outside could safely create another variable or function with the same name. What if you remove `static`, There would be a lot of symbols of these `_register` in the library. Try:
```bash
readelf --wide -s mmcv/_ext.cpython-37m-x86_64-linux-gnu.so |grep _register
```
See what is the difference between static and dynamic.

www516717402:
@grimoire 
Sorry for commit uesless code and thanks for pointing out my mistake.

PR #1864, opened 2022-04-08
[Feature] Add decorator `prohibit_cache_randomness`

ly015:

## Motivation

Add decorator `prohibit_cache_randomness` for data transforms with non-cacheable random behavior. This decorator explicitly prevents unintentional use of such data transforms within the context of caching randomness which may lead to expected results.

## Modification

- Add `prohibit_caching_randomness` and its unit tests
- Modify `cache_random_params` to check if caching randomness is allowed

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


mzr1996:
Please add use cases, thanks~

mmcv/transforms/utils.py
```diff
@@ -79,10 +83,43 @@ def __get__(self, obj, cls):
         return self
 
 
+def prohibit_cache_randomness(cls):
```
ZwwWayne, line 86: 1. would avoid_cache_randomness more easy to understand?
2. should we consider the case like below:
```python

@prohibit_cache_randomness
class A(BaseTransforms):
    pass

class B(A):
    @cache_randomness
    def random_func(self)
```

ly015: For question 2, I am wondering if the subclass should inherit the attribute "avoid_cache_randomness" or not?

ZwwWayne: If we want to allow case 2, I think then class B should have different avoid_cache_randomness value from A

ly015: Ok, we can do that.

gaotongxiao (approved):
The title and the PR msg should also be updated.

gaotongxiao (approved):

gaotongxiao:
The Modification section needs to be update

PR #1872, opened 2022-04-12
[Fix] Fix spconv unit test by releasing cache on GPU

VVsssssk:
## Motivation
Some ops unit tests will report errors on circleci, and the main problem is out of memory. So I try to move some op ut to run in sub-threads; when the sub-thread finishes, the sub-thread will release the memory, thus avoiding out of memory.

### DIFF
--- a/tests/test_ops/test_spconv.py
+++ b/tests/test_ops/test_spconv.py
@@ -76,6 +76,7 @@ def make_sparse_convmodule(in_channels,
 @pytest.mark.skipif(
     not torch.cuda.is_available(), reason='requires CUDA support')
 def test_make_sparse_convmodule():
+    torch.cuda.empty_cache()
     voxel_features = torch.tensor([[6.56126, 0.9648336, -1.7339306, 0.315],
                                    [6.8162713, -2.480431, -1.3616394, 0.36],
                                    [11.643568, -4.744306, -1.3580885, 0.16],


tests/test_ops/test_spconv.py
```diff
@@ -127,3 +127,11 @@ def test_make_sparse_convmodule():
     assert isinstance(sparse_block1[0], torch.nn.BatchNorm1d)
     assert isinstance(sparse_block1[1], torch.nn.ReLU)
     assert isinstance(sparse_block1[2], SparseInverseConv3d)
+
+
+@pytest.mark.skipif(
+    not torch.cuda.is_available(), reason='requires CUDA support')
+def test_make_sparse_convmodule():
+    torch.multiprocessing.set_start_method('spawn')
+    test_process = Process(target=make_sparse_convmodule_forward)
```
zhouzaida, line 136: why do we need to use `Process` to test the op?

zhouzaida:
> ## Motivation
> Try to resolve spconv ut error in circle ci.

briefly describe why the test failed in circleci like `out of memory`.

VVsssssk:
# https://github.com/open-mmlab/mmdetection3d/pull/1388

tests/test_ops/test_carafe.py
```diff
@@ -1,28 +1,39 @@
...
-        if not torch.cuda.is_available():
-            return
-        from mmcv.ops import CARAFENaive
-        feat = torch.randn(
-            2, 64, 3, 3, requires_grad=True, device='cuda').double()
-        mask = torch.randn(
-            2, 100, 6, 6, requires_grad=True,
-            device='cuda').sigmoid().double()
-        gradcheck(CARAFENaive(5, 4, 2), (feat, mask), atol=1e-4, eps=1e-4)
+        process = Process(target=carafe_naive_gradcheck)
```
ZwwWayne, line 30: This modification seems to be weird, though I can understand the motivation.
Can we manually use `torch.cuda.empty_cache() ` to release the memory?

ZwwWayne:
Can be merged after fixing linting issues and passed the CI

zhouzaida (approved):

PR #1874, opened 2022-04-12
Sassd

wenbo-yu:

## Motivation
We added points_in_boxes_with_offsets_cpu_forward under /ops. The purpose is in addition to get the box indexes of each points, it also return the offsets (distance to the box center). The purpose is that we want to add a new model in mmdet3d. This function is crucial to calculate the loss. mmdet3dd suggested us to add this to mmcv.

## Modification
points_in_boxes_with_offsets_cpu_forward related functions under ops. We don't know the purpose of the file under ops/csrc/parrots. Though we implement the functions referring to the others.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/points_in_boxes.py
+++ b/mmcv/ops/points_in_boxes.py
@@ -4,7 +4,7 @@ from ..utils import ext_loader
 
 ext_module = ext_loader.load_ext('_ext', [
     'points_in_boxes_part_forward', 'points_in_boxes_cpu_forward',
-    'points_in_boxes_all_forward'
+    'points_in_boxes_cpu_forward_with_offsets', 'points_in_boxes_all_forward'
 ])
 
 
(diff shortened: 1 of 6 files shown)


zhouzaida:
Hi @wenbo-yu , thanks for your contribution. The lint failed and you can refer to https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md#python to install pre-commit to automatically fix the lint problem.

filaPro:
Just out of curiosity, why not to call `points_in_boxes` and then calculate offsets in `pytorch`? Will it be noticeably slower?

mmcv/ops/points_in_boxes.py
```diff
@@ -94,6 +94,51 @@ def points_in_boxes_cpu(points, boxes):
     return point_indices
 
 
+def points_in_boxes_with_offsets_cpu(points, boxes):
```
AllentDan, line 97: may add type hint here

AllentDan (commented):
Please fix the lint error and add a unit test. We also need to update those ops in https://github.com/open-mmlab/mmcv/blob/master/docs/en/understand_mmcv/ops.md and https://github.com/open-mmlab/mmcv/blob/master/docs/zh_cn/understand_mmcv/ops.md

PR #1780, opened 2022-03-09
[Feature] Support IPU

HuDi2018:
ipu model training and inference are basing on static graph.
so we need to run model training/inference once to compile graph on IPU.
all the convert and config parsing things are controlled by ipu_runner

## Motivation

implement mmcv on IPU and support a demo(ViT training) on IPU

## Modification

add an ipu_runner and some help functions and classes in runner/ipu

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

use case is in  a PR of mmclassification which will be referenced in comments

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls.
- [x] CLA has been signed and all committers have signed the CLA in this PR.

PR #1882, opened 2022-04-16
[Feature] Support MMCV on IPU

HumberMe:
This is a new duplicate of PR(https://github.com/open-mmlab/mmcv/pull/1780), the old PR has some permission problem
ipu model training and inference are basing on static graph.
so we need to run model training/inference once to compile graph on IPU.
all the convert and config parsing things are controlled by ipu_runner

## Motivation

implement mmcv on IPU and support a demo(ViT training) on IPU

## Modification

add an ipu_runner and some help functions and classes in runner/ipu

## BC-breaking (Optional)

## Use cases (Optional)

use case is in  a PR of mmclassification which will be referenced in comments

## Checklist

**Before PR**:

**After PR**:


zhouzaida (approved):

mmcv/device/ipu/dataloader.py
```diff
@@ -0,0 +1,157 @@
...
+            size of the tensors loaded by the workers.
+            Default to the combined batch size.
+            If specified the ``rebatched_worker_size`` must be less than
+            or equal to the combined batch size.
+        kwargs (Dict[str, Any], optional): Other options to pass to PyTorch's
+            ``DataLoader`` constructor.
+    """
+
+    def __init__(self,
+                 options,
```
ZwwWayne, line 100: Recommend to switch the order between options and dataset, since dataset should be the most important arguments of this module

ZwwWayne: If we want to change the order, we should also update the order in the docstring

mmcv/device/ipu/utils.py
```diff
@@ -0,0 +1,244 @@
...
+            options_node(cfg)
+        else:
+            error_msg = f'options_node type {type(options_node)} not supported'
+            raise NotImplementedError(error_msg)
+    else:
+        error_msg = f'cfg type {type(cfg)} not supported'
+        raise NotImplementedError(error_msg)
+
+
+def cast_to_options(cfg):
```
ZwwWayne, line 31: cast_to_options -> cfg2options might be more accurate since the type of cfg should be dict or Config?

tests/test_device/test_ipu/test_ipu_runner.py
```diff
@@ -0,0 +1,126 @@
...
+import torch.nn as nn
+from torch.utils.data import Dataset
+
+from mmcv.runner import build_runner
+from mmcv.utils import IS_IPU_AVAILABLE
+
+if IS_IPU_AVAILABLE:
+    from mmcv.device.ipu import IPUDataLoader, runner
+
+skip_no_ipu = pytest.mark.skipif(
```
ZwwWayne, line 16: It seems that this function is in many files, can we put it into mmcv.utils.testing?

ZwwWayne:
Final review comments. This PR can be merged after resolving the remaining comments.

tests/test_device/test_ipu/test_ipu_runner.py
HumberMe: copy this usage from other tests

zhouzaida (approved):

ZwwWayne (approved):

PR #1902, opened 2022-04-21
[Enhance] Replace BEV IoU with 3D IoU

filaPro:
## Motivation

BEV IoU is not needed now in `open-mmlab` projects after [replacing](https://github.com/open-mmlab/mmdetection3d/pull/1403) `iou3d` ops with `box_iou_rotated` and `nms` ops. So now we can support fair 3d IoU instead of BEV IoU. This was discussed with @ZwwWayne and @Tai-Wang as a part of work on FCAF3D in `mmdetection3d`.

## Modification

Modify `python` and `cuda` functions in `mmcv.ops.iou3d` to accept not `(x1, y1, x2, y2, heading)` but `(x, y, z, dx, dy, dz, heading)` boxes. 

## BC-breaking (Optional)

It breaks all 3 functions in `mmcv.ops.iou3d`.

## Use cases (Optional)

Fair 3D IoU is a must have function for indoor 3D object detection.

May be @grimoire to have a look on `cuda` part.


grimoire (approved):
LGTM

zhouzaida:
Please @Tai-Wang have a look

grimoire:
As current mmdetection3d master is using bev iou ops. Have you PR a fix there?

Tai-Wang:
> As current mmdetection3d master is using bev iou ops. Have you PR a fix there?

@grimoire A related PR in mmdet3d has been merged into the dev branch and will be merged into master in the next release soon.

Tai-Wang:
Hi @filaPro , great work! The only concern is that does the current modification remove the original ops directly? If other users use related bev ops in their projects, it will trigger errors? If yes, I think it would be better to keep original ones for several mmcv versions and just throw deprecated warnings to avoid these potential issues.

Besides, if possible, I recommend we can still support the bev ops based on these more comprehensive 3D ops. (I think the deprecated bev ops should be just special cases of these 3D ops?)

filaPro:
Hi @Tai-Wang,

Deprecated functions can be replaced with either `box_iou_rotated` + `nms` or novel `iou3d` ops. We chose the first option in `mmdetection3d`. But yes, we can probably support the second options here with `DeprecationWarning`.

filaPro:
So, I reverted 3 original python functions with minimal changes in code. Also, I checked that original 3 tests are also passing now.

Tai-Wang (approved):

mmcv/ops/iou3d.py
```diff
@@ -19,21 +109,28 @@ def boxes_iou_bev(boxes_a, boxes_b):
     Returns:
         torch.Tensor: IoU result with shape (M, N).
     """
+    warnings.warn(
+        '`iou3d.boxes_iou_bev` is deprecated and will be removed in'
+        ' the future. Please, use `box_iou_rotated.box_iou_rotated`.',
+        DeprecationWarning)
     ans_iou = boxes_a.new_zeros(
         torch.Size((boxes_a.shape[0], boxes_b.shape[0])))
+    boxes_a = _xyxyr2xyzwlhr(boxes_a)
```
ZwwWayne, line 118: It might be better to directly use box_iou_rotated.box_iou_rotated here, which makes boxes_iou_bev as a wrapper of box_iou_rotated.box_iou_rotated. In this way we do not need `_xyxyr2xyzwlhr` and the code is more straightforward, which tells readers that they are indeed the same as box_iou_rotated.box_iou_rotated

mmcv/ops/iou3d.py
```diff
@@ -65,25 +164,28 @@ def nms_bev(boxes, scores, thresh, pre_max_size=None, post_max_size=None):
...
         torch.Tensor: Remaining indices with scores in descending order.
     """
+    warnings.warn(
+        '`iou3d.nms_normal_bev` is deprecated and will be removed in'
+        ' the future. Please, use `nms.nms`.', DeprecationWarning)
     assert boxes.shape[1] == 5, 'Input boxes shape should be [N, 5]'
     order = scores.sort(0, descending=True)[1]
 
-    boxes = boxes[order].contiguous()
+    boxes = _xyxyr2xyzwlhr(boxes)[order].contiguous()
```
ZwwWayne, line 185: Similar suggestion as that for boxes_iou_bev. We can simply use nms.nms here to simplify the code without breaking the compatibility.

mmcv/ops/iou3d.py
```diff
@@ -1,14 +1,104 @@
...
+    keep = torch.zeros(boxes.size(0), dtype=torch.long)
+    num_out = torch.zeros(size=(), dtype=torch.long)
+    ext_module.iou3d_nms3d_forward(
+        boxes, keep, num_out, nms_overlap_thresh=iou_threshold)
+    keep = order[keep[:num_out].cuda(boxes.device)].contiguous()
+    return keep
+
+
+def nms3d_normal(boxes, scores, iou_threshold):
+    """Normal 3D NMS function GPU implementation. The overlap of two boxes for
```
ZwwWayne, line 58: GPU implementation of normal 3D NMS. The overlap of two 3D boxes is calculated as the exact overlapping area of the two boxes `with their yaw angle set to 0`.

filaPro:
Hi @ZwwWayne ,

Sorry for late reply. I've update the deprecated bev iou/nms with current mmcv ops (without 3d nms/iou). Also checked that old tests are still passed.

ZwwWayne (approved):

PR #1940, opened 2022-05-08
Imporve ability to view usage data during training

wlf-darkmatter:
## Motivation

The wrong labeling of the dataset may cause problems such as gradient explosion during the training process. The wrong labeling can be found by setting up a hook function.

## Problem

However, when a for loop is used to traverse the iterator of the pytorch `DataLoader` class, the `hook` cannot obtain the information of the currently read image, and thus cannot determine the source of the error.

## Modification

Load the data_batch information into the `runner` during the train process, and then pass it to the hook function to solve the problem.

It's just a very small imporvement.

### DIFF
--- a/mmcv/runner/epoch_based_runner.py
+++ b/mmcv/runner/epoch_based_runner.py
@@ -62,6 +62,7 @@ class EpochBasedRunner(BaseRunner):
         self.call_hook('before_val_epoch')
         time.sleep(2)  # Prevent possible deadlock during epoch transition
         for i, data_batch in enumerate(self.data_loader):
+            self.data_batch = data_batch
             self._inner_iter = i
             self.call_hook('before_val_iter')
             self.run_iter(data_batch, train_mode=False)
--- a/mmcv/runner/epoch_based_runner.py
+++ b/mmcv/runner/epoch_based_runner.py
@@ -50,6 +50,7 @@ class EpochBasedRunner(BaseRunner):
             self.call_hook('before_train_iter')
             self.run_iter(data_batch, train_mode=True, **kwargs)
             self.call_hook('after_train_iter')
+            del self.data_batch
             self._iter += 1
 
         self.call_hook('after_train_epoch')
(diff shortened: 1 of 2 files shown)


zhouzaida:
Hi, @wlf-darkmatter , thanks for your contribution. Please also update https://github.com/open-mmlab/mmcv/blob/a86e807105d28a58565fa3a199a1f8eb47e1c548/mmcv/runner/epoch_based_runner.py#L48 and https://github.com/open-mmlab/mmcv/blob/a86e807105d28a58565fa3a199a1f8eb47e1c548/mmcv/runner/iter_based_runner.py#L76

wlf-darkmatter:
Thanks, I have updated the files you suggested, please check.

mmcv/runner/epoch_based_runner.py
```diff
@@ -45,6 +45,7 @@ def train(self, data_loader, **kwargs):
         self.call_hook('before_train_epoch')
         time.sleep(2)  # Prevent possible deadlock during epoch transition
         for i, data_batch in enumerate(self.data_loader):
+            self.data_batch = data_batch
```
zhouzaida, line 48: How about deleting `self.data_batch` in line 53 to strict the scope of `runner.data_batch`

```
del self.data_batch
```

zhouzaida, line 48: How about deleting `self.data_batch` in line 53 to strict the scope of `runner.data_batch`

```
del self.data_batch
```

wlf-darkmatter: Ok, just saw the message, I've made changes

zhouzaida (approved):

OpenMMLab-Assistant-004:
Hi @wlf-darkmatter ！First of all, we want to express our gratitude for your significant PR in the mmcv project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

If you have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

Issue #1261, opened 2021-08-12
[Bug] #1252 seems causing `KeyError: 'data_time'` when evaluation interval is a multiple of logging interval


haotian-liu:
By default, `log_config.interval=50`.
When `evaluation.interval={50,200}`, there is an error _KeyError: 'data_time'_ (error log pasted below).
When `evaluation.interval=120`, the error does not occur at first several iterations (because the evaluation is quite long, I was unable to try testing when it reaches iteration 600, but I guess that it might trigger the error when it reaches there).
Not sure if there is anything that I am not doing correctly. Thank you for your great work and effort in maintaining this!

```
  File "/home/haotian/code/mmcv/mmcv/runner/hooks/logger/text.py", line 96, in _log_info
    log_str += f'time: {log_dict["time"]:.3f}, ' \
KeyError: 'data_time'
Traceback (most recent call last):
  File "./tools/train.py", line 223, in <module>
    main()
  File "./tools/train.py", line 219, in main
    meta=meta)
  File "/home/haotian/code/mmdetection3d/mmdet3d/apis/train.py", line 25, in train_model
    meta=meta)
  File "/home/haotian/code/mmsegmentation/mmseg/apis/train.py", line 116, in train_segmentor
    runner.run(data_loaders, cfg.workflow)
  File "/home/haotian/code/mmcv/mmcv/runner/iter_based_runner.py", line 133, in run
    iter_runner(iter_loaders[i], **kwargs)
  File "/home/haotian/code/mmcv/mmcv/runner/iter_based_runner.py", line 66, in train
    self.call_hook('after_train_iter')
  File "/home/haotian/code/mmcv/mmcv/runner/base_runner.py", line 307, in call_hook
    getattr(hook, fn_name)(self)
  File "/home/haotian/code/mmcv/mmcv/runner/hooks/logger/base.py", line 152, in after_train_iter
    self.log(runner)
  File "/home/haotian/code/mmcv/mmcv/runner/hooks/logger/text.py", line 177, in log
    self._log_info(log_dict, runner)
  File "/home/haotian/code/mmcv/mmcv/runner/hooks/logger/text.py", line 96, in _log_info
    log_str += f'time: {log_dict["time"]:.3f}, ' \
```

haotian-liu:
This seems to happen when #1252 is pushed to master branch by @zhouzaida. It would be really helpful if Zaida could kindly take a look. Thanks!

zhouzaida:
[...]

PR #1972, opened 2022-05-20
[Fix] Fix invalid access to training-state log while evaluation

mob5566:
## Motivation

Previously, the LoggerHook identified the mode of log_buffer by checking the 'time' attribute in the log_buffer of the runner, which is not reasonable and causes an issue ([open-mmlab/mmsegmentaion#1502](https://github.com/open-mmlab/mmsegmentation/issues/1502)). Also, in #1252 and #1261, suggesting setting the priority of EvalHook to "LOW" was just a workaround and did not solve the root cause of this issue.

To solve this issue, it should identify the mode by a reliable state flag such that the LoggerHook can identify the correct mode of the log_buffer of the runner during training. Therefore, in this commit, I add a 'log_state' attribute while evaluating to help identify the correct mode.

## Modification

 * Fix opem-mmlab/mmsegmentation#1502 (KeyError: 'data_time')

 * Add an attribute 'log_state' to log_buffer during evaluation

 * Identify the log_buffer mode by 'log_state' attribute instead of
   'time' attribute

 * No need to log the 'log_state'

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tests/test_runner/test_eval_hook.py
+++ b/tests/test_runner/test_eval_hook.py
@@ -464,7 +464,7 @@ def test_logger(runner, by_epoch, eval_hook_priority):
     with tempfile.TemporaryDirectory() as tmpdir:
         logger = get_logger('test_logger')
         optimizer = optim.SGD(model.parameters(), lr=0.001, momentum=0.9)
-        runner = EpochBasedRunner(
+        runner = runner(
             model=model, optimizer=optimizer, work_dir=tmpdir, logger=logger)
         runner.register_logger_hooks(
             dict(


HAOCHENYE (commented):
Thanks for your contributions! using `time` to infer `mode` is actually weird. But should `log_state` reset to 'train' in 'after_train_iter' or 'after_train_epoch' ?
Besides， please sign the CLA~~
image

mob5566:
> But should `log_state` reset to 'train' in 'after_train_iter' or 'after_train_epoch' ?
Sure! That's a good idea. I will submit an amendment ASAP. Thanks.

mob5566:
Hi @HAOCHENYE ,

Instead of using `log_state`, I added a property `mode` for `LogBuffer` to present the mode of the content of the log buffer.
Before each iteration, the mode will be set to a specific mode according to its runner's mode.
While evaluating the model by EvalHook, EvalHook will set the mode to `'val'` for evaluation.

By the way, I found that `test_logger` did not use the parameter `runner` accordingly, and I fixed it with another commit.

HAOCHENYE:
> Hi @HAOCHENYE ,
> 
> Instead of using `log_state`, I added a property `mode` for `LogBuffer` to present the mode of the content of the log buffer. Before each iteration, the mode will be set to a specific mode according to its runner's mode. While evaluating the model by EvalHook, EvalHook will set the mode to `'val'` for evaluation.
> 
> By the way, I found that `test_logger` did not use the parameter `runner` accordingly, and I fixed it with another commit.

Hi~ Does it feel more comfortable that mode would be more appropriate as an attribute of the runner? In fact, the runner has the mode attribute, but it seems that it is not set correctly during training.

mob5566:
> Hi~ This reminds me: why we don't use `runner.mode` ? In fact, the runner has the mode attribute, but it seems that it is not set correctly during training.

I think both ways work. But, I think it's more like "evaluating the model during the training phase (mode) of a runner", so it's much clear to define *the mode of the `log_buffer`*. Moreover, I'm not sure if the mode of the runner changed, there would be any potential risk. Therefore, I decided to add a mode attribute for the `log_buffer`.

Issue #1948, opened 2022-05-10
Easy contributions! Add type annotation for mmcv~


HAOCHENYE:
Hi guys, we are planning to add type annotation for python code in MMCV. PRs of adding type annotation from community are welcomed :stuck_out_tongue_closed_eyes:. To make the community fully aware of the progress, we list the progress here. Please feel free to leave a message and create a PR if you are willing to add type annotation for one of the following modules  :rocket::rocket:.

- [x] runner part1
  - [x] [base_runner.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/base_runner.py)  #2003 
  - [x] [epoch_based_runner.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/epoch_based_runner.py)  #2003
  - [x] [iter_based_runner.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/iter_based_runner.py)  #2003
  - [x] [base_module.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/base_module.py)  #2020
  - [x] [builder.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/builder.py)  @ytzhao  #1985 
  - [x] [checkpoint.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/checkpoint.py)  #2020
  - [x] [default_constructor.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/default_constructor.py)  #2020
  - [x] [dist_utils.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/dist_utils.py)  #2020
  - [x] [fp16_utils.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/fp16_utils.py)  #2020
  - [x] [log_buffer.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/log_buffer.py)  #2020
  - [x] [priority.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/priority.py)  #2020
  - [x] [utils.py](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/utils.py)  #2008
  - [x] [optimizer](https://github.com/open-mmlab/mmcv/tree/master/mmcv/runner/optimizer)  #2008 

- [x] runner part2
[...]

PR #1991, opened 2022-05-26
Add type hints for mmcv/runner

triwahyuu:

## Motivation

completing my tasks for #1948 

## Modification

add type hints to:
- mmcv/runner/hooks/checkpoint.py
- mmcv/runner/hooks/closure.py
- mmcv/runner/hooks/ema.py
- mmcv/runner/hooks/evaluation.py
- mmcv/runner/hooks/lr_updater.py
- mmcv/runner/hooks/memory.py
- mmcv/runner/hooks/momentum_updater.py
- mmcv/runner/hooks/optimizer.py
- mmcv/runner/hooks/sync_buffer.py

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/runner/hooks/evaluation.py
+++ b/mmcv/runner/hooks/evaluation.py
@@ -2,6 +2,7 @@
 import os.path as osp
 import warnings
 from math import inf
+from typing import Callable, List, Optional
 
 import torch.distributed as dist
 from torch.nn.modules.batchnorm import _BatchNorm
@@ -83,17 +84,17 @@ class EvalHook(Hook):
     _default_less_keys = ['loss']
 
     def __init__(self,
-                 dataloader,
-                 start=None,
-                 interval=1,
-                 by_epoch=True,
-                 save_best=None,
-                 rule=None,
-                 test_fn=None,
-                 greater_keys=None,
-                 less_keys=None,
-                 out_dir=None,
-                 file_client_args=None,
+                 dataloader: DataLoader,
+                 start: Optional[int] = None,
+                 interval: int = 1,
+                 by_epoch: bool = True,
+                 save_best: Optional[str] = None,
+                 rule: Optional[str] = None,
+                 test_fn: Optional[Callable] = None,
+                 greater_keys: Optional[List[str]] = None,
+                 less_keys: Optional[List[str]] = None,
+                 out_dir: Optional[str] = None,
+                 file_client_args: Optional[dict] = None,
                  **eval_kwargs):
         if not isinstance(dataloader, DataLoader):
             raise TypeError(f'dataloader must be a pytorch DataLoader, '
@@ -131,6 +132,7 @@ class EvalHook(Hook):
             self.greater_keys = self._default_greater_keys
         else:
             if not isinstance(greater_keys, (list, tuple)):
+                assert isinstance(greater_keys, str)
                 greater_keys = (greater_keys, )
             assert is_seq_of(greater_keys, str)
             self.greater_keys = greater_keys
@@ -139,6 +141,7 @@ class EvalHook(Hook):
             self.less_keys = self._default_less_keys
         else:
             if not isinstance(less_keys, (list, tuple)):
+                assert isinstance(greater_keys, str)
                 less_keys = (less_keys, )
             assert is_seq_of(less_keys, str)
             self.less_keys = less_keys
@@ -150,7 +153,7 @@ class EvalHook(Hook):
         self.out_dir = out_dir
         self.file_client_args = file_client_args
 
-    def _init_rule(self, rule, key_indicator):
+    def _init_rule(self, rule: Optional[str], key_indicator: Optional[str]):
         """Initialize rule, key_indicator, comparison_func, and best score.
 
         Here is the rule to determine which rule is used for key indicator
@@ -178,6 +181,7 @@ class EvalHook(Hook):
             if key_indicator != 'auto':
                 # `_lc` here means we use the lower case of keys for
                 # case-insensitive matching
+                assert isinstance(key_indicator, str)
                 key_indicator_lc = key_indicator.lower()
                 greater_keys = [key.lower() for key in self.greater_keys]
                 less_keys = [key.lower() for key in self.less_keys]
@@ -439,20 +443,20 @@ class DistEvalHook(EvalHook):
     """
 
     def __init__(self,
-                 dataloader,
-                 start=None,
-                 interval=1,
-                 by_epoch=True,
-                 save_best=None,
-                 rule=None,
-                 test_fn=None,
-                 greater_keys=None,
-                 less_keys=None,
-                 broadcast_bn_buffer=True,
-                 tmpdir=None,
-                 gpu_collect=False,
-                 out_dir=None,
-                 file_client_args=None,
+                 dataloader: DataLoader,
+                 start: Optional[int] = None,
+                 interval: int = 1,
+                 by_epoch: bool = True,
+                 save_best: Optional[str] = None,
+                 rule: Optional[str] = None,
+                 test_fn: Optional[Callable] = None,
+                 greater_keys: Optional[List[str]] = None,
+                 less_keys: Optional[List[str]] = None,
+                 broadcast_bn_buffer: bool = True,
+                 tmpdir: Optional[str] = None,
+                 gpu_collect: bool = False,
+                 out_dir: Optional[str] = None,
+                 file_client_args: Optional[dict] = None,
                  **eval_kwargs):
 
         if test_fn is None:
--- a/mmcv/runner/hooks/evaluation.py
+++ b/mmcv/runner/hooks/evaluation.py
@@ -153,7 +153,7 @@ class EvalHook(Hook):
         self.out_dir = out_dir
         self.file_client_args = file_client_args
 
-    def _init_rule(self, rule: Optional[str], key_indicator: Optional[str]):
+    def _init_rule(self, rule: Optional[str], key_indicator: str):
         """Initialize rule, key_indicator, comparison_func, and best score.
 
         Here is the rule to determine which rule is used for key indicator
--- a/mmcv/runner/hooks/momentum_updater.py
+++ b/mmcv/runner/hooks/momentum_updater.py
@@ -30,9 +30,9 @@ class MomentumUpdaterHook(Hook):
         self.warmup_iters = warmup_iters
         self.warmup_ratio = warmup_ratio
 
-        self.base_momentum: Union[List[float], Dict[str, List[float]]] = [
+        self.base_momentum: Union[list, dict] = [
         ]  # initial momentum for all param groups
-        self.regular_momentum: Union[List[float], Dict[str, List[float]]] = [
+        self.regular_momentum: Union[list, dict] = [
         ]  # expected momentum if no warming up is performed
 
     def _set_momentum(self, runner, momentum_groups):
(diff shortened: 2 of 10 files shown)


mmcv/runner/hooks/evaluation.py
```diff
@@ -150,7 +153,7 @@ def __init__(self,
         self.out_dir = out_dir
         self.file_client_args = file_client_args
 
-    def _init_rule(self, rule, key_indicator):
+    def _init_rule(self, rule: Optional[str], key_indicator: Optional[str]):
```
HAOCHENYE, line 156: It seems there is not default `None` value for  `rule` and `key_indicator`. 
```suggestion
    def _init_rule(self, rule: Optional[str] = None, key_indicator: str):
```
Will this be more properer?

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -49,28 +52,33 @@ def _set_momentum(self, runner, momentum_groups):
                 elif 'betas' in param_group.keys():
                     param_group['betas'] = (mom, param_group['betas'][1])
 
-    def get_momentum(self, runner, base_momentum):
+    def get_momentum(self, runner, base_momentum) -> float:
```
HAOCHENYE, line 55: type hint for `base_momentum`

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -49,28 +52,33 @@ def _set_momentum(self, runner, momentum_groups):
...
                 for _base_momentum in self.base_momentum
             ]
 
-    def get_warmup_momentum(self, cur_iters):
+    def get_warmup_momentum(
+            self,
+            cur_iters: int) -> Union[List[float], Dict[str, List[float]]]:
 
         def _get_warmup_momentum(cur_iters, regular_momentum):
+            warmup_momentum = None
```
HAOCHENYE, line 81: Why we need to set `warmup_momentum=None`?

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -404,11 +426,13 @@ class OneCycleMomentumUpdaterHook(MomentumUpdaterHook):
     """
 
     def __init__(self,
-                 base_momentum=0.85,
-                 max_momentum=0.95,
-                 pct_start=0.3,
-                 anneal_strategy='cos',
-                 three_phase=False,
+                 base_momentum: Union[float, List[float], Dict[str,
+                                                               float]] = 0.85,
```
HAOCHENYE, line 430: The indent is weird, will this format better?
```suggestion
                 base_momentum: Union[
                     float, List[float], Dict[str, float]] = 0.85,
```

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -404,11 +426,13 @@ class OneCycleMomentumUpdaterHook(MomentumUpdaterHook):
     """
 
     def __init__(self,
-                 base_momentum=0.85,
-                 max_momentum=0.95,
-                 pct_start=0.3,
-                 anneal_strategy='cos',
-                 three_phase=False,
+                 base_momentum: Union[float, List[float], Dict[str,
```
HAOCHENYE, line 429: It could be better to modify the docstring

mmcv/runner/hooks/evaluation.py
triwahyuu: I don't think it will work since positional arguments should come first before keyword arguments   
I actually just followed the function docs here, but from the use case I would think it would be better to remove both the `Optional` type hints. what do you think?

mmcv/runner/hooks/momentum_updater.py
triwahyuu: I have added this before, but is declined by mypy because the function signature in [`OneCycleMomentumUpdaterHook.get_momentum`](https://github.com/triwahyuu/mmcv/blob/54292936e3c189fc21d908c6855cdbd49e998e1b/mmcv/runner/hooks/momentum_updater.py#L563) is different compared to other child class from this.  
Any idea how to resolve this better?

mmcv/runner/hooks/momentum_updater.py
triwahyuu: this is to resolve "possibly unbound" warning from Pylance. Other solution to this is to use `else` inplace of `elif self.warmup == 'exp'` since `self.warmup` has already been check to have those 3 possible values, how do you think?

mmcv/runner/hooks/momentum_updater.py
triwahyuu: you mean to match the type hints?

triwahyuu:
Hi @HAOCHENYE, there are several files that I think could not be added the type hints any more, such as `mmcv/runner/hooks/sampler_seed.py` and `mmcv/runner/hooks/iter_timer.py` since it only has `runner` type hints, but I couldn't add it since it will cause circular imports

zhouzaida:
> Hi @HAOCHENYE, there are several files that I think could not be added the type hints any more, such as `mmcv/runner/hooks/sampler_seed.py` and `mmcv/runner/hooks/iter_timer.py` since it only has `runner` type hints, but I couldn't add it since it will cause circular imports

got it

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -28,8 +30,9 @@ def __init__(self,
         self.warmup_iters = warmup_iters
         self.warmup_ratio = warmup_ratio
 
-        self.base_momentum = []  # initial momentum for all param groups
-        self.regular_momentum = [
+        self.base_momentum: Union[List[float], Dict[str, List[float]]] = [
+        ]  # initial momentum for all param groups
+        self.regular_momentum: Union[List[float], Dict[str, List[float]]] = [
         ]  # expected momentum if no warming up is performed
```
zhouzaida, line 36: The type is a bit long. Can we simplify it with `Union[List, Dict]`?

triwahyuu: in that case I think it would be better to use `Union[list, dict]`, I think that would be better, will fix this

triwahyuu:
anyway the unit test is failed there, but I don't think I touched that part of the code, any idea whether I should fix this? @zhouzaida  @HAOCHENYE

mmcv/runner/hooks/evaluation.py
HAOCHENYE: Agree. I think the docstring is correct. However it forgets to give the default value for `rule`. BTW, could it passes through mypy if we simply setting rule as `rule: str`?

mmcv/runner/hooks/momentum_updater.py
HAOCHENYE: allright, If the signature of the function has changed, we don't need to add Type Hints.

mmcv/runner/hooks/momentum_updater.py
HAOCHENYE: OK, I thought adding an extra else after `'exp'` to raise an error could be better. The original implementation is at risk of returning an undefined value.

mmcv/runner/hooks/momentum_updater.py
HAOCHENYE: It seems the `dict` type of base_momentum is new.

mmcv/runner/hooks/momentum_updater.py
triwahyuu: okay, I think I can resolve this then

mmcv/runner/hooks/momentum_updater.py
triwahyuu: yepp, I already added this in the new commit

mmcv/runner/hooks/momentum_updater.py
triwahyuu: but from the implementation it is also expected to have dict as well

mmcv/runner/hooks/evaluation.py
triwahyuu: apparently it does not passes, so I think the `rule` still needs the optional but `key_indicator` is not. How do you think?

zhouzaida:
> anyway the unit test is failed there, but I don't think I touched that part of the code, any idea whether I should fix this? @zhouzaida @HAOCHENYE

Hi, sorry for my late reply. This branch has conflicts that need to be resolved.

triwahyuu:
done @zhouzaida

zhouzaida:
Hi, we also need to add type hints for other methods in Hooks.

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -28,8 +30,9 @@ def __init__(self,
         self.warmup_iters = warmup_iters
         self.warmup_ratio = warmup_ratio
 
-        self.base_momentum = []  # initial momentum for all param groups
-        self.regular_momentum = [
+        self.base_momentum: Union[list, dict] = [
+        ]  # initial momentum for all param groups
+        self.regular_momentum: Union[list, dict] = [
         ]  # expected momentum if no warming up is performed
```
zhouzaida, line 36:
```suggestion
        # initial momentum for all param groups
        self.base_momentum: Union[list, dict] = []
        # expected momentum if no warming up is performed
        self.regular_momentum: Union[list, dict] = []
```

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -87,6 +94,10 @@ def _get_warmup_momentum(cur_iters, regular_momentum):
                 warmup_momentum = [
                     _momentum / k for _momentum in regular_momentum
                 ]
+            else:
+                raise ValueError(
+                    f"Invalid 'self.warmup' value of {self.warmup}, "
+                    'report this as a bug')
```
zhouzaida, line 100:
```suggestion
                    'Expected values of `self.warmup` can be "constant", "linear" or "exp", '
                    f'but got {self.warmup}')
```

mmcv/runner/hooks/momentum_updater.py
```diff
@@ -404,11 +429,13 @@ class OneCycleMomentumUpdaterHook(MomentumUpdaterHook):
...
-                 pct_start=0.3,
-                 anneal_strategy='cos',
-                 three_phase=False,
+                 base_momentum: Union[float, List[float], Dict[str,
+                                                               float]] = 0.85,
+                 max_momentum: Union[float, List[float], Dict[str,
+                                                              float]] = 0.95,
+                 pct_start: float = 0.3,
+                 anneal_strategy: str = 'cos',
+                 three_phase: bool = False,
```
zhouzaida, line 438:
```python
def __init__(
    self,
    base_momentum: Union[float, List[float], Dict[str, float]] = 0.85,
    max_momentum: Union[float, List[float], Dict[str, float]] = 0.95,
    pct_start: float = 0.3,
    anneal_strategy: str = 'cos',
    three_phase: bool = False,
    **kwargs,
)
```

triwahyuu: this one is actually from yapf, so I can't change it, if I do it will be formatted back like that

zhouzaida: Yes, but it is not enough readable when type hints were broken into two lines.

teamwong111: We can use `# yapf: disable` to avoid formatting automatically.

triwahyuu: I think this would be better to make the type hint less verbose, i.e. `Union[float, list, dict]`, to make it shorter rather than use the workaround, what do you think?

triwahyuu: I have tried it and it passed

zhouzaida (approved):

zhouzaida (approved):

PR #2151, opened 2022-07-28
🎨 Delay the YAPF import until it's actually needed.

coderanger:
## Motivation

This avoids a hard dependency on lib2to3 which is not included in most minimal Python distributions such as in a container environment.

## Modification

The import is moved inside the function so yapf is only needed if the pretty-print functionality is actually in use.

## BC-breaking (Optional)

No compat effect.

## Use cases (Optional)

N/A other than making it easier to use MMCV in minimal environment.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/utils/config.py
+++ b/mmcv/utils/config.py
@@ -16,7 +16,6 @@ from importlib import import_module
 from pathlib import Path
 
 from addict import Dict
-from yapf.yapflib.yapf_api import FormatCode
 
 from .misc import import_modules_from_strings
 from .path import check_file_exist
@@ -417,6 +416,7 @@ class Config:
 
     @property
     def pretty_text(self):
+        from yapf.yapflib.yapf_api import FormatCode
 
         indent = 4
 


HAOCHENYE (commented):
Since `FormatCode` is optional, maybe we can import it by `try... except` and throw a not formatted string warning.

coderanger:
> Since `FormatCode` is optional, maybe we can import it by `try... except` and throw a not formatted string warning.

Also an option, but having features that vary based on which libraries are installed can lead to unexpected deployment woes. If going that route, I would probably add a `format=True/False` parameter to the method so the user can request the desired behavior.

HAOCHENYE:
> > Since `FormatCode` is optional, maybe we can import it by `try... except` and throw a not formatted string warning.
> 
> Also an option, but having features that vary based on which libraries are installed can lead to unexpected deployment woes. If going that route, I would probably add a `format=True/False` parameter to the method so the user can request the desired behavior.

Sorry for my late reply,  If `format=True/False` is a parameter of the method, what's the default behavior of `pretty_text`, BTW, it is a `property` actually. 
In my opinion, the default behavior of `xxx.pretty_text` should not raise an `ImportError` whether `yapf` is installed or not. Consider the case user export the config by `cfg.pretty_text` at the end of training， the `ImportError` could break the saving process.

OpenMMLab-Assistant-004:
Hi @coderanger ！We are grateful for your efforts in helping improve mmcv open-source project during your personal time. 

PR #2153, opened 2022-07-28
fix a bug in `_convert_output_type_range`

wwhio:
## Motivation

Bug fix. 

Keep the data range when convert it using `_convert_input_type_range` then `_convert_output_type_range`.

And the calculation in `rgb2ycbcr` and `bgr2ycbcr` is based on data with range [0, 255].

## Modification

In function `_convert_input_type_range` :
    Keep the data with range [0, 255] when dtype=np.uint8.
    Enlarge the data with range [0, 1] when dtype=np.float32.

## BC-breaking (Optional)

No.

## Use cases (Optional)

Two function calls : `rgb2ycbcr` , `bgr2ycbcr` and their inverse functions.

## Checklist

**Before PR**:

- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness. (**need help**)
- [ ] The documentation has been modified accordingly, including docstring or example tutorials. (**need help**. only docstring is modified.)

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls. (**need help**. MMEdit as far as I know)

### DIFF
--- a/mmcv/image/colorspace.py
+++ b/mmcv/image/colorspace.py
@@ -97,14 +97,14 @@ def _convert_input_type_range(img: np.ndarray) -> np.ndarray:
 
     Returns:
         (ndarray): The converted image with type of np.float32 and range of
-            [0, 1].
+            [0, 255].
     """
     img_type = img.dtype
     img = img.astype(np.float32)
     if img_type == np.float32:
-        pass
+        img *= 255.
     elif img_type == np.uint8:
-        img /= 255.
+        pass
     else:
         raise TypeError('The img type should be np.float32 or np.uint8, '
                         f'but got {img_type}')


wwhio:
I need some help to fix the inverse function of `rgb2ycbcr` and `bgr2ycbcr`.

It seems they have already work with data range [0, 255], so the fix should be remove `* 255` in the line `img = _convert_input_type_range(img) * 255` (line [236](https://github.com/wwhio/mmcv/blob/cec8e44445a5c2fe73d9668ef4af6ae66699e9ad/mmcv/image/colorspace.py#L236) and line [268](https://github.com/wwhio/mmcv/blob/cec8e44445a5c2fe73d9668ef4af6ae66699e9ad/mmcv/image/colorspace.py#L268)) ?

HAOCHENYE:
- `_convert_input_type_range` will convert data range to 0~1
-  `_convert_output_type_range` will convert data to its original range, float: 0\~1 and uint8: 0\~255, which is a standard range of opencv c++ api.

The formula of xxx2ycbcr assumes that the range of img is 0\~1, so we need to call `_convert_input_type_range` to convert input data to 0~1.

The formula of ycbcr2xxx assumes that the range of img is 0\~255, so we need to additionally call `*255` to convert input data to 0~255.

The convert function can be found in https://en.wikipedia.org/wiki/YCbCr#JPEG_conversion.

wwhio:
@HAOCHENYE 

Thank you for your reply.

I think the problem is that the value used in xxx2ycbcr assumes the range of image is 0~255.

So, if I understand it correctly, what I should do next are:
1. Revert my current changes.
2. Fix xxx2ycbcr by change the values.

The conversion is according to the digital Y′CbCr (8 bits per sample) in https://en.wikipedia.org/wiki/YCbCr#ITU-R_BT.601_conversion.

wwhio:
Oh, i get it. Thank you. @HAOCHENYE

PR #2200, opened 2022-08-16
[Fix] Fix a potential bug in PrRoIPool op

JingweiZhang12:
Fix a potential bug in PrRoIPool op

## Motivation
Due to the limitation of memory in https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/csrc/common/cuda/common_cuda_helper.hpp#L24,  we should prevent memory overflow of pointers in PrRoIPool op when there are many roi bboxes.

## Modification

Fix a potential bug about memory overflow in PrRoIPool op

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


grimoire (approved):
LGTM

PR #2450, opened 2022-11-23
[Fix] fix assertion at RandomChoiceResize

xiexinch:

## Motivation

Fix the problem of assertion and arguments don't match 

## Modification

- mmcv/transforms/processing.py

## BC-breaking (Optional)

## Use cases (Optional)

### DIFF
--- a/mmcv/transforms/processing.py
+++ b/mmcv/transforms/processing.py
@@ -1063,7 +1063,7 @@ class RandomChoiceResize(BaseTransform):
             self.scales = scales
         else:
             self.scales = [scales]
-        assert mmengine.is_list_of(self.scales, tuple)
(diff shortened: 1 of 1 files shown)


zhouzaida (approved):

mmcv/transforms/processing.py
```diff
@@ -1063,7 +1063,7 @@ def __init__(
             self.scales = scales
         else:
             self.scales = [scales]
-        assert mmengine.is_list_of(self.scales, tuple)
+        assert mmengine.is_list_of(self.scales, (tuple, int))
```
HAOCHENYE, line 1066: It could be better to use `is_seq_of`

PR #2468, opened 2022-12-04
[Refactor] Use the new interface of fileio from mmengine

zhouzaida:

## Motivation

Use the new interface of fileio from MMEngine which requires mmengine>=0.2.0.

## Modification

- Update `mmengine>=0.2.0` in requirements/runtime.txt
- Add a new parameter `backend_args` and remove `file_client_args`

## BC-breaking (Optional)

Yes. This PR removed the `file_client_args` parameter in several functions or classes.

## Use cases (Optional)

Before this PR

```python
img = mmcv.imread(s3_img_path, file_client_args={'backend': 'http'})
```

After this PR

```python
img = mmcv.imread(http_img_path, backend_args={'backend': 'http'})
```

The only difference between `file_client_args` and `backend_args` is that the former may contain a prefix field, because in the previous fileio (mmengine < 0.2.0), there were multiple file backends with the same prefix, so you need to specify the prefix in `file_client_args`.

## Checklist

**Before PR**:

**After PR**:


mmcv/transforms/loading.py
```diff
@@ -35,23 +36,50 @@ class LoadImageFromFile(BaseTransform):
...
+                '"file_client_args" will be deprecated in future. '
+                'Please use "backend_args" instead', DeprecationWarning)
+            if backend_args is not None:
+                raise ValueError(
+                    '"file_client_args" and "backend_args" cannot be set '
+                    'at the same time.')
+        else:
+            file_client_args = dict(backend='disk')
+        if backend_args is None:
+            backend_args = dict(backend='disk')
```
mzr1996, line 72: Can we set these arguments to None by default?

zhouzaida: OK

mmcv/transforms/loading.py
```diff
@@ -33,25 +33,24 @@ class LoadImageFromFile(BaseTransform):
...
 
     def __init__(self,
                  to_float32: bool = False,
                  color_type: str = 'color',
                  imdecode_backend: str = 'cv2',
-                 file_client_args: dict = dict(backend='disk'),
-                 ignore_empty: bool = False) -> None:
+                 ignore_empty: bool = False,
+                 *,
+                 backend_args: Optional[dict] = None) -> None:
```
HAOCHENYE, line 48:
```suggestion
                 ignore_empty: bool = False,
                 *,
                 backend_args: Optional[dict] = None):
```

HAOCHENYE (approved):

PR #2056, opened 2022-06-15
Support receptive field search of CNN models. (TPAMI paper rf-next)

gasvn:
## Motivation

Merging a general receptive field search method to mmcv and mmdet.
(paper: RF-Next: Efficient Receptive Field Search for Convolutional Neural Networks TPAMI 2022 [pdf](http://mftp.mmcheng.net/Papers/22TPAMI-ActionSeg.pdf))

## Modification

The RFSearch module is included.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

This pr is a part of pr in mmdet https://github.com/open-mmlab/mmdetection/pull/8191

## Checklist

**Before PR**:

- [x] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to create this PR.
- [x] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls.
- [x] CLA has been signed and all committers have signed the CLA in this PR.

PR #2489, opened 2022-12-15
Fix skip_layer for RF-Next

lzyhha:
## Motivation

RF-Next(TPAMI2022) has been merged in [PR](https://github.com/open-mmlab/mmcv/pull/2056).
But when I apply RF-Next to `mmsegmentation`, I found that some codes should be revised. 
Current code will skip a layer when the `config.search.skip_layer` includes one of the subnames of this layer, 
rather than the full name. 

For example,  if I want to skip a layer named `layer.0.conv` during searching, I should add `layer` / `0` / `conv`  to `config.search.skip_layer`. However, this will cause other layers to be incorrectly skipped, e.g., `layer.1.conv`.
So the codes should be revised to to judge whether a layer should be skipped according to its full name.

## Modification

Determine whether a layer should be skipped according to the full_name.

`
if any(layer in fullname for layer in self.config['search']['skip_layer']): continue
`

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/cnn/rfsearch/search.py
+++ b/mmcv/cnn/rfsearch/search.py
@@ -163,6 +163,10 @@ class RFSearchHook(Hook):
                 fullname = 'module.' + name
             else:
                 fullname = prefix + '.' + name
+            if self.config['search']['skip_layer'] is not None:
+                if any(layer in fullname
+                       for layer in self.config['search']['skip_layer']):
+                    continue
             if isinstance(module, eval(op)):
                 if 1 < module.kernel_size[0] and \
                     0 != module.kernel_size[0] % 2 or \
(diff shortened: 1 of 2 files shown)


zhouzaida:
Hi @lzyhha , thanks for your contribution. Could you update unit tests for this modification?

lzyhha:
OK, I will add the corresponding tests ASAP. 
After this PR is merge, I will create a new PR in mmsegmentation.

lzyhha:
I have update the unit tests. But there are some problems with pip.

zhouzaida:
> I have update the unit tests. But there are some problems with pip.

This error can be ignored.

lzyhha:
The new PR in [mmseg](https://github.com/open-mmlab/mmsegmentation/pull/2439).

zhouzaida (approved):

PR #2530, opened 2023-01-07
[Fix] Fix CI failed caused by 'Python.h' not found

ice-tong:
## Motivation
The CI job failed: https://github.com/open-mmlab/mmcv/actions/runs/3826436209/jobs/6566398257

The mmcv-full compile failed because the Python C-API header include directory is wrong:
image
In the CI container:
image

This is a bug introduce by setuptools [v65.2.0](https://github.com/pypa/setuptools/pull/3258), and the verison of setuptools for python 3.10.9 in setup-python action is 65.5.0 (while 63.2.0 for python3.10.8).

Related links:
- https://github.com/pypa/setuptools/issues/3589
- https://github.com/pypa/distutils/issues/178

## Modification
A workaround is use `setuptools<=65.1.0` when python==3.10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -277,6 +277,10 @@ jobs:
       # pstuil is an optional package to detect the number of CPU for compiling mmcv
       - name: Install psutil
         run: python -m pip install psutil
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        with:
+          sudo: true
       - name: Build and install
         run: rm -rf .eggs && python -m pip install -e .
       - name: Validate the installation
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -277,6 +277,10 @@ jobs:
       # pstuil is an optional package to detect the number of CPU for compiling mmcv
       - name: Install psutil
         run: python -m pip install psutil
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        with:
+          sudo: true
```
ice-tong, line 283:
```suggestion
          sudo: false
```

zhouzaida (approved):

PR #2536, opened 2023-01-09
CodeCamp #15

NoneLand:
## Motivation

This PR adds sigmoid focal loss CPU implementation on the basis of CUDA implementation.

## Modification

* Add sigmoid focal loss cpu implementation

* Add focal loss unit test on cpu

* Add ops to EN/ZH documents

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
+++ b/mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
@@ -1,0 +1,104 @@
+// Copyright(c) OpenMMLab.All rights reserved.
+#include <cfloat>
+#include <cmath>
+#include "pytorch_cpp_helper.hpp"
+#include "pytorch_device_registry.hpp"
+
+template <typename T>
+void sigmoid_focal_loss_forward_cpu_kernel(const int N, const T* input, const int64_t* target,
+                               const T* weight, T* output,
+                               const float gamma,
+                               const float alpha,
+                               const int num_classes) {
+    for (int i = 0; i < N; i++) {
+        T p = (T) 1. / ((T)1. + expf(-input[i]));
+        int64_t t = target[i / num_classes];
+        if (t == (i % num_classes)) { // positive
+            output[i] = -alpha * pow(((T)1. - p), gamma) * log(std::max(p, (T)FLT_MIN));
+        }
+        else {
+            output[i] = -((T)1. - alpha) * pow(p, gamma) * log(std::max((T)1. - p, (T)FLT_MIN));
+        }
+        if (weight != NULL) {
+            output[i] *= weight[t];
+        }
+    }
+}
+
+template <typename T>
+void sigmoid_focal_loss_backward_cpu_kernel(const int N, const T* input, const int64_t* target,
+                               const T* weight, T* grad_input,
+                               const float gamma,
+                               const float alpha,
+                               const int num_classes) {
+    for (int i = 0; i < N; i++) {
+        T p = (T) 1. / ((T)1. + expf(-input[i]));
+        int64_t t = target[i / num_classes];
+        if (t == (i % num_classes)) { // positive
+            grad_input[i] = -alpha * pow((T)1. - p, gamma) *
+               ((T)1. - p - (gamma * p * log(std::max(p, (T)FLT_MIN))));
+        }
+        else {
+            grad_input[i] = -((T)1. - alpha) * pow(p, gamma) *
+               (gamma * ((T)1. - p) * log(std::max((T)1. - p, (T)FLT_MIN)) - p);
+        }
+        if (weight != NULL) {
+            grad_input[i] *= weight[t];
+        }
+    }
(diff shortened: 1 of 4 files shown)


mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+void sigmoid_focal_loss_forward_impl(Tensor input, Tensor target,
+                               Tensor weight, Tensor output,
+                               const float gamma,
+                               const float alpha);
+void sigmoid_focal_loss_backward_impl(Tensor input, Tensor target,
+                                      Tensor weight, Tensor grad_input,
+                                      float gamma, float alpha);
+
+REGISTER_DEVICE_IMPL(sigmoid_focal_loss_forward_impl, CPU, sigmoid_focal_loss_forward_cpu);
```
grimoire, line 103: As the input type are the same between `TensorSigmoidFocalLossForwardCPUKernelLaucher` and `sigmoid_focal_loss_forward_cpu`, we can remote `sigmoid_focal_loss_forward_cpu` and register `TensorSigmoidFocalLossForwardCPUKernelLaucher`.
Same for `TensorSigmoidFocalLossBackwardCPUKernelLaucher`

mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "pytorch_device_registry.hpp"
+
+template <typename T>
+void sigmoid_focal_loss_forward_cpu_kernel(const int N, const T* input, const int64_t* target,
+                               const T* weight, T* output,
+                               const float gamma,
+                               const float alpha,
+                               const int num_classes) {
+    for (int i = 0; i < N; i++) {
+        T p = (T) 1. / ((T)1. + expf(-input[i]));
```
grimoire, line 14: expf can be replaced with exp since c++ has overloaded it.

grimoire (commented):
Since this op can be implemented with native torch. Please provide a benchmark between this ops and torch implementation.

mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
NoneLand: Thanks for your review opinion. I would fix this because it saves code.

mmcv/ops/csrc/pytorch/cpu/sigmoid_focal_loss.cpp
NoneLand: OK. I would fix this in next commit.

NoneLand:
Thanks for reviewer's opinion. I have some question about the benchmark. Could you please help me figure out?

1. Is there a benchmark example for this op? I think this answer would be YES due to the existing cuda op.
2. I would like to know the goal for mmcv to reimplement this op? I would guess the goal is performance due to the benchmark requirement. If the benchmark result shows current op impl has no superior performance than torch impl, would merge delay until the optimization is completed or would merge before optimization?(HAHA, codecamp has a ddl^_^)
3. If there was no benchmark before, here is my plan, could you please help me check? I would test for different batch (2, 4, 8, ..., 8192) and different num_classes (2, 4, 8, ..., 8192) for the python api, each case run 10 times, and cal the average latency of both impl.

Thanks again^_^

grimoire:
1. Yes, it would help us to find the bottleneck of the ops.
2. It would be cool if the benchmark can guide you to improve the performance of the op. And we will merge this PR even when the performance is not the best (less is better than nothing). Your benchmark would help us to optimize it in the future.
3. The plan sounds cool. It would be even better if you can perform some tests on the data size of the real scene. [pytest-benchmark](https://pypi.org/project/pytest-benchmark/) or [PyTorch profiler](https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html) can help you build the benchmark. You can place the benchmark in this PR.

NoneLand:
I have used pytest-benchmark to benchmark the implemented sigmoid focal loss op.  Benchmark code is in the latest commit. And on my 8 core notebook, conclusions could be drawn as below: 1. In small data size, implemented op is superior. 2. In medium and large data size, implemented op is superior.  The processed benchmark data could be seen in [benchmark_round10.csv](https://github.com/open-mmlab/mmcv/files/10426594/benchmark_round10.csv) and the raw benchmark data is in 
[0003_4e4fa93e7c5ee5a15dd63068c0fe9992068bc881_20230116_012432_uncommited-changes.txt](https://github.com/open-mmlab/mmcv/files/10426612/0003_4e4fa93e7c5ee5a15dd63068c0fe9992068bc881_20230116_012432_uncommited-changes.txt)

grimoire:
Cool.
Do remember to remove the benchmark test code before the final commit. Benchmark test might slow down github action.

PR #2556, opened 2023-01-16
[Fix] Check the version of torchvision in __init__ of DCN

mengpenghui:
## Motivation

When performing the MLU-based PyTorch1.6 test, because PyTorch1.6-mlu does not support DCN,
 the local test case will report an error.

## Modification

mmcv/ops/deform_conv.py： 
-Put the assert of torchvision version in the function
-Simplified weight syntax
mmcv/ops/modulated_deform_conv.py：
-Put the assert of torchvision version in the function

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/deform_conv.py
+++ b/mmcv/ops/deform_conv.py
@@ -438,12 +438,9 @@ class DeformConv2dPack(DeformConv2d):
 
 if IS_MLU_AVAILABLE:
     import torchvision
+    from torchvision.ops import deform_conv2d as tv_deform_conv2d
 
     from mmcv.utils import digit_version
-    assert digit_version(torchvision.__version__) >= digit_version(
-        '0.10.0a0'), 'the version of torchvision should be >= 0.10.0'
-
-    from torchvision.ops import deform_conv2d as tv_deform_conv2d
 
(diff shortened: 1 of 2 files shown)


mmcv/ops/deform_conv.py
```diff
@@ -489,12 +486,13 @@ def init_offset(self):
             self.conv_offset.bias.data.zero_()
 
         def forward(self, x: Tensor) -> Tensor:  # type: ignore
+            assert digit_version(torchvision.__version__) >= digit_version(
+                '0.10.0a0'), 'the version of torchvision should be >= 0.10.0'
```
zhouzaida, line 490: suggest moving it to `__init__`.

mmcv/ops/modulated_deform_conv.py
```diff
@@ -402,6 +400,8 @@ def init_weights(self):
                 self.conv_offset.bias.data.zero_()
 
         def forward(self, x):
+            assert digit_version(torchvision.__version__) >= digit_version(
+                '0.10.0a0'), 'the version of torchvision should be >= 0.10.0'
```
zhouzaida, line 404: same as above

zhouzaida (approved):

PR #2644, opened 2023-03-07
[Fix] Prevent divide-by-zero error on Ascend device for bbox_overlaps

dflhw:
Change the value of offset to adapt to the Ascend device.


ckirchhoff2021 (approved):
/LGTM

PR #2646, opened 2023-03-07
[Fix] Prevent divide-by-zero error on Ascend device for bbox_overlaps

dflhw:
Change the value of offset to adapt to the Ascend device.

### DIFF
--- a/mmcv/ops/csrc/pytorch/npu/bbox_overlaps_npu.cpp
+++ b/mmcv/ops/csrc/pytorch/npu/bbox_overlaps_npu.cpp
@@ -12,6 +12,10 @@ void bbox_overlaps_npu(const Tensor bboxes1, const Tensor bboxes2, Tensor ious,
   if (mode == 1) {
     modeStr = "iof";
   }
+  float offset_ = 1;
+  if (offset == 0) {
+    offset_ = 0.01;
+  }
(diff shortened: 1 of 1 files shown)


mmcv/ops/csrc/pytorch/npu/bbox_overlaps_npu.cpp
```diff
@@ -22,7 +26,7 @@ void bbox_overlaps_npu(const Tensor bboxes1, const Tensor bboxes2, Tensor ious,
       .Input(gtboxes)
       .Output(ious)
       .Attr("mode", modeStr)
-      .Attr("eps", (float)offset)
+      .Attr("eps", offset_)
```
zhouzaida, line 29: Hi, why is `eps (offset)` used as the denominator?

zhouzaida: Why is it set as `0.01` rather than other value?

dflhw: This is due to the limitation of the development mode and model scenario. 0.01 and 1 need to be enumerated.

grimoire (approved):
LGTM

PR #2665, opened 2023-03-15
[Feature] Add support for RANS encoding & decoding

lizhihao6:

## Motivation

**Description:**

RANS (Range ANd Symbol) encoding is a powerful compression technique that has been shown to outperform other state-of-the-art compression algorithms, such as gzip and Brotli, in terms of compression ratio and speed. It is particularly effective in compressing data with high entropy, such as natural language text, images, and videos.

As a widely used computer vision library, mmcv could greatly benefit from integrating RANS encoding into its compression and decompression modules. By adding support for RANS encoding, mmcv could provide users with a more efficient and flexible compression solution, especially for large-scale image and video datasets.

Some potential use cases of RANS encoding in mmcv include:

1. Compressing image and video data for storage and transmission

2. Preprocessing large-scale datasets by compressing them on-the-fly

3. Improving the efficiency and speed of data loading and augmentation

Given the benefits of RANS encoding, it would be great to see mmcv integrate this technology into its compression and decompression modules.

**References:**

1. [High-Efficiency Lossy Image Coding Through Adaptive Neighborhood Information Aggregation](https://arxiv.org/abs/2204.11448)

2. [Understanding the ANS Compressor](https://kedartatwawadi.github.io/post--ANS/)

## Modification

We have made major modifications to the following files to support RANS encoding and decoding: 

> mmcv/ops/csrc/common/utils/rans/rans64.h
mmcv/ops/csrc/common/utils/rans/rans_interface.hpp
mmcv/ops/csrc/pytorch/cpu/rans.cpp
mmcv/ops/rans.py

It is worth noting that we introduced class-level pybind in mmcv/ops/csrc/pytorch/pybind.cpp for the first time, as shown below:

```python
py::class_<BufferedRansEncoder>(m, "BufferedRansEncoder")
    .def(py::init<>())
    .def("encode_with_indexes", &BufferedRansEncoder::encode_with_indexes)
    .def("flush", &BufferedRansEncoder::flush);

py::class_<RansEncoder>(m, "RansEncoder")
    .def(py::init<>())
    .def("encode_with_indexes", &RansEncoder::encode_with_indexes);

py::class_<RansDecoder>(m, "RansDecoder")
    .def(py::init<>())
    .def("set_stream", &RansDecoder::set_stream)
    .def("decode_stream", &RansDecoder::decode_stream)
    .def("decode_with_indexes", &RansDecoder::decode_with_indexes,
         "Decode a string to a list of symbols");
```
**!!!This is significantly different from the previous bindings that only bound functions.**

## BC-breaking (Optional)

No.

## Use cases (Optional)

**BufferedRansEncoder**

The BufferedRansEncoder class is a buffered RANS encoder. It allows encoding a sequence of integers with different probability distributions specified by their cumulative distribution functions (CDFs) using a RANS codec. The encoder buffers the output bitstream in memory until it is flushed.

To use the BufferedRansEncoder class in Python, first import it from the module:

```python
from mmcv.ops import BufferedRansEncoder
encoder = BufferedRansEncoder()
```

The encoder has a single public method, `encode_with_indexes()`, which encodes a sequence of integers using the specified CDFs:

```python
symbols = np.random.randint(0, 256, size=1000).tolist()
indexes = np.random.randint(0, 256, size=1000).tolist()
cdfs = [[0] * 257 for _ in range(100)]
for i in range(100):
    for j in range(257):
        cdfs[i][j] = np.random.randint(0, 10000)
cdfs_sizes = np.random.randint(0, 100, size=100).tolist()
offsets = np.random.randint(0, 10000, size=100).tolist()
encoder.encode_with_indexes(symbols, indexes, cdfs, cdfs_sizes, offsets)
```

The `symbols` argument is a list of integers to be encoded. The `indexes` argument is a list of indices into the `cdfs` list specifying which CDF to use for each symbol. The `cdfs` argument is a list of lists of integers specifying the CDFs. The `cdfs_sizes` argument is a list of integers specifying the number of elements in each CDF. The `offsets` argument is a list of integers specifying the starting index of each CDF in the flattened `cdfs` list.

Finally, to flush the encoder and get the encoded bitstream as bytes, call the `flush()` method:

```python
encoded_bytes = encoder.flush()
```

**RansEncoder**

The RansEncoder class is a non-buffered RANS encoder. Using a RANS codec, it allows encoding a sequence of integers with different probability distributions specified by their cumulative distribution functions (CDFs). Unlike the BufferedRansEncoder class, the RansEncoder class encodes the output bitstream immediately.

To use the RansEncoder class in Python, first import it from the module:

```python
from mmcv.ops import RansEncoder
encoder = RansEncoder()
```

The encoder has a single public method, `encode_with_indexes()`, which encodes a sequence of integers using the specified CDFs and returns the encoded bitstream as bytes:

```python
encoded_bytes = encoder.encode_with_indexes(symbols, indexes, cdfs, cdfs_sizes, offsets)
```

The arguments are the same as for the `encode_with_indexes()` method of the `BufferedRansEncoder` class.

**RansDecoder**

The RansDecoder class is a RANS decoder. It allows decoding a sequence of integers encoded using a RANS codec with different probability distributions specified by their cumulative distribution functions (CDFs).

To use the RansDecoder class in Python, first import it from the module:

```python
from mmcv.ops import RansDecoder
decoder = RansDecoder()
```

The decoder has two public methods. The first method, `decode_with_indexes()`, decodes the specified encoded bitstream and returns the decoded sequence of integers:

```python
decoded = decoder.decode_with_indexes(encoded_bytes, indexes, cdfs, cdfs_sizes, offsets)
```

The `encoded_bytes` argument is the encoded bitstream as bytes. The other arguments are the same as for the `encode_with_indexes()` method of the `BufferedRansEncoder` class.

The second method, `decode_stream()`, decodes the bitstream specified by `set_stream()` and returns the decoded sequence of integers:

```python
decoder.set_stream(encoded_bytes)
decoded = decoder.decode_stream(indexes, cdfs, cdfs_sizes, offsets)
```

## Checklist

**Before PR**:

**After PR**:


mmcv/ops/rans.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ..utils import ext_loader
+
+ext_module = ext_loader.load_ext('_ext', [
+    'rans_encode_with_indexes', 'rans_decode_with_indexes',
+    'pmf_to_quantized_cdf'
+])
+
+
+def rans_encode_with_indexes(symbols: Tensor,
```
grimoire, line 12: Please add docstring and wrap `rans_decode_with_indexes` and `pmf_to_quantized_cdf` into functions.

tests/test_ops/test_rans.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _move_data_to_target_device(data, device):
+    if isinstance(data, torch.Tensor):
+        return data.to(device)
+    return [d.to(device) for d in data]
+
+
+def _calculate_mibs(fn, device, num_threads, num_symbols, args):
+    if torch.cuda.is_available():
```
grimoire, line 56: use torch.cuda.event to do the cuda timing

tests/test_ops/test_rans.py
```diff
@@ -0,0 +1,123 @@
...
+@pytest.mark.parametrize('device', DEVICES)
+@pytest.mark.parametrize('num_threads', [1, 2, 4])
+def test_rans_encode_with_indexes(rans_input_data, rans_output_data, device,
+                                  num_threads):
+    data = _move_data_to_target_device(rans_input_data, device)
+    symbols, indexes, cdfs, cdfs_sizes, offsets = data
+    gt_encoded = rans_output_data
+    encoded = rans_encode_with_indexes(symbols, indexes, cdfs, cdfs_sizes,
+                                       offsets, num_threads)
+    print(100 * (len(encoded) - 49886) / 49886)
```
grimoire, line 79: remove unused print.

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
```diff
@@ -0,0 +1,206 @@
...
+      raw_val = -2 * value - 1;
+      value = max_value;
+    } else if (value >= max_value) {
+      raw_val = 2 * (value - max_value);
+      value = max_value;
+    }
+
+    assert(value >= 0 && value < max_value + 1);
+
+    rans_symbols[i].start = cdf[value];
```
grimoire, line 58: load and store from global memory is heavy. Please cache the value in variable, do the compute and write it to the global memory.

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
```diff
@@ -0,0 +1,206 @@
...
+    rans_symbols[i].range = cdf[value + 1] - cdf[value];
+    assert(rans_symbols[i].range != 0);
+    results[i].raw_value = raw_val;
+    results[i].bypass_mode = (value == max_value);
+  }
+}
+
+// vector would be optimized by compiler more than 10x faster than Tensor
+void rans_encode_cpu_kernel(const std::vector<RansSymbol>& symbols,
```
grimoire, line 67: Do not place non-kernel code in .cuh

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
```diff
@@ -0,0 +1,206 @@
...
+  } else {
+    shared_cdf[j] = round(pmf[j - 1] * (1 << precision));
+  }
+
+  __shared__ uint32_t total;
+  if (j == 0) {
+    total = 0;
+  }
+  __syncthreads();
+  atomicAdd(&total, shared_cdf[j]);
```
grimoire, line 148: atomicAdd should be avoid. you can create another smem cache with size THREADS_PER_BLOCK and do blockwise reduce. [like this](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf)

grimoire, line 148: atomic add is heavy. use [block reduce](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf) here.

grimoire: you might need another cache with size THREADS_PER_BLOCK/2 to avoid pollute shared_cdf

tests/test_ops/test_rans.py
```diff
@@ -0,0 +1,123 @@
...
+        torch.cuda.synchronize()
+    end_time = time.time()
+    latency = (end_time - start_time) / 100
+    mibs = num_symbols / latency / 1024 / 1024
+    print(f'\ndevice {device}, num_threads {num_threads}: \
+            {fn.__name__}: {latency:.4f}s, {mibs:.2f}MiB/s')
+
+
+@pytest.mark.parametrize('device', DEVICES)
+@pytest.mark.parametrize('num_threads', [1, 2, 4])
```
grimoire, line 71: Test failed with num_threads > 8 on my device. Is there any limit on the device? can we detect it and set the num_threads according to the environment?

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
```diff
@@ -0,0 +1,206 @@
...
+  __syncthreads();
+  assert(total != 0);
+
+  shared_cdf[j] =
+      (static_cast<uint64_t>(1 << precision) * ((uint32_t)shared_cdf[j])) /
+      total;
+  __syncthreads();
```
grimoire, line 155: Since every thread read it's own shared_cdf. There is no need to add syncthreads here.

mmcv/ops/rans.py
lizhihao6: I added the docstrings and wrap functions for these functions.
```suggestion
// Copyright(c) OpenMMLab.All rights reserved.
#include <pybind11/pybind11.h>

#include "pytorch_cpp_helper.hpp"
#include "pytorch_device_registry.hpp"

namespace py = pybind11;
std::string rans_encode_with_indexes_impl(
    const Tensor symbols, const Tensor indexes, const Tensor cdfs,
    const Tensor cdfs_sizes, const Tensor offsets, int num_threads) {
  return DISPATCH_DEVICE_IMPL(rans_encode_with_indexes_impl, symbols, indexes,
                              cdfs, cdfs_sizes, offsets, num_threads);
}

py::bytes rans_encode_with_indexes(const Tensor symbols, const Tensor indexes,
                                   const Tensor cdfs, const Tensor cdfs_sizes,
                                   const Tensor offsets, int num_threads) {
  return rans_encode_with_indexes_impl(symbols, indexes, cdfs, cdfs_sizes,
                                       offsets, num_threads);
}

Tensor rans_decode_with_indexes_impl(const std::string& encoded,
                                     const Tensor indexes, const Tensor cdfs,
                                     const Tensor cdfs_sizes,
                                     const Tensor offsets) {
  return DISPATCH_DEVICE_IMPL(rans_decode_with_indexes_impl, encoded, indexes,
                              cdfs, cdfs_sizes, offsets);
}

Tensor rans_decode_with_indexes(const std::string& encoded,
                                const Tensor indexes, const Tensor cdfs,
                                const Tensor cdfs_sizes, const Tensor offsets) {
  return rans_decode_with_indexes_impl(encoded, indexes, cdfs, cdfs_sizes,
                                       offsets);
}

Tensor pmf_to_quantized_cdf_impl(const Tensor pmfs, const Tensor pmf_lengths,
                                 const Tensor tail_masses) {
  return DISPATCH_DEVICE_IMPL(pmf_to_quantized_cdf_impl, pmfs, pmf_lengths,
                              tail_masses);
}

Tensor pmf_to_quantized_cdf(const Tensor pmfs, const Tensor pmf_lengths,
                            const Tensor tail_masses) {
  return pmf_to_quantized_cdf_impl(pmfs, pmf_lengths, tail_masses);
}

```

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
```diff
@@ -0,0 +1,206 @@
...
+
+  if (j == 0) {
```
grimoire, line 157: could use [block scan](https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda)

tests/test_ops/test_rans.py
lizhihao6: We add max cpu threds limitation.
```python
NUM_THREDS = [i for i in [1, 2, 4] if i <= cpu_count()]
```

mmcv/ops/csrc/pytorch/cuda/rans.cu
```diff
@@ -0,0 +1,204 @@
...
+  const int max_cdf_bin_size = cdfs.size(1);
+
+  // set kernel params
+  int cuda_threads = min(num_symbols, THREADS_PER_BLOCK);
+  dim3 blocks((num_symbols + cuda_threads - 1) / cuda_threads);
+  dim3 threads(cuda_threads);
+
+  // malloc device memory for cache results
+  RansSymbol *rans_symbols;  // store encoded symbols
+  CUDA_MALLOC(rans_symbols, num_symbols);
```
grimoire, line 36: Do not use cudamalloc here. you can create a int8 tensor with the size you want. reinterprate the data_ptr into the RansSymbol

grimoire: If you do so. you do not need synchronize the stream and move to cpu. just leave it to pytorch.

tests/test_ops/test_rans.py
lizhihao6: I change it into torch.cuda.event.
```suggestion
    if torch.cuda.is_available():
        start_event = torch.cuda.Event(enable_timing=True)
        end_event = torch.cuda.Event(enable_timing=True)
        start_event.record()
```

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
lizhihao6: I move this funcion into rans.cu

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
lizhihao6: in the following lines, thread  0 will be added to the value from all threads, so we need to make sure the previous command is done.
```cuh
  shared_cdf[j] =
      (static_cast<uint64_t>(1 << precision) * ((uint32_t)shared_cdf[j])) /
      total;
  __syncthreads();

  if (j == 0) {
    for (int k = 1; k < pmf_length + 1; k++) {
      shared_cdf[k] += shared_cdf[k - 1];
    }
    shared_cdf[pmf_length + 1] = 1 << precision;
  }
  __syncthreads();
```

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
lizhihao6: While this operation has some similarities to the order-free sum computation performed by block scan, I think it cannot be directly optimized using block scan because it is not order-free. Each `shared_cdf[k]` is dependent on the previous sum value. Furthermore, this function is used infrequently and is only used once during the network initialization.

mmcv/ops/csrc/pytorch/cuda/rans.cu
lizhihao6: I have changed it into torch tensor.
```suggestion
   auto rans_symbols_tensor =
      CREATE_TENSOR(num_symbols * sizeof(RansSymbol), torch::ScalarType::Byte,
                    symbols.device());  // store encoded symbols;
  RansSymbol *rans_symbols = (RansSymbol *)rans_symbols_tensor.data_ptr();
```

mmcv/ops/csrc/pytorch/cuda/rans.cu
```diff
@@ -0,0 +1,259 @@
...
+  int num_threads = header.num_threads;
+
+  // decode nbytes
+  uint32_t *nbytes = new uint32_t[num_threads];
+  int nbytes_bytes = num_threads * get_nbytes_size(header.nbytes_mode);
+  decompress_nbytes(
+      encoded.substr(encoded.length() - COMPRESS_HEADER_BYTES - nbytes_bytes),
+      nbytes, num_threads, header.nbytes_mode);
+
+  // multi-threading
```
grimoire, line 198: Please add guards and stream

lizhihao6: since the `stream` is not used in this function, could we only add the `guards`?
```python
  // set device
  at::cuda::CUDAGuard device_guard(symbols.device());
```

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
grimoire: Sure I see.

mmcv/ops/csrc/pytorch/cpu/rans.cpp
```diff
@@ -0,0 +1,185 @@
...
+  Tensor max_length_tensor = torch::max(pmf_lengths);
+  int cdf_max_length = max_length_tensor.item<int>() + 2;
+  Tensor quantized_cdfs = zeros({pmfs.size(0), cdf_max_length},
+                                TensorOptions().dtype(ScalarType::Int));
+  pmf_to_quantized_cdf_cpu_kernel(pmfs, pmf_lengths, tail_masses,
+                                  quantized_cdfs);
+  return quantized_cdfs;
+}
+
+// ------------------- BLIND RANS ENCODE -------------------
```
grimoire, line 147: I guess `BLIND` is a typo?

mmcv/ops/csrc/pytorch/cuda/rans.cu
grimoire: The guard leads to a compile error, `symbols` has not been defined in this function.

mmcv/ops/csrc/pytorch/cuda/rans.cu
```diff
@@ -0,0 +1,265 @@
...
+      symbols.device());  // store whether value == max_value;
+  RansCUDAKernelResult *results =
+      (RansCUDAKernelResult *)results_tensor.data_ptr();
+  // get encoded symbols
+  rans_encode_with_indexes_cuda_kernel<<<blocks, threads, 0, stream>>>(
+      symbols.data_ptr<int>(), indexes.data_ptr<int>(), cdfs.data_ptr<int>(),
+      cdfs_sizes.data_ptr<int>(), offsets.data_ptr<int>(), num_symbols,
+      num_cdfs, max_cdf_bin_size, rans_symbols, results);
+  // sync
+  cudaStreamSynchronize(stream);
```
grimoire, line 104: cudaStreamSynchronize is not necessary since tensor.cpu would automatically sync the stream.

mmcv/ops/csrc/pytorch/cuda/rans.cu
```diff
@@ -0,0 +1,265 @@
...
+
+  // move results to cpu vector
+  std::vector<RansSymbol> rans_symbols_cpu(num_symbols);
+  std::vector<RansCUDAKernelResult> results_cpu(num_symbols);
+  CUDA_TO_CPU(rans_symbols, rans_symbols_cpu.data(), num_symbols);
```
grimoire, line 109: rans_symbols_tensor.cpu() will do the memory copy + sync stream for you.

grimoire:
Since part of the cuda implementations are completed on cpu, Please add some `// TODO: ...` comment so we can optimize it in the future.

mmcv/ops/csrc/pytorch/cuda/rans.cu
lizhihao6: I have fixed this typo, and add stream into pmf_to_quantized_cdf_cuda_kernel.

mmcv/ops/csrc/pytorch/cpu/rans.cpp
lizhihao6: Fixed it into `BIND`

lizhihao6:
TODO has been added.
```cu
  // TODO: this part of code is done on cpu, maybe we can do it on gpu
```

mmcv/ops/csrc/common/cuda/rans_cuda_kernel.cuh
lizhihao6: Thanks, I will keep it now and add a //TODO
```cuh
  // TODO: this could be accelerated by using a block scan
```

grimoire (approved):
LGTM

tests/test_ops/test_rans.py
```diff
@@ -0,0 +1,129 @@
...
+
+import pytest
+import torch
+
+from mmcv.ops import (pmf_to_quantized_cdf, rans_decode_with_indexes,
+                      rans_encode_with_indexes)
+
+EXTEND_TIMES = 100
+DEVICES = ['cpu', 'cuda'] if torch.cuda.is_available() else ['cpu']
+NUM_THREDS = [i for i in [1, 2, 4] if i <= cpu_count()]
```
zhouzaida, line 14: https://github.com/open-mmlab/mmcv/blob/3269278ed0ef2f340bb8d5fc3334e85c08139e1d/setup.py#L187

We can call `cpu_affinity` of `psutil` to get the number of CPUs.

See https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count and https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_affinity for their difference.

zhouzaida:
Please also update docs/en/understand_mmcv/ops.md and docs/zh_cn/understand_mmcv/ops.md

PR #2697, opened 2023-03-24
Modify adaptation of bbox_overlaps op for ascend device

dflhw:
1. At present, compared with the bbox_overlaps operator, the iou operator on ascend device has precision error in the fp16 case and does not meet the acceptance standard, so go to the fp32 branch temporarily to avoid the problem.
2. Add support for offset at 0.
3. This change is independent of torch_npu.


ckirchhoff2021 (approved):
Add more notes to the modifications next time.

dflhw:
> Add more notes to the modifications next time.

ok.

PR #2765, opened 2023-04-17
Dkx/support diopi mlu

CokeDong:
support  mlu ops


CokeDong:
will replaced by :
https://github.com/open-mmlab/mmengine/pull/1127
and
https://github.com/open-mmlab/mmcv/pull/2790

PR #2775, opened 2023-04-22
Add cv2 graphics / headless extras to setup.py

Erotemic:
## Motivation

The mmcv library requires the cv2 module, but there are two flavors of the module that different users may want in different circumstances, either `opencv-python` or `opencv-python-headless`. These libraries are incompatible with each other, and if you incorrectly install the wrong one it becomes a pain to clean them up and install the right one. 

This PR introduces a way for the user to specify which one of these they want at install time. 

## Modification

I've added two new files in `requirements`. `cv2-headless.txt` and `cv2-graphics.txt` which specify reasonable minimum versions of opencv based on the version of Python the user is on.  These are registered as new `extras_require` options. So the new usage would be either:

```bash
pip install mmcv[cv2-headless]

OR

pip install mmcv[cv2-graphics]
```

This means you no longer have to install cv2 before / after mmcv, and you can be sure you have the right one for your use case based on the install command you use. I've been using this construct in [kwimage](https://gitlab.kitware.com/computer-vision/kwimage) and other cv repos for a while now with good success. 

This also makes a modification to the `parse_requirements` function in setup.py. I was the original author of this construct and I figured I would update it to the lastest version I'm using in my [cookiecutter repo](https://github.com/Erotemic/xcookie/blob/a80785354a2d8547ec4483c22aa7d19b86ce4c3b/setup.py#L67) while I was here. I don't think it matters to parse the new extra reqs, but it does support more cases that could exist in the future. That could be reverted if needbe. 
 
## Checklist

**Before PR**:

**After PR**:


Erotemic:
I'm not sure why the windows / osx dashboards failed. I don't think any change I've introduced here would cause that. I've rebased on a current main to see if updating helps.

Having this change would be helpful because currently pip complains about package conflicts if python-opencv-headless is installed instead of python-opencv. Here is a recent error message I've gotten:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
mmengine 0.7.4 requires opencv-python>=3, which is not installed.
mmcv 2.0.0 requires opencv-python>=3, which is not installed.
```

It might require this change be ported to mmengine as well.

PR #2779, opened 2023-04-25
Dkx/support diopi torch

CokeDong:
Dkx/support diopi torch


CokeDong:
will replaced by :
https://github.com/open-mmlab/mmengine/pull/1127
and
https://github.com/open-mmlab/mmcv/pull/2790

PR #2790, opened 2023-05-05
[Feature] Support mmcv ext with DIOPI impl

CokeDong:
Support mmcv ext+dipu


CokeDong:
depends on : https://github.com/open-mmlab/mmengine/pull/1127

setup.py
```diff
@@ -221,7 +223,25 @@ def get_extensions():
         except ImportError:
             pass
 
-        if is_rocm_pytorch or torch.cuda.is_available() or os.getenv(
+        if os.getenv('MMCV_WITH_DIOPI', '0') == '1':
+            print(f'Compiling {ext_name} with CPU and DIPU')
+            define_macros += [('MMCV_WITH_DIOPI', None)]
+            define_macros += [('DIOPI_ATTR_WEAK', None)]
```
grimoire, line 229: What is this macro for?

mmcv/ops/focal_loss.py
```diff
@@ -43,6 +43,11 @@ def forward(ctx,
 
         output = input.new_zeros(input.size())
 
+        if IS_DIPU_AVAILABLE:
```
grimoire, line 46: can we put the type cast inside .cpp? Python api better be device agnostic

grimoire:
how to build mmcv with dipu?

setup.py
CokeDong: for defining  ```__attribute__((weak))``` ref: https://github.com/DeepLink-org/DIOPI/blob/main/DIOPI-PROTO/include/diopi/diopirt.h#LL16C18-L16C40

CokeDong:
> how to build mmcv with dipu?

e.g. MMCV_WITH_DIOPI=1 MMCV_WITH_OPS=1 python setup.py build_ext -i

grimoire:
Is there any guide about how to setup `DIPU_ROOT`, `DIOPI_PATH` and `DIPU_PATH`?

setup.py
```diff
@@ -221,7 +223,25 @@ def get_extensions():
...
+            op_files = glob.glob('./mmcv/ops/csrc/pytorch/*.cpp') + \
+                glob.glob('./mmcv/ops/csrc/pytorch/cpu/*.cpp')
+            extension = CppExtension
+            include_dirs.append(os.path.abspath('./mmcv/ops/csrc/common'))
+            dipu_root = os.getenv('DIPU_ROOT')
```
grimoire, line 234: Is all of these env (`DIPU_ROOT`, `DIOPI_PATH`, `DIPU_PATH`, `VENDOR_INCLUDE_DIRS`) required? If some of it are optional, assign a default value would be cool.

CokeDong: yes, these are required. the meaning of its should alwayd corrsponds to ```dipu(torch_dipu)``` .which  is under rapid devlopment.(https://github.com/DeepLink-org/dipu_poc) . hopefully, these env para would be provided by ```dipu(torch_dipu)``` in the final. Regard ```dipu(torch_dipu)``` as one pytorch ext which will support a general device called ```dipu```.

grimoire: would you mind providing me the access to the repo?

CokeDong: invitaion sent

mmcv/ops/focal_loss.py
CokeDong: this arises from some bug in `dipu` and `diopi` which is fixed now. will delete these codes.

grimoire (approved):
LGTM

mmcv/utils/device_type.py
```diff
@@ -1,8 +1,10 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-from mmengine.device import (is_cuda_available, is_mlu_available,
-                             is_mps_available, is_npu_available)
+from mmengine.device import (is_cuda_available, is_dipu_available,
```
zhouzaida, line 2: This requires `mmengine>=0.7.4`.

setup.py
```diff
@@ -221,7 +223,25 @@ def get_extensions():
         except ImportError:
             pass
 
-        if is_rocm_pytorch or torch.cuda.is_available() or os.getenv(
+        if os.getenv('MMCV_WITH_DIOPI', '0') == '1':
```
zhouzaida, line 226: need to assert the version of `mmengine>=0.7.4`

setup.py
```diff
@@ -221,7 +223,33 @@ def get_extensions():
...
+            import mmengine  # NOQA: F401
+            major, minor, micro = mmengine.version_info
+            if major == 0 and (minor < 7 or (minor == 7 and micro < 4)):
+                raise RuntimeError(f'mmengine >= 0.7.4 is required but \
+                        {mmengine.__version__} is installed')
```
zhouzaida, line 231: We can use the function `digit_version` in
https://github.com/open-mmlab/mmengine/blob/a2e410bd48890c54f16c527d67c371ed8562facd/mmengine/utils/version_utils.py#L9

tests/test_ops/test_voxelization.py
```diff
@@ -140,6 +140,10 @@ def test_voxelization_nondeterministic():
 
 
 @pytest.mark.parametrize('device_type', [
+    pytest.param(
+        'cuda:0',
+        marks=pytest.mark.skipif(
+            not IS_CUDA_AVAILABLE, reason='requires CUDA support')),
```
zhouzaida, line 146: Why do we need to add this test which has been tested in https://github.com/open-mmlab/mmcv/pull/2790/files#diff-92c8d3007d5a6331f491492485d0ab27e088ed6088bef9b0e228ea0adfc7248cR22

CokeDong: this is for mlu self test, dipu will mock its to cuda device, so here add `device = 'cuda:0',`(so we can reuse testcase called test_voxelization_mlu). we will not used above testcase `test_voxelization` because this is only used for real cuda device.

zhouzaida: Need to add a comment for this

PR #2799, opened 2023-05-18
[Fix] Fix typehint in `_scale_size` and `rescale_size`

MambaWong:


mmcv/image/geometric.py
```diff
@@ -27,7 +27,7 @@ def _scale_size(
     Returns:
         tuple[int]: scaled size.
     """
-    if isinstance(scale, (float, int)):
+    if isinstance(scale, float):
```
zhouzaida, line 30: Hi @MambaWong , `scale` should support accepting `int` type so we only need to update the docstring.

MambaWong: I'm confused about the need for `int`, can you give me an example.

zhouzaida: Hi, this is just for the convenience of the user, for example, if you want to double the size of the image, you can pass in 2, but of course, you can also pass in 2.0.

zhouzaida (approved):

PR #2820, opened 2023-06-05
[Refactor] Keep the shape value of iou op's parameter 2 smaller than parameter 1

dflhw:
## Motivation

 It is found that parameter 2 of the iou operator does not support oversized shape, so add a check.

## Modification

If the value of gtboxesFP32‘s shape is less than bboxesFP32, swap them.


ckirchhoff2021 (approved):
/LGTM

PR #2821, opened 2023-06-05
[Fix] Keep the shape value of ascend iou op's parameter 2 smaller than parameter 1

dflhw:
## Motivation

It is found that parameter 2(ground truth) of the iou op does not support oversized shape([~32000+, 4]), so add a check.

## Modification

If the value of gtboxesFP32‘s shape is greater than bboxesFP32, swap them.


ckirchhoff2021 (approved):
/LGTM

Issue #2718, opened 2023-03-31
[Bug] active_rotated_filter op get a wrong value on cuda


dflhw:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmcv/issues) and [Discussions](https://github.com/open-mmlab/mmcv/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmcv).

### Environment

mmcv = 1.7.1
pytorch = 1.13.1
cuda = 11.2
python = 3.9.15

### Reproduces the problem - code sample

import torch
import numpy as np
from mmcv.ops import active_rotated_filter

device = torch.device("cpu")
print('===============>cpu')
feature = torch.tensor([[[[[385., 658., 852.],
                                         [619., 716., 106.],
                                         [189., 706., 143.]],
                                        [[394., 779., 399.],
                                         [268., 891., 822.],
                                         [262., 163., 958.]]]]], dtype=torch.float32, device=device)
indices = torch.tensor([[[[5, 4], [6, 5], [2, 3]], [[3, 2], [1, 6], [7, 7]], [[4, 8], [9, 1], [8, 9]]],
                                     [[[7, 1], [2, 5], [3, 4]], [[5, 2], [1, 7], [9, 8]], [[4, 3], [8, 6], [6, 9]]]], 
                                     dtype=torch.int32, device=device)
cpu_out = active_rotated_filter(feature, indices).cpu().numpy()
print("cpu_out: ", cpu_out[0])

print('===============>cuda')
feature = feature.cuda()
indices = indices.cuda()
cuda_out = active_rotated_filter(feature, indices).cpu().numpy()
print("cuda_out: ", cuda_out[0])

assert np.allclose(cuda_out, cpu_out, atol=1e-4)

### Reproduces the problem - command or script

When num_orientations is 2, the calculation result of active_rotated_filter op on cpu is inconsistent with that on cuda.

### Reproduces the problem - error message

Assertion Error

PR #2822, opened 2023-06-06
[Fix] Fix arf op's write conflict when num_orientations is not 1

dflhw:
## Motivation

1. When num_orientations is not 1, the calculation result of active_rotated_filter op on cpu is inconsistent with that on cuda.
2. Issue link: [2718](https://github.com/open-mmlab/mmcv/issues/2718)，the modified code can pass the sample test.
3. This change does not affect scenarios where num_orientations is 1.

## Modification

Indexes are added to prevent data from being written to only the first feature map.

## BC-breaking (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/ops/csrc/common/cuda/active_rotated_filter_cuda_kernel.cuh
+++ b/mmcv/ops/csrc/common/cuda/active_rotated_filter_cuda_kernel.cuh
@@ -25,9 +25,9 @@ __global__ void active_rotated_filter_forward_cuda_kernel(
     scalar_t val = *(weight_data + index);
     for (k = 0; k < num_rotations; k++) {
       int idx = (int)(*(indices_data + l * num_rotations + k)) - 1;
-      scalar_t* target = output_data +
-                         i * (num_rotations * num_input_planes * nEntry) +
-                         k * (num_input_planes * nEntry) + j * (nEntry) + idx + fmIndex;
+      scalar_t* target =
+          output_data + i * (num_rotations * num_input_planes * nEntry) +
+          k * (num_input_planes * nEntry) + j * (nEntry) + idx + fmIndex;
       *target = val;
     }
   }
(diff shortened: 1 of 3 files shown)


grimoire:
Hi, Thanks for the PR.
Actually, I am a little bit confused with this OP. 
Let's say, if two threads have the same `i`, `j`, `k`, `idx` (and `fmIndex`) and different `l`. They will store different `val` to the same `target`. That does seems not make sense.
I will try to contact the origin repo owner to see if it is a bug.

dflhw:
> Hi, Thanks for the PR. Actually, I am a little bit confused with this OP. Let's say, if two threads have the same `i`, `j`, `k`, `idx` (and `fmIndex`) and different `l`. They will store different `val` to the same `target`. That does seems not make sense. I will try to contact the origin repo owner to see if it is a bug.

Hi, Thanks for your reply. 
To ensure that every input value is taken, the values of indices here are combinations of permutation of 1~kH*kW.
Keep i, j, k have the same value.
When num_orientations is 1, the value range of l is 1~kH*kW.  In this case, the value of idx ranges from 1~kH*kW, there is no write conflit.
But when num_orientations is 2(or other num), the value range of l is 1~2*kH*kW, the value range of idx is still 1~kH*kW. So the value range of l is kH*kW~2*kH*kW will write to the same address as 1~kH*kW. In this case, it is need to add kH * kW to idx.
If num_orientations is 3, it is need to add kH*kW to idx when the value range of l is kH*kW~2*kH*kW and add 2*kH*kW to idx when the value range of l is 2*kH*kW~3*kH*kW.

grimoire:
Can I assume elements of `indices[oir_id, :, :, rot_id]` are shuffle of `arange(kH * kW)`, and each element is unique? I have no rotation detection background. 

PS: `bank conflict` is a term that is widely used to describe shared memory visit problems in CUDA. I think retitling this PR would be cool to avoid Ambiguity.

dflhw:
> Can I assume elements of `indices[oir_id, :, :, rot_id]` are shuffle of `arange(kH * kW)`, and each element is unique? I have no rotation detection background.
> 
> PS: `bank conflict` is a term that is widely used to describe shared memory visit problems in CUDA. I think retitling this PR would be cool to avoid Ambiguity.

Yeah, i think it is.
It seems that in this operator, in order to rotate a certain angle, there is a certain order of the indices[oir_id, :, :, rot_id].
Thanks for correcting the title, I will edit it.

grimoire (approved):
LGTM

dflhw:
> LGTM

Sorry, I misunderstood the use of indices，it seems that indices[:, :, :, rot_id] is permutation of arange(num_orientations * kH * kW), so the usage of indices in issue [2718](https://github.com/open-mmlab/mmcv/issues/2718) is wrong, it is not a bug, this pr should be reverted.

refer to: mmrotate [orconv.py](https://github.com/open-mmlab/mmrotate/blob/main/mmrotate/models/utils/orconv.py)

PR #2883, opened 2023-08-03
[Enhancement] Change the order of condition to make fx wok

youkaichao:
## Motivation

`mmcv.cnn.bricks.wrappers` has many forward functions in the form of `if x.numel() == 0 and obsolete_torch_version(TORCH_VERSION, (1, 5))`, where the first condition `x.numel() == 0` would fail `torch.fx`. We can take advantage of shortcut evaluation, by changing the order of condition to `if obsolete_torch_version(TORCH_VERSION, (1, 5)) and x.numel() == 0`. This way, the torch_version compares first, and will not fail `torch.fx` in modern pytorch versions.

## Modification

Change the order of condition, and add tests for it.

## BC-breaking (Optional)

This PR does not break any compatibility.

### DIFF
--- a/tests/test_cnn/test_wrappers.py
+++ b/tests/test_cnn/test_wrappers.py
@@ -4,6 +4,8 @@ from unittest.mock import patch
 import pytest
 import torch
 import torch.nn as nn
+from mmengine.utils import digit_version
+from mmengine.utils.dl_utils import TORCH_VERSION
 
 from mmcv.cnn.bricks import (Conv2d, Conv3d, ConvTranspose2d, ConvTranspose3d,
                              Linear, MaxPool2d, MaxPool3d)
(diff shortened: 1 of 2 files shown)


tests/test_cnn/test_wrappers.py
```diff
@@ -374,3 +374,26 @@ def test_nn_op_forward_called():
...
+            gm_module = fx.symbolic_trace(net)
+            print(gm_module.code)
+        for Net in (Linear, ):
+            net = Net(1, 1)
+            gm_module = fx.symbolic_trace(net)
+            print(gm_module.code)
+        for Net in (Conv2d, ConvTranspose2d, Conv3d, ConvTranspose3d):
+            net = Net(1, 1, 1)
+            gm_module = fx.symbolic_trace(net)
+            print(gm_module.code)
```
zhouzaida, line 396: Is it necessary to print the code?

tests/test_cnn/test_wrappers.py
```diff
@@ -374,3 +374,26 @@ def test_nn_op_forward_called():
...
+    except ImportError:
+        # torch.fx might not be available
+        pass
```
zhouzaida, line 399:
```suggestion
from mmengine.utils import digit_version
from mmengine.utils.dl_utils import TORCH_VERSION

@pytest.mark.skipif(
    digit_version(TORCH_VERSION) < digit_version('1.8'),
    reason='torch.fx needs Pytorch 1.8 or higher')
def test_fx_compatibility():
    from torch import fx

    # ensure the fx trace can pass the network
    for Net in (MaxPool2d, MaxPool3d):
        net = Net(1)
        gm_module = fx.symbolic_trace(net)
        print(gm_module.code)
    for Net in (Linear, ):
        net = Net(1, 1)
        gm_module = fx.symbolic_trace(net)
        print(gm_module.code)
    for Net in (Conv2d, ConvTranspose2d, Conv3d, ConvTranspose3d):
        net = Net(1, 1, 1)
        gm_module = fx.symbolic_trace(net)
        print(gm_module.code)
```

tests/test_cnn/test_wrappers.py
youkaichao: oh, that's used to supress the `unused variable` error.

tests/test_cnn/test_wrappers.py
youkaichao: Is mmengine always available in mmcv?

youkaichao: The skip condition is problematic: for pytorch 1.8 and pytorch 1.9, we have `torch.fx`, but pytorch native `MaxPool2d, MaxPool3d` does not support empty input. Therefore, `MaxPool2d, MaxPool3d` wrapped in mmcv will also fail `fx.symbolic_trace`.

zhouzaida: > Is mmengine always available in mmcv?

Right

zhouzaida: > The skip condition is problematic: for pytorch 1.8 and pytorch 1.9, we have `torch.fx`, but pytorch native `MaxPool2d, MaxPool3d` does not support empty input. Therefore, `MaxPool2d, MaxPool3d` wrapped in mmcv will also fail `fx.symbolic_trace`.

Got it, We can modify the 1.8 to 1.10.

tests/test_cnn/test_wrappers.py
zhouzaida: Oh, you can replace `gm_module = fx.symbolic_trace(net)` with `fx.symbolic_trace(net)`.

tests/test_cnn/test_wrappers.py
youkaichao: Fixed in [fc883ec](https://github.com/open-mmlab/mmcv/pull/2883/commits/fc883ec85f8bbd08fe934d1fa19a22b7051b3c04).

tests/test_cnn/test_wrappers.py
youkaichao: Fixed in [fc883ec](https://github.com/open-mmlab/mmcv/pull/2883/commits/fc883ec85f8bbd08fe934d1fa19a22b7051b3c04).

zhouzaida (approved):

PR #2905, opened 2023-08-21
Adapt new dipu device

CokeDong:
adapt new dipu device


CokeDong:
also fix: https://github.com/DeepLink-org/DIOPI/issues/266

grimoire (approved):

Issue #2916, opened 2023-08-30
[Bug] error: #error C++17 or later compatible compiler is required to use ATen.  


Mxbonn:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmcv/issues) and [Discussions](https://github.com/open-mmlab/mmcv/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmcv).

### Environment

N/A as problem is with installation

### Reproduces the problem - code sample

N/A -> installation issue

### Reproduces the problem - command or script

Use latest NVIDIA pytorch container (https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pytorch): 23-08
`pip install openmim`
`mim install mmcv`

### Reproduces the problem - error message

[log.txt](https://github.com/open-mmlab/mmcv/files/12475232/log.txt)
TL;DR:  installation fails due to multiple `error: #error C++17 or later compatible compiler is required to use ATen.` errors

Mxbonn:
Seems related to #2860

PR #2865, opened 2023-07-10
update c++17

grimoire:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

https://github.com/open-mmlab/mmcv/issues/2860

## Modification

comment constructor template of ShapeBase
c++14 -> c++17

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects, like MMDet or MMCls.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.

PR #2917, opened 2023-08-30
use c++17 in PyTorch 2

Mxbonn:
This fixed the issues in #2916 for me.
However, I'm not sure what the best practices are when it comes to c++ compilation so there is probably a cleaner solution for this.


Mxbonn:
Closing in favour of #2865

Issue #2926, opened 2023-09-06
[Bug] cache_randomness 🐞bug


HarborYuan:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmcv/issues) and [Discussions](https://github.com/open-mmlab/mmcv/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmcv).

### Environment

```
[...]

PR #2927, opened 2023-09-07
[Fix] fix data race risk of `cache_randomness`

HarborYuan:

## Motivation

**fix the bug mentioned in #2926**

## Modification

**The bug mentioned in #2926 is caused by the data race of multithreading. I do not think it is a good idea to use a pointer here to point to different instances. This can lead to disaster in concurrent situations. To keep modifications minimal, I implemented a version that creates a new copy every time "cache_randomness" is accessed to avoid data races of multithreading. Since this class "cache_randomness" only contains two pointers, one pointing to the class method (rather than the bound method) and the other to the class, the additional overhead brought here is subtle.**

## BC-breaking (Optional)

**I do not think it breaks**

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcv/transforms/utils.py
+++ b/mmcv/transforms/utils.py
@@ -81,6 +81,9 @@ class cache_randomness:
 
     def __get__(self, obj, cls):
         self.instance_ref = weakref.ref(obj)
+        # Return a copy to avoid multiple transform instances sharing
+        # one `cache_randomness` instance, which may cause data races
+        # in multithreading cases.
         return copy.copy(self)
 
 
--- a/mmcv/transforms/utils.py
+++ b/mmcv/transforms/utils.py
@@ -81,7 +81,7 @@ class cache_randomness:
 
     def __get__(self, obj, cls):
         self.instance_ref = weakref.ref(obj)
-        return copy.deepcopy(self)
+        return copy.copy(self)
 
 
 def avoid_cache_randomness(cls):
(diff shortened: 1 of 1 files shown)


mmcv/transforms/utils.py
```diff
@@ -80,7 +81,7 @@ def __call__(self, *args, **kwargs):
 
     def __get__(self, obj, cls):
         self.instance_ref = weakref.ref(obj)
-        return self
+        return copy.deepcopy(self)
```
ly015:
```suggestion
        return copy.copy(self)
```

HarborYuan: Done. Thanks for your suggestion.

mmcv/transforms/utils.py
```diff
@@ -80,7 +81,7 @@ def __call__(self, *args, **kwargs):
 
     def __get__(self, obj, cls):
         self.instance_ref = weakref.ref(obj)
-        return self
+        return copy.copy(self)
```
ly015, line 84: I suggest adding a few comments to briefly explain the copy operation.

ly015, line 84:
```suggestion
        # Return a copy to avoid multiple transform instances sharing
        # one `cache_randomness` instance, which may cause data races
        # in multithreading cases.
        return copy.copy(self)
```

PR #2973, opened 2023-10-20
DIOPI adapt codes for npu device

CokeDong:
diopi adapt codes for npu device


setup.py
```diff
@@ -244,10 +244,12 @@ def get_extensions():
             dipu_path = os.getenv('DIPU_PATH')
             vendor_include_dirs = os.getenv('VENDOR_INCLUDE_DIRS')
             nccl_include_dirs = os.getenv('NCCL_INCLUDE_DIRS')
+            pytorch_dir = os.getenv('PYTORCH_DIR')
```
grimoire, line 247: Can we get the pytorch_dir from `torch.__file__` or `torch.utils.cpp_extension.TORCH_LIB_PATH` instead of environment?

CokeDong: this is one env var pre-setted by `dipu` env script. ref: https://github.com/DeepLink-org/DIPU/blob/main/dipu/scripts/ci/ascend/ci_ascend_env.sh

grimoire (approved):
LGTM

zhouzaida (approved):

PR #3337, opened 2026-08-19
[Fix] Honor channel order for RGB TIFF images

### LINKED ISSUE #1101: rendering tiff file with wrong colors
**Describe the Issue**
Rendering tiff files with wrong color.
Original picture:
a
But got
Screenshot from 2021-06-15 15-05-11

**Reproduction**
Any tiff files, my example can be downloaded from [here](https://file-examples.com/index.php/sample-images-download/sample-tiff-download/)
```
import mmcv
img = mmcv.imread('file_example_TIFF_1MB.tiff', backend='tifffile')
mmcv.imshow(img)
```

**Bug fix**
Channel ordering problem

tandede:
## Motivation

Fixes #1101.

The `tifffile` backend returns standard color TIFF data in native RGB order, but MMCV's documented default is BGR. The backend previously returned the raw array without consulting `channel_order`, so default reads displayed red and blue channels incorrectly.

## Modification

- Read TIFF metadata to identify images whose photometric interpretation is RGB and locate their sample axis.
- Reorder the first three samples for BGR output while preserving alpha or other extra samples and supporting both contiguous and planar layouts.
- Keep non-RGB multi-channel TIFF data unchanged, preserving the backend's remote-sensing use case.
- Add regression coverage for RGB, RGBA, planar RGB, explicit RGB output, and the existing five-channel non-RGB fixture.

## BC-breaking (Optional)

No. This makes the `tifffile` backend honor the existing `channel_order` contract. Callers that need native RGB output can continue to pass `channel_order='rgb'` explicitly.

## Validation

- `pytest -q tests/test_image` (61 passed)
- Flake8, isort, YAPF, codespell, and whitespace checks pass for both changed files.
- No documentation update is needed because this restores the documented `channel_order` behavior.

### DIFF
--- a/mmcv/image/io.py
+++ b/mmcv/image/io.py
@@ -276,8 +276,15 @@ def imfrombytes(content: bytes,
             img = _pillow2array(img, flag, channel_order)
         return img
     elif backend == 'tifffile':
-        with io.BytesIO(content) as buff:
-            img = tifffile.imread(buff)
+        with io.BytesIO(content) as buff, tifffile.TiffFile(buff) as tif:
+            img = tif.asarray()
+            axes = tif.series[0].axes
+            if (channel_order == 'bgr'
+                    and tif.pages.first.photometric == tifffile.PHOTOMETRIC.RGB
+                    and 'S' in axes):
+                sample_axis = axes.index('S')
+                indices = [2, 1, 0] + list(range(3, img.shape[sample_axis]))
+                img = np.take(img, indices, axis=sample_axis)
         return img
     else:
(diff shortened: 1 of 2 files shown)


tandede:
Hi, I checked the current failed statuses. The CircleCI pipeline stops during configuration fetch with “No configuration was found in your project,” while both Read the Docs builds stop during configuration validation because `sphinx.configuration` is missing. None of these jobs reaches the TIFF implementation or its tests, so they appear to be repository CI configuration issues. Could a maintainer rerun or adjust these checks when convenient? Thanks!

PR #3339, opened 2026-08-30
Fix TIFF channel order

Marchematics:
Fix TIFF channel order.


Marchematics:
Preserved alpha for RGBA TIFFs and added BGRA coverage.

Marchematics:
The ReadTheDocs logs stop after checkout and configuration parsing, before any documentation command runs. The TIFF change and RGBA test are already included; this failure does not expose a documentation error in the patch.

